PR #26, opened 2014-03-27
added support for aac audio codec

chunder:
My ffmpeg binaries don't have support for the libfdk_aac audio codec. I downloaded my binaries directly from ffmpeg.org, for both OSX and Ubuntu. Neither had libfdk_aac support.

So this changes lets moviepy support the native ffmpeg aac audio codec. The only wrinkle is that if you use aac, you need to tell ffmpeg it's allowed to use "experimental features", hence the addition of the "-strict -2" arguments to the ffmpeg command line.

### DIFF
--- a/moviepy/audio/io/ffmpeg_audiowriter.py
+++ b/moviepy/audio/io/ffmpeg_audiowriter.py
@@ -52,6 +52,7 @@ class FFMPEG_AudioWriter:
                  [ "-i", input_video, '-vcodec', 'copy'])
             + ['-acodec', codec]
             + ['-ar', "%d"%fps_input]
+            + ['-strict', '-2']  # needed to support codec 'aac'
             + (['-ab',bitrate] if (bitrate!=None) else [])
             + [ filename ])
         
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -192,6 +192,7 @@ class VideoClip(Clip):
                 D_ext = {'libmp3lame': 'mp3',
                        'libvorbis':'ogg',
                        'libfdk_aac':'m4a',
+                       'aac':'m4a',
                        'pcm_s16le':'wav',
                        'pcm_s32le': 'wav'}
                 


dannguyen:
Sorry to bump this up, but I'm getting an error when trying to write a video file with the audio codec set to `"libfdk_aac"`

OS X (El Capitan) and have just done a fresh brew install of ffmpeg (2.8.6) with the `--with-fdk-aac` flag. I'm on Python 3.5.1 (installed with Anaconda) and moviepy __0.2.2.11__

ffmpeg from the command-line seems to work fine:

           ffmpeg -i input.wav -c:a libfdk_aac output.m4a


This is what I'm trying to do with moviepy:

~~~py
v = VideoFileClip('/tmp/test.mp4')
v.write_videofile("/tmp/test-x.mp4", audio_codec="libfdk_aac")
~~~



The following is an excerpt of the error message I get...the "bytes" error just seems to be a bug of string parsing...but it seems like the error is being raised in the first place because of an Unknown encoder error.


~~~
/Users/dtown/.pyenv/versions/anaconda3-2.4.0/lib/python3.5/site-packages/moviepy/audio/io/ffmpeg_audiowriter.py in write_frames(self, frames_array)
     88             error = (str(err)+ ("\n\nMoviePy error: FFMPEG encountered "
     89                      "the following error while writing file %s:"%self.filename
---> 90                      + "\n\n"+ffmpeg_error))
     91 
     92             if "Unknown encoder" in ffmpeg_error:

TypeError: Can't convert 'bytes' object to str implicitly
~~~

thenewguy:
@dannguyen if you want to use your own ffmpeg you need to provide it to moviepy.  For example from a python script you can do os.environ['IMAGEIO_FFMPEG_EXE'] = '/usr/local/bin/ffmpeg' or wherever your ffmpeg is located

Also, verify ffmpeg understands the file on the command line without using moviepy.

dannguyen:
ahhhh...thanks. Sorry I had thought IMAGEIO by default would detect an existing ffmpeg installation, but I guess that'd be complicated on the implementation side. Thanks!

PR #52, opened 2014-06-26
Added hasFFMPEGSupport to tools.py

autonomoid:
This is the tool I mentioned on one of the videogrep issues. It will check which codecs. formats, encoders and decoders your FFMPEG build supports. I've implemented it as a utility function. Hope its of use.


Zulko:
Hi there, I am closing this as old, since now a recent version of FFMPEG is automatically installed by ImageIO.

PR #65, opened 2014-09-23
add Dockerfile

srid:
Docker is becoming one of the most popular ways to install things. I built a docker image for moviepy so that I can install it (on my Mac) easily using a single command:

```
docker pull srid/moviepy
```

The above image comes from a "trusted build" I setup here https://registry.hub.docker.com/u/srid/moviepy/ - which automatically runs whenever my fork changes. Should this pull request be accepted, I imagine the maintainers of this repo would setup a similar "Zulko/moviepy" trusted build and recommend its use from the README. Or, the image can be created manually on demand during releases.

Any project that uses moviepy can leverage docker to build itself, using its own Dockerfile like for instance:

```
FROM srid/moviepy

ADD requirements.txt /app/requirements.txt
WORKDIR /app
RUN pip install -r requirements.txt

ADD . /app
RUN mkdir /app/output
```

Input and output videos can be managed using docker volumes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu
+
+# Install numpy using system package manager
+RUN apt-get -y update && apt-get -y install python-numpy python-pip
+
+# Add optional but recommended libraries (scipy and friends)
+RUN apt-get -y install python-scipy python-sklearn
+
+# MoviePy requires the ffmpeg binary. Ubuntu PPA doesn't work at the
+# moment, so install directly from official static binary release.
+ADD http://johnvansickle.com/ffmpeg/releases/ffmpeg-2.4-64bit-static.tar.xz /f.tar.xz
+RUN cd / && tar xvfJ f.tar.xz
+RUN ls -l /ffmpeg*
+RUN ln -s /ffmpeg-*/ffmpeg /usr/bin/
+RUN ln -s /ffmpeg-*/ffprobe /usr/bin/
+# Verify that ffmpeg works.
+RUN /usr/bin/ffmpeg || true
+
+ADD . /moviepy
+WORKDIR /moviepy
+RUN pip install .


earney:
much of the contents of the dockerfile can be modified to the following:

```
FROM ubuntu
 
# Install numpy using system package manager
RUN apt-get -y update && apt-get -y install python-numpy python-pip

# Add optional but recommended libraries (scipy and friends)
RUN apt-get -y install python-scipy python-sklearn

RUN apt-get -y install ffmpeg

# Verify that ffmpeg works.
RUN /usr/bin/ffmpeg || true
```

should we look into adding some documentation about how to create a dockerfile for moviepy, instead of adding a file?  Many users may not be interested in the dockefile, or its presence may cause confusion.

earney:
since only a small percentage of users maybe interested in this open, lets look at creating some documentation for this.  Maybe use the updated dockerfile I posted above, as a sample someone could use?

kencochrane:
The advantage of including the dockerfile in the git repo is so that you can create an official automated build on docker hub. When ever a commit is made it will automatically rebuild and update the image. 

If you don't want the dockerfile at the root of the git repo you can bury in a sub directory if that is better.

I would love to have an official dockerfile for this project, it would make getting started much easier.

In addition to your comment above, if you base the dockerfile off an official Python image, you can remove a few more steps. The one issue is that those images are based on Debian and alpine, not ubuntu so not sure if that is a problem.

https://hub.docker.com/_/python/

If you need help with the docs or the dockerfile, let me know

earney:
I was not aware of those advantages.  Thanks for sharing that.  I will look into this asap.

kencochrane:
@earney cool, let me know if you need any help, or have any questions.

earney:
@kencochrane     here is my current Dockerfile.  It seems to have some problems installing ffmpeg on debian jessie.   any ideas?  IF not, I think we can stick with the dockerfile I posted above..

```
From python:3

# Install numpy using system package manager
RUN apt-get -y update && apt-get -y install python-numpy python-pip

# Add optional but recommended libraries (scipy and friends)
RUN apt-get -y install python-scipy python-sklearn

RUN apt-get -y install ffmpeg

# Verify that ffmpeg works.
RUN /usr/bin/ffmpeg || true
```

Zulko:
does it work if you run imageio's ffmpeg downloader instead of downloading ffmpeg yourself ?

kencochrane:
@earney I read somewhere (forgot where), that we shouldn't install ffmpeg from the distribution, we should instead let imageio install it for us.

Here is my current Dockerfile, it still needs to be cleaned up. If this was imbedded in the git repo, we can remove the need to clone the repo, and instead ADD or CP the files directly, which is safer.

```
FROM python:3

# Install numpy using system package manager
RUN apt-get -y update && apt-get -y install libav-tools imagemagick libopencv-dev python-opencv

RUN apt-get install -y locales && \
    locale-gen C.UTF-8 && \
    /usr/sbin/update-locale LANG=C.UTF-8

ENV LC_ALL C.UTF-8

# do we need all of these, maybe remove some of them?
RUN pip install imageio numpy scipy matplotlib pandas sympy nose decorator tqdm pillow

# install scikit-image after the other deps, it doesn't cause errors this way.
RUN pip install scikit-image

# install from github since there are bugs in latest versions.
RUN git clone https://github.com/Zulko/moviepy.git /var/src/moviepy
RUN cd /var/src/moviepy/ && python setup.py install

# install ffmpeg from imageio.
RUN python -c "import imageio; imageio.plugins.ffmpeg.download()"

# Verify that ffmpeg works.
RUN /usr/bin/ffmpeg || true
```

I'm still working on cleaning it up, but figured it would be good to show you what I have so far.

kencochrane:
> does it work if you run imageio's ffmpeg downloader instead of downloading ffmpeg yourself ?

@Zulko yes it does, and that is what I did in my example above.

earney:
Awesome!  I'll give your dockerfile a try, and will let you know.

kencochrane:
@earney cool thanks, I just realized the last line ```RUN /usr/bin/ffmpeg || true``` won't work since it actually installs ffmpeg in a different place, so I'll remove that line.

kencochrane:
Here are the steps I use to test the image, for those who also want to test at home. 

If you don't have docker installed, you can use Docker for Mac, Docker for Windows, or if you are on Linux you can follow directions here, based on your distro: https://docs.docker.com/engine/installation/

```
# build the dockerfile and tag it as moviepy. make sure you are in the same directory as the Dockerfile.
docker build -t moviepy -f Dockerfile .

# run the moviepy image and volume mount the current directory into /code in the container and start a bash prompt.
docker run -it -v .:/code moviepy bash

# inside of the container I run my moviepy code inside of videocode.py
cd /code
python videocode.py
```

earney:
@kencochrane  when I try to run the second command I get:

```
docker run -it -v .:/code moviepy1 bash
docker: Error response from daemon: create .: volume name is too short, names should be at least two alphanumeric characters.
See 'docker run --help'.
```

kencochrane:
@earney sorry about that, you can either use a absolute path for where your code is located (/home/myname/mycode)

```
docker run -it -v /home/myname/mycode:/code moviepy1 bash
```
or do the following to mount the current directory.
```
docker run -it -v `pwd`:/code moviepy1 bash
```

kencochrane:
Also, incase anyone else wants to know, here is how you can install ffmpeg in debian jessie.

```
RUN sed -i "s/jessie main/jessie main contrib non-free/" /etc/apt/sources.list
RUN echo "deb http://http.debian.net/debian jessie-backports main contrib non-free" >> /etc/apt/sources.list
RUN apt-get update && apt-get install -y ffmpeg
```

I was having some issues with that version, and switched to the imageio way as described above. I can't remember what the issues I had with this version of ffmpeg, so try at your own risk.

earney:
I've had to make some modifications to get a few of the tests working under the config provided by the dockerfile, but it seems to be working now.  One of the main issues deal with what fonts are available, so I had to modify some of the tests so that they work with fonts that exist under the dockerfile and exists on my system (ubuntu 16.10).  I see that we probably will need to adjust this in the future by setting default fonts based on OS, etc.

Feel free to take the dockerfile that exists on the dockerfile branch of earney/moviepy and modify your existing Pull request with this file.   I can PR the changes to the tests sometime later.

kencochrane:
@earney good point about the fonts, I forgot about that one, I guess we just need to find a way to install the same fonts in the image, not sure if possible, but something to look at.

earney:
I have not heard back from @srid, so I will put in a pull request for my dockerfile, and will close this PR once that is done.

earney:
closing since pr #522 creates the same result.

PR #112, opened 2015-01-02
Use six to address py2/py3 urllib.urlretrieve module change.

Who8MyLunch:


carlodri:
Sorry I'm mostly py-ignorant, but why not simply say:
```python
from urllib.request import urlretrieve
```
is it a matter of `python2`/`python3` compatibility?

Who8MyLunch:
You have it right @carlodri .  The six library makes it easy for a single codebase to easily support both Python 2 and Python 3.

Zulko:
Thanks for the report @Who8MyLunch ! What is the bug exactly ? Does it only bug when you call the downloader module, or does this prevent installation with Python 3 ?

I am a little against adding a new dependency. Could this bug be solved with a with/try statement ?

Who8MyLunch:
@Zulko The bug was in the form of an import error when importing editor.py. The module `urllib.urlretrieve` does not exist in Python 3.  Instead you have `urllib.request.urlretrieve`.

This is just as easily addressed via try: statement.  I'll do that tonight when I get home.

Zulko:
So it has been broken one month for Py3 users ??? Wow, I definitely need a test suite (I have an unofficial one actually, I just forgot to run it).

carlodri:
It is broken only if the user updates the package explicitly (like me), or
for new users. Thanks @Zulko, moviepy is a great package!

Zulko:
I corrected it (on github and pypi). tell me guys if it works for you.

Who8MyLunch:
It works fine for me!

Zulko:
Ok, so I close this one.

carlodri:
Confirmed working on Mac OS X, thanks, will test on Win8x64 in the afternoon and report back.

Issue #80, opened 2014-10-12
[Windows users: help !] Finding ImageMagick automatically on windows


Zulko:
Just in case there are Windows users watching the project: on linux and mac the imagemagick executable is easy to access (it is simply called `convert`). But in windows there is a name conflict so the users need to enter the path to the binary themselves. This prevents windows users from installing with pip, and it seems to be difficult to do for python beginners and non-computer experts in general. The path to provide looks like this

```
C:\Program Files\ImageMagick-6.8.3-Q16\convert
```

I can easily imagine that there could be a way to auto-detect that file so that windows users would have nothing set manually. But I don't want to write it myself because I am not on windows, so that would be difficult to test. Could someone do that ?

As I see it it would consist in changing this part of config.py:

```
if IMAGEMAGICK_BINARY == 'auto-detect':

    if try_cmd(['convert'])[0]:
        IMAGEMAGICK_BINARY = 'convert'
    else:
        IMAGEMAGICK_BINARY = 'unset'
```

by adding:

```
if os.name == "nt":
    # find the file
    IMAGEMAGICK_BINARY = (# found file else 'unset')
elif try_cmd(['convert'])[0]:
    etc.
```

kerstin:
Looks like this issue was fixed by the linked (and already merged) PR and can thus be closed. Thanks again for the contribution, @carlodri.

PR #118, opened 2015-02-02
auto detection of ImageMagick binary on Windows

carlodri:
This is how I would implement the enhancement in issue #80. I'm very much a Python (and GitHub..) beginner, so please be patient if I made silly mistakes.

This works for me on both Win8x64 and Mac OS X 10.10.1 with Python 3.4


Zulko:
Thanks for this, this is really helpful !

Issue #126, opened 2015-02-12
Getting a TypeError in FramesMatch


Prooffreader:
When running the following:

&gt;&gt;&gt; from moviepy.editor import VideoFileClip
&gt;&gt;&gt; from moviepy.video.tools.cuts import FramesMatches
&gt;&gt;&gt; clip = VideoFileClip("myvideo.mp4")
&gt;&gt;&gt; clip_small = clip.resize(width=150)
&gt;&gt;&gt; matches = FramesMatches.from_clip(clip_small, 5, 3)
&gt;&gt;&gt; matches

I get the following output:
    &lt;repr(&lt;moviepy.video.tools.cuts.FramesMatch at 0x86d3a90&gt;) failed: TypeError: not all arguments converted during string formatting&gt;,
    &lt;repr(&lt;moviepy.video.tools.cuts.FramesMatch at 0x86d3a58&gt;) failed: TypeError: not all arguments converted during string formatting&gt;,
    etc.

Prooffreader:
This error is with a fresh install of Ubuntu 14.04 and Anaconda Python Distribution (Python 3,4) on a micro AWS instance; all that was done is pip install moviepy and pip install pillow.

Prooffreader:
Same error occurs with a fresh install of Ubuntu 14.04 on AWS, Python 2.7.6, pip install moviepy and pillow.

filipochnik:
Should be fixed in https://github.com/Zulko/moviepy/pull/131

kerstin:
Looks like this issue was fixed by the linked (and already merged) PR and can thus be closed. Please feel free to reopen the issue/create a new one if the problem persists.

PR #131, opened 2015-02-22
Issue #126 Fix FramesMatch repr and str.

filipochnik:
A small bug in the repr and str of FramesMatch.


Zulko:
Thanks !

filipochnik:
No problem :)

PR #132, opened 2015-02-22
setup.py: Link to website and state license

techtonik:
This will make package appear in results when you're explicitly
searching for projects with permissive licenses.


Zulko:
Good point. Thanks.

Issue #146, opened 2015-03-25
Error in /video/fx/freeze_region.py


savannahniles:
Hi Zulko. I was having trouble with getting good results from the [freeze_region function](https://github.com/Zulko/moviepy/blob/master/moviepy/video/fx/freeze_region.py) when I specified a region. Specifying an outer_region gave exactly the expected response, but specifying a region positioned the region at 0,0 regardless of the values of x1 and y1. 

The portion of the code that freezes a region currently reads:

```
if region is not None:

        x1, y1, x2, y2 = region
        freeze = (clip.fx(crop, *region)
                      .set_position((x1,y1))
                      .to_ImageClip(t=t)
                      .set_duration(clip.duration))
        return CompositeVideoClip([clip, freeze])
```

However I get the correct results when I rewrote and tested the function by setting the position last instead of first like this:

```
if region is not None:

        x1, y1, x2, y2 = region
        freeze = (clip.fx(crop, *region)
                      .to_ImageClip(t=t)
                      .set_duration(clip.duration)
                      .set_position((x1,y1)))
        return CompositeVideoClip([clip, freeze])
```

Thanks for building this amazing tool. :purple_heart:

Zulko:
Thanks for the report ! Do you want to make a pull request ?

Zulko:
Solved by the PR referenced above.

PR #147, opened 2015-03-25
fixing region parameter on freeze_region

savannahniles:
I was having trouble with getting good results from the freeze_region function when I specified a region. Specifying an outer_region gave exactly the expected response, but specifying a region positioned the region at 0,0 regardless of the values of x1 and y1. I get the correct results when I rewrote and tested the function by setting the position last instead of first (lines 38 - 40). Issue [here](https://github.com/Zulko/moviepy/issues/146#issuecomment-85999861)


Zulko:
Thanks !

Issue #128, opened 2015-02-16
Progress bar


PlanPh:
Hi,

I am really fine with moviepy - its a really great toolbox! I´ve writed a prgramm with a GUI to cut some videos in a specific way. This works really well. 
Now i want to display the status of the cutting process on the interface like its displayed on my ipython console. But unfortunally i am not able to display the progress bar on my GUI.
I´ve tried "verbose_print" - but i cant call it the correct way.
Does anybody know the solution?

Thanks a lot! P

Zulko:
I don't know. Maybe the best is to ask the question on reddit/python, reddit/learnpython or stackoverflow. Your question would be: "How can I retrieve and display in a [Tkinter ?] GUI the progress bar printed in the console by the library tqdm ?"

BAXTER001:
I made a patch that extends the *write_video methods with a parameter to specifying the filelike object tqdm writes to, makes it very simple to pipe the progress between threads/processes, would you be interested in a pull request for it?

Zulko:
@BAXTER001 Thanks, but wouldn't it be simpler to run moviepy in a separate thread and get the stdout ? With your solution we would have to amend all the functions that use the progress bar at some point, in case someone needs to print it externally.

mgaitan:
That's interesting, but changing just the progress bar's file to we will have a kind "progress bar screenshots" as text, and to do get the actual data (current percentage, ETA, etc) we should parse it. 

I guess that a better approach  should be to pass an optional 'callable' to the the progressbar function that receive `n, total, percentage, elapsed_str, left_str, rate` on each loop. Then this callable could do anything, for example, print the to the stdout/stderr as currently do, update a widget in a GUI, etc. 

This imply a pull-request to tqdm or, probably faster, handle this in our own way. 

I could make a pull-request with this refactor. What do you think?

BAXTER001:
@Zulko I avoided that out of the perhaps irrational unease of doing the sys.stdout replacement dance when the specificity of just the tqdm output on a per-clip basis was possible.

@mgaitan That does sound like a far cleaner and more useful method.

Zulko:
Ok, looks like we are progressing towards a solution. @BAXTER001 what do you mean by "setter on clip ?" I'd propose that, maybe that's what you and @mgaitan had in mind ?

    clip.iterframe_callback = some_function # some_function(iframe, frame, number_of_frames)

[...]

PR #150, opened 2015-04-03
Adds iterframe_callback as per #128

flovilmart:


Zulko:
Thanks for the commit. I'm still unsure how this should be implemented. In #128 @BAXTER001 proposed *a generator function to this method to be wrapped around the current generator()*, which is also a good idea. An objection I have in your implementation is that you propose iterframe_callback as a class methods, and in python 3 these can only be overwritten by creating a new subclass. I would prefer if it was just an attribute. But yet again, I'm not sure it shouldn't be a wrapper around range().

flovilmart:
Ok. The thing is that I hacked it quickly as I needed to estimate the processing time for the rest of my pipeline.




I generally don't care much about the form of the solution, as long as it's effectively working.




Replace the tqdm generator with a custom progress would be a solution. But also I'm not sure this is what you 'want'.

lrq3000:
Hey there, the [tqdm project is now maintained by another team of developpers](https://github.com/tqdm/tqdm) (which I am part of), and the version has been updated to v3.1.4 lately (but the API should be the same, this should not break anything from what I've seen by quickly skimming through moviepy code).

We have described in the readme an example of [how to do a wrapper/callback to tqdm](https://github.com/tqdm/tqdm#hooks-and-callbacks), without needing to modify tqdm's code.

If this is not enough, the project is also now modularized and object-oriented, so that you can overload the tqdm class to easily interface with your setup, like we did for [IPython](https://github.com/tqdm/tqdm/blob/ipython-notebook/tqdm/_tqdm_notebook.py) and [pandas](https://github.com/tqdm/tqdm/blob/master/tqdm/_tqdm_pandas.py). This allows to let tqdm do the ugly work (time prediction, skipping useless iterations to lower CPU usage, etc.), and just focus on interfacing with your app.

Feel free to come and ask us if you cannot sort this out, we can maybe help you achieve what you want (and make tqdm better by the way).

flovilmart:
the PR was in response to #128 and I actually needed that for a project. As this branch is stable, it will still be in use in the project and we'll see in the future if we need to update.

Thanks.

F.

lrq3000:
Ok :) Feel free to come talk with us if you have a specific need, we try to make tqdm more and more flexible without compromising on the performances ;)

mbeacom:
@flovilmart Thank you for the contribution!

Would you like to rebase this for further review or would you prefer to close the PR?

Thanks

flovilmart:
I don't use the project anymore, but I could give it a shot if it's something you want in main codebase.

jameshhx:
One note I'll make on this is that when using the `CompositeAudioClip`, this will throw an error due to `iterchunk_callback` being undefined. This is due to the fact `CompositeAudioClip` calls `__init__` on `Clip`, instead of the class it's extending `AudioClip`. I've fixed this issue locally by changing where `CompositeAudioClip` calls to `__init__`. This feature is something that I'd find useful - is it something you'd be willing to consider merged, if I can get it to a complete state?

flovilmart:
Closing as I'm not using the project anymore. Feel free to pick it up!

Zulko:
Oh yeah sorry for that and thanks for the past efforts. The project now uses [Proglog](https://github.com/Edinburgh-Genome-Foundry/Proglog) loggers for logging, which makes it possible to have custom callbacks, redirect logs to a frontend, etc. by configuring the Proglog loggers.

flovilmart:
No problem! Thanks you for still maintaining itb

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #153, opened 2015-04-12
Removed superfluous +

adalq:
The + in front of np.xxx are not required

### DIFF
--- a/moviepy/video/io/bindings.py
+++ b/moviepy/video/io/bindings.py
@@ -11,7 +11,7 @@ def PIL_to_npimage(im):
     return np.array(im)
     #w,h = im.size
     #d = (4 if im.mode=="RGBA" else 3)
-    #return +np.frombuffer(im.tobytes(), dtype='uint8').reshape((h,w,d))
+    #return np.frombuffer(im.tobytes(), dtype='uint8').reshape((h,w,d))
 
 
 def mplfig_to_npimage(fig):
@@ -19,7 +19,7 @@ def mplfig_to_npimage(fig):
     fig.canvas.draw() # update/draw the elements
     w,h = fig.canvas.get_width_height()
     buf = fig.canvas.tostring_rgb()
-    image= +np.fromstring(buf,dtype=np.uint8)
+    image = np.fromstring(buf,dtype=np.uint8)
     return image.reshape(h,w,3)
 
 


Zulko:
It is required. It is a dirty way to make a copy of the array. Without it, the array is just a view of the string, but that string gets erased when the function ends, and if that string is overwritten in memory before the frame is written to a file, you get a very strange (like, noisy or partly noisy) rendering of your clip. If you don't see that on your machine, then maybe it's platform-specific or python-version specific.

adalq:
That's strange, the numpy docs don't mention that `fromstring` returns a view: http://docs.scipy.org/doc/numpy/reference/generated/numpy.fromstring.html

I can also see a `memcpy` in the `fromstring` implementation. Call tracing:
https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/multiarraymodule.c#L3923
https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/multiarraymodule.c#L2082
https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/multiarraymodule.c#L2097
https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/ctors.c#L3528
https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/ctors.c#L3584

I've also done a test (Python 2.7 64 bit, Windows 8.1 64 bit). Source code and output below. It uses the `psutil` package for memory stats.

You can see that the `fromstring` call allocates another memory chunk, and that deleting the original string (and confirming the OS got the memory back) doesn't corrupt the numpy array content.

```python
from __future__ import division, print_function

import os

import psutil
import numpy as np

pid = os.getpid()
proc = psutil.Process(pid)

def print_mem_stats():
    proc_mem = proc.memory_info()
    print("rss={:5,} MB   vms={:5,} MB".format(proc_mem.rss // (1024 * 1024), proc_mem.vms // (1024 * 1024)))
    print("")

print_mem_stats()

print("Alloc 1 GB string")
s = "\x00\x01\x02\x03" * ((1024 * 1024 * 1024) // 4)
print_mem_stats()

print("Create numpy array")
v = np.fromstring(s, dtype=np.uint8)
print_mem_stats()

print("Delete string")
del s
print_mem_stats()

print("Check numpy array content")
for i in xrange(64):
    print(v[i], end="")
print("")
print_mem_stats()

print("Delete numpy array")
del v
print_mem_stats()
```

Output:

    rss=   19 MB   vms=   15 MB

    Alloc 1 GB string
    rss=1,043 MB   vms=1,041 MB

    Create numpy array
    rss=2,067 MB   vms=2,067 MB

    Delete string
    rss=1,043 MB   vms=1,041 MB

    Check numpy array content
    0123012301230123012301230123012301230123012301230123012301230123
    rss=1,043 MB   vms=1,041 MB

    Delete numpy array
    rss=   19 MB   vms=   15 MB

Maybe the corruption you noticed happened in a multi-threaded environment, where another thread deleted the string before `numpy.fromstring` returned? Or maybe `fromstring` behaved differently in past numpy versions.

earney:
closing this PR, since it looks like we do need the plus sign.

PR #157, opened 2015-04-15
poor error message from ffmpeg_reader.py

ryanfox:
Per [http://stackoverflow.com/questions/11497234/typeerrorexceptions-must-be-old-style-classes-or-derived-from-baseexception-no](http://stackoverflow.com/questions/11497234/typeerrorexceptions-must-be-old-style-classes-or-derived-from-baseexception-no), as of python 2.6, you can't raise just a string.  If this particular error is triggered, you just get a message

    {TypeError} exceptions must be old-style classes or derived from BaseException, not unicode

The nearby exceptions are IOErrors, so I made this one an IOError as well.  Now it shows the real exception, instead of the cryptic message above.


ryanfox:
For what it's worth, I'm using python 2.7.6 on Linux.

Zulko:
Thanks !

Issue #160, opened 2015-04-25
embed clip in a jupyter notebook


laurentperrinet:
using a video clip generated by:

```
duration = 20.
fig_mpl, ax = plt.subplots(1, figsize=(6, 5), facecolor='white')

def draw_elementary_pattern(ax, center):
    ax.add_artist(mpl.patches.Wedge(center, 1., 0, 180, width=.1))

def make_frame_mpl(t):
    ax.cla()
    draw_elementary_pattern(ax, t/duration)
    return mplfig_to_npimage(fig_mpl) # RGB image of the figure

animation = mpy.VideoClip(make_frame_mpl, duration=duration)
```

I am trying to embed it in the notebook using:

```
mpy.ipython_display("circles.webm", fps=20, loop=1, autoplay=1)
```

This was working nicely when I was using python 2, but seems to be broken when using python 3:

(The video generated by

```
animation.write_videofile("circles.webm", fps=20)
```

would show fine in the browser when opened directly)

```
$ python3 --version
Python 3.4.3

$ ipython --version
3.1.0
```

any clue?

(and thanks a lot for this excellent piece of code!)

![screen shot 2015-04-25 at 16 26 43](https://cloud.githubusercontent.com/assets/381808/7333248/53c10966-eb68-11e4-93b8-4fa57a259d37.png)

Zulko:
I just tried that exact line and it works for me on my jupyter... are you using firefox or chrome ?

Also, hint: instead of first writing to a file and then importing in the notebook, you can directly preview the clip with

    clip.ipython_display(fps=20, loop=1, autoplay=1)

which is faster to write in the notebook.

meduz:
when doing 

````
animation.write_videofile("circles.mp4", fps=20)
from IPython.core.display import HTML
HTML("""
    <video controls src="{0}" loop=1 autoplay width=100%/> 
    """.format('circles.mp4')
)
````
it would work while 

````
def video(fname, mimetype):
    """Load the video in the file `fname`, with given mimetype, and display as HTML5 video.
    """
    from IPython.display import HTML
    video_encoded = open(fname, "rb").read().encode("base64")
    video_tag = '<video controls alt="test" src="data:video/{0};base64,{1}">'.format(mimetype, video_encoded)
    return HTML(data=video_tag)
video('circles.mp4', 'mp4')
````

gives

````
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
[...]

PR #162, opened 2015-04-26
fix to work with python3

meduz:
tries to solve https://github.com/Zulko/moviepy/issues/160

tested with python  2.7.9 and 3.4.3

(sorry my editor automatically "cleaned" spaces - I fixed that)

### DIFF
--- a/moviepy/video/io/bindings.py
+++ b/moviepy/video/io/bindings.py
@@ -16,9 +16,16 @@ def PIL_to_npimage(im):
 
 def mplfig_to_npimage(fig):
     """ Converts a matplotlib figure to a RGB frame after updating the canvas"""
-    fig.canvas.draw() # update/draw the elements
-    w,h = fig.canvas.get_width_height()
-    buf = fig.canvas.tostring_rgb()
+    from matplotlib.backends.backend_agg import FigureCanvasAgg
+
+    agg = fig.canvas.switch_backends(FigureCanvasAgg)
+    agg.draw() # update/draw the elements
+
+    # get the width and the height to resize the matrix
+    l,b,w,h = agg.figure.bbox.bounds
+    w, h = int(w), int(h)
+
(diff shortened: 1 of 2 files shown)


Zulko:
Hey there, can you explain the changes in mplfig_to_npimage ?

meduz:
To port moviepy's code to python 3, a difference is that strings are now represented as unicode and data as bytes while this was different (and sometimes problematic) in python 2:

https://docs.python.org/3.3/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit

This was already implemented in matplotlib, trying to make this backportable with python 2.7, see for instance

https://github.com/matplotlib/matplotlib/commit/f4fee350f9fbc639853bee76472d8089a10b40bd

in fact, there is no change to the API, but the correct backend has to be called.

basically, what I am doing here is to call for the correct backend (Agg) before calling the same functions and return a numpy array.

hope it is a bit clearer, I am still learning my way to help contribute to the libraries that I use.

meduz:
I still have problems rendering movies in a notebook using python 3 - see http://blog.invibe.net/posts/2015-04-26-moviepy-and-python3.html - do you have the same problem?
cheers!

Zulko:
I didn't have time to look into this (im still on Python 2 in my notebooks), now its unclear, does the commit actually solves the problem ?

meduz:
Yes the commit solves the problem in py3k - and should be back-portable with python 2.7 and higher.

flothesof:
Hi everyone,

I would like to add that I have encountered the same problem than @meduz. Would it be possible to accept the PR so as to fix this problem with Python 3?

Best regards,

Florian

Krastanov:
Any progress on this?

Zulko:
Sorry for being slow. Too bad moviepy doesn't have a test suite.

PR #165, opened 2015-04-29
set correct lastindex variable in mask_make_frame

Dennovin:
The mask_make_frame function checks self.lastindex but sets self.mask.lastindex.


Zulko:
Thanks. Actually I just spotted that bug on my machine a few days ago but didn't push the correction on github.

PR #171, opened 2015-05-23
Support for SRT files with any kind of newline

factorial:
Fixes bug where an SRT file with Windows newlines ("\r\n") was not properly parsed on OSes that do not use Windows newlines. After this change, successfully parses my The Big Lebowski SRT on OS X.


Zulko:
Thanks !

PR #175, opened 2015-05-31
Fixes Windows permission error

psykzz:
```
  File "test_downloader.py", line 30, in <module>
    main()
  File "test_downloader.py", line 27, in main
    download('_UfYjd50c6I')
  File "test_downloader.py", line 22, in download
    le_file = (VideoFileClip(_file))
  File "C:\Python27\lib\site-packages\moviepy\video\io\VideoFileClip.py", line 5
5, in __init__
    reader = FFMPEG_VideoReader(filename, pix_fmt=pix_fmt)
  File "C:\Python27\lib\site-packages\moviepy\video\io\ffmpeg_reader.py", line 3
2, in __init__
    infos = ffmpeg_parse_infos(filename, print_infos, check_duration)
  File "C:\Python27\lib\site-packages\moviepy\video\io\ffmpeg_reader.py", line 2
40, in ffmpeg_parse_infos
    proc.terminate()
  File "C:\Python27\lib\subprocess.py", line 1002, in terminate
    _subprocess.TerminateProcess(self._handle, 1)
WindowsError: [Error 5] Access is denied
Exception AttributeError: "VideoFileClip instance has no attribute 'reader'" in
<bound method VideoFileClip.__del__ of <moviepy.video.io.VideoFileClip.VideoFile
Clip instance at 0x03E5FBC0>> ignored
```

### DIFF
--- a/moviepy/video/io/ffmpeg_reader.py
+++ b/moviepy/video/io/ffmpeg_reader.py
@@ -237,7 +237,13 @@ def ffmpeg_parse_infos(filename, print_infos=False, check_duration=True):
     proc = sp.Popen(cmd, **popen_params)
 
     proc.stdout.readline()
-    proc.terminate()
+    # Terminating the process on windows will cause you to 
+    # get a Permission denied error. You can however safely 
+    # ignore that error and continue as normal.
+    try:
+        proc.terminate()
+    except:
+        pass
     infos = proc.stderr.read().decode('utf8')
     del proc
 


Zulko:
Thanks, but if it doesn't terminate on Windows, does it mean that you have a ghost (unclosed) process ? (my concern is that these processes can take place in memory, and people generally don't like them, so if there is a better solution than just ignoring the error, it's better).

psykzz:
It wasn't found running In my case, but I do agree.

peircej:
I was concerned about the same but I don't think so. Using this code:
```
for n in range(100):
   mov = moviepy.video.io.VideoFileClip(......)
```
after applying the patch, Task Manager doesn't show any gradual increase in processes. Maybe they're getting terminated by subprocess as the process is garbage collected?

Incidentally, I've applied a similar patch to the `close` methods of both `FFMPEG_VideoReader` and `FFMPEG_AudioReader`. In those cases the error message creeps in during `__del__` and is harmless but annoying. The same trick prevents it and doesn't seem to lead to increased numbers of processes on the system

psykzz:
Any updates?

earney:
@psykzz ..  are you still having these issues with the current repo?   If so, we'll look at adding this PR.

psykzz:
Not actively anymore. I was requesting an update just to close off the PR really.
If its useful, we can merge, otherwise ill close the PR.

earney:
okay.. thanks.. I believe this has been resolved in another PR.  If you continue having this issue, please reopen this PR..

Droppers:
This is a really annoying issue and I can not use this library without this fix.

This does not seem to be solved in any PR, including https://github.com/Zulko/moviepy/pull/626

PR #179, opened 2015-06-20
Use of communicate() and explicit str cast for python 3

rutherford:
Replacing stderr.read() call with safer communicate(). Also in python 3.4.3 I found a byte object is returned and not a string so an explicit cast is also required. I put in the necessary cast before the underlying ffmpeg error can be dealt with in the expected manner so this also fixes an existing defect for moviepy on python 3.


earney:
@rutherford ..   thanks..  we decided to use the error codes to be byte strings so the str(...) isn't necessary, but we did change the stderr.read, and used your recommendation (and used the communicate function instead)..   Thanks!

Issue #158, opened 2015-04-15
AudioFileClip bombs on file read


ryanfox:
Related: https://github.com/Zulko/moviepy/pull/157

I am trying to create an AudioFileClip from an mp3 file like so:

    clip = AudioFileClip('myaudio.mp3')

On line 275 of ffmpeg_reader.py, it looks like whether a file is video or not is determined by whether there is the string " Video: " anywhere in the result from ffmpeg.  At least in my case, this is throwing false positives.  This is the string that triggers it for me:

    u'    Stream #0:1: Video: png, 90k tbr, 90k tbn, 90k tbc'

I think the "video" is the album artwork.  Anyway, this causes ffmpeg_reader to think it's a video file instead of audio, and try to extract video dimensions.  Needless to say, those don't exist, and it throws an exception when it can't find them.

I am only interested in the audio data, not the png, whatever that is.  Is there a way to make ffmpeg_reader a little smarter about what kind of file it's looking at?

Zulko:
This case is supposed to work. What's the error message exactly ?

ryanfox:
MoviePy error: failed to read video dimensions in file 'myaudio.mp3'.
    Here are the file infos returned by ffmpeg:

And then a list of the information returned by ffmpeg.  I don't have the exact information available right now, I can get that later if it would be useful.

Zulko:
Can you send a link to a file that doesn't work ?

ryanfox:
Here it is: http://traffic.libsyn.com/geeknights/20150414.mp3  Warning: it's a 64mb file.

Also, I looked at the console printout a little more carefully.  It prints an error about "failed to read video dimensions" but doesn't crash there.  Later on, in line 305:

    float(line[match.start():match.end()].split(' ')[1])

`match` is None, so that line throws an error.
This is the value in `line`:

    u'    Stream #0:1: Video: png, 90k tbr, 90k tbn, 90k tbc'

kerier:
I have encounter the same issue, AudioFileClip can't read mp3 file, but when i put a video file, it works fine! It's a very weird problem.

here is my code:
audio = AudioFileClip("test.mp3")

here is the except:
Traceback (most recent call last):
  File "testanimation.py", line 5, in <module>
    audio = AudioFileClip("test.mp3")
  File "/usr/local/lib/python2.7/dist-packages/moviepy-0.2.2.11-py2.7.egg/moviepy/audio/io/AudioFileClip.py", line 63, in __init__
[...]

PR #188, opened 2015-08-04
added check for resolution before processing video stream

ryanfox:
Fixes https://github.com/Zulko/moviepy/issues/158

A simple fix, and not perhaps the most efficient, but prevents errors for me.
I added a check that there is resolution info attached to a video stream before attempting to process it.


Zulko:
Thanks ! Trusting you on this ! (I still don't have time to make a test suite)

PR #197, opened 2015-08-24
Fixed copy-paste typo

temerick:
Changed documentation for the color parameter to distinguish from bg_color


earney:
Is this pull request accurate in fixing a typo in documentation?   (ie,  Does TextClip.list('color') list the available background colors, or does it list the available colors of the text?

If it lists the background colors, then this PR is invalid, but if it lists the colors of the text, then this PR needs to be applied to fix the error in documentation.

PR #211, opened 2015-09-20
Fixed missing list

LunarLanding:
Otherwise, it attempts to iterate over int's.
(using python 3)


earney:
committing since PR looks valid.   enumerate([wf,hf]) is valid vs enumerate(wf,hf)  wf is a width of a image, so this should cause an error, while enumerate([wf, hf]) will not.   I'm not 100% sure that the logic is correct, but what is currently there will cause a TypeError in Python3

PR #215, opened 2015-10-07
fixed module hierarchy for Trajectory

bwagner:


earney:
this pull request make two changes.  one to addy which seems valid to me, and the second change modifies a comment and corrects an import path.  PR seems valid.  looking for further reviews before merging into repo.

earney:
no response from other reviewers, so pushing this PR from Oct, 2015.

Issue #227, opened 2015-11-04
CompositeVideoClip produces bad output with use_bgclip and a transparent background clip


Humper:
If use_bgclip is true, the background clip is removed from self.clips.

However, the mask for the final composite clip is the composition of self.clips.

This means that the background clip's mask is not included in the final clip's composite mask, yielding bad results.

This can be fixed by simply iterating "for c in clips" instead of "for c in self.clips" when computing the set of mask clips to composite (around line 99 or CompositeVideoClip.py).

PR #228, opened 2015-11-04
fix bug with transparent background clips

Humper:
Fixes bug #227

### DIFF
--- a/moviepy/video/compositing/CompositeVideoClip.py
+++ b/moviepy/video/compositing/CompositeVideoClip.py
@@ -96,7 +96,7 @@ class CompositeVideoClip(VideoClip):
         if transparent:
             maskclips = [(c.mask if (c.mask is not None) else
                           c.add_mask().mask).set_pos(c.pos)
-                          for c in self.clips]
+                          for c in clips]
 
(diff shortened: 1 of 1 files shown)


earney:
@Humper  can you give us an example where this fails before your PR?   I've tried a few examples and those examples work for me.

earney:
@Humper    Here is the example I'm trying to use to duplicate your issue (#227) and verify your PR (#228).

```
green = ColorClip((800, 600), color=(0,255,0))
    green.mask = ColorClip((800,600), color=1.0, ismask=True)
    red = ColorClip((400, 300), color = (255,0,0), duration=1)
    blue = ColorClip((400, 300), color = (0,0,255), duration = 2)
    clip=CompositeVideoClip([green, blue, red], use_bgclip=True).set_duration(3)
    clip.write_videofile("/tmp/issue_227.mp4", fps=30)
```
This seems to work the same way before and after the PR.  What am I overlooking?  Can you fix my example, or can you supply one of your own?
Thanks!

frazras:
Try using a Textclip in your example, that fails for me

mondeja:
This was addressed [here](https://github.com/Zulko/moviepy/commit/fdc51d29b7ce2f3b2c4d0d907c1c62b5da168128).

PR #236, opened 2015-12-11
Document inherited members in reference documentation

achalddave:
This should make it significantly easier to skim the reference for a class and view all its methods and attributes. Other projects, such as [flask](https://raw.githubusercontent.com/mitsuhiko/flask/7f3867491570746a4c14bdaa5bd59ec1b64cbfea/docs/api.rst), also do this.


earney:
I think this is definitely a plus.  Any objections?

Zulko:
Agreed.

PR #242, opened 2015-12-22
fixed ffmpeg error handling

carlodri:
`proc.stderr.read()` needs to be `.decode`d, otherwise the subsequent `in` operator with a string will cause an `AttributeError`

Using Python 3


carlodri:
ok, I realized now that this would conflict with pull request #179. Honestly, I don't know which is the best approach.

carlodri:
#179 is definitely better. Closing this one.

PR #243, opened 2015-12-23
small recipe (mirroring a video)

zodman:


earney:
Thanks for the contribution!.. Sorry it took so long to merge in..

Zulko:
Sorry I never commented this one before, but I object to it. The code is for command-line (Id like to avoid that in the docs), the use or variables is not ideal (clip gets overwritten), and there is a syntax error on the last line.

earney:
okay..   since it was just documentation, and the code looked correct (I didn't see the syntax error), I saw no harm in merging it.   I'll start waiting for comments before merging in the future.

Zulko:
There was indeed no harm, but I am a bit picky, and since I had 3 objections for just 5 lines of code, I'd prefer not to merge this one.

kerstin:
I suggested e-mail, but could we otherwise maybe create one (new) thread (issue) here to lay out some ground rules or workflow ideas? (I'm just thinking that might save everyone some time in the long run.)

PR #251, opened 2016-01-23
fixes #248 issue with VideoFileClip() not reading all frames

### LINKED ISSUE #248: Problem with VideoFileClip not loading all frames
I am trying to run this code with this [large.gif](http://i.imgur.com/hlN3NZZ.gif):

```
clip = VideoFileClip('large.gif')
clip.write_gif('new.gif')
```

The result is [new.gif](http://i.imgur.com/vYKrAMD.gif). As you can see, for some reason, new.gif is a lot shorter than large.gif. I tried to count the number of frames being loaded using:

```
i=0
    for frame in clip.iter_frames():
        print(i)
        i+=1
```

The number of frames varied but it was generally around 140 frames while the original large.gif has around 375 frames.

Did anyone else run into the same problem?

aldilaff:


MicahRamirez:
Why -1 ?

aldilaff:
The issue I was facing is that moviepy was not reading the data of large gifs correctly. Looking at the code, the line I changed used to read the first element of a list. I found out that for large gifs, the data is being read incrementally and added to the list. So the element with the "full" length of large gifs was actually the last element. -1 is just the last element of a list.

earney:
@aldilaff   can you change this so that it reads something like:

```
if is_Gif:
   line = [l for l in lines if keyword in l][-1]
else:
  line = [l for l in lines if keyword in l][0]
```

earney:
@aldilaff   I'll give you a few more days to respond, if not, I'll create another pull request that makes the necessary changes I have suggested, and will close this PR.

aldilaff:
@earney My apologies for the delay. Updated.

Gloin1313:
Maybe there should be a comment in there explaining why, or at least linking to this PR?

earney:
@achalddave   Gloin1313 is probably right..  Can you add the following comment to your submission?

#for large GIFS the "full" length is presented as the last element in the output.

Thanks!

moviepy/video/io/ffmpeg_reader.py
```diff
@@ -261,7 +261,10 @@ def ffmpeg_parse_infos(filename, print_infos=False, check_duration=True):
     if check_duration:
         try:
             keyword = ('frame=' if is_GIF else 'Duration: ')
-            line = [l for l in lines if keyword in l][0]
+            if is_GIF:
+                line = [l for l in lines if keyword in l][-1]
+            else:
+                line = [l for l in lines if keyword in l][0]
```
mbeacom: Instead of writing out the `line =` twice, we could probably replace L264-L267 with something like:

```python
index = -1 if is_GIF else 0
line = [l for l in lines if keyword in l][index]
```
Granted, there is a lot we could refactor in there to get more DRY, but you get the point!

Thanks for the PR, btw!

earney:
@achalddave   thanks for updating this..  We will try to merge it ASAP..

achalddave:
Hmm, I assume I'm being mentioned here due to a typo? I think you want @aldilaff :)

PR #268, opened 2016-02-19
Updated resize function to use cv2.INTER_LINEAR when upsizing images …

kuchi:
…& cv2 is available (downsizing still uses cv2.INTER_AREA). This gives better image quality while still being about as fast.

### DIFF
--- a/moviepy/video/fx/resize.py
+++ b/moviepy/video/fx/resize.py
@@ -7,8 +7,14 @@ try:
     import numpy as np
     def resizer (pic, newsize):
         lx, ly = int(newsize[0]), int(newsize[1])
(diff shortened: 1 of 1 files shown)


Zulko:
Thanks, do you have a reference link or image examples to document the issue ?

kuchi:
I couldn't find a good companion of the different opencv interpolation algorithms so I wrote up a little comparison of the algorithms showing the image quality and the computation time:
http://tanbakuchi.com/posts/comparison-of-openv-interpolation-algorithms/

Hope this shows the difference. Mainly nicer not to have jagged edges when upsizing the images. But the inter_area is better for downsizing.

Zulko:
Thanks, that's great !

Issue #274, opened 2016-03-14
Fails to list fx after freezing an app with moviepy


nad2000:
Import of import **moviepy.video.fx.all as vfx** since it's based on directly listing files in the package:

```
C:\Users\nad2000\IM>C:\Users\nad2000\im\build\exe.win32-2.7\clip.exe -P %PW% -s "1,13.4" -e "1,13.9" -H 0.5 --x1=145 --x2=400 %VIDEO% kris_sven.gif
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\cx_Freeze\initscripts\Console.py", line 27, in <module>
    exec(code, m.__dict__)
  File "clip.py", line 8, in <module>
    from moviepy.editor import VideoFileClip
  File "C:\Python27\lib\site-packages\moviepy\editor.py", line 34, in <module>
    import moviepy.video.fx.all as vfx
  File "C:\Python27\lib\site-packages\moviepy\video\fx\all\__init__.py", line 16, in <module>
    _files = os.listdir(_directory)
WindowsError: [Error 3] The system cannot find the path specified: 'C:\\Users\\nad2000\\im\\build\\exe.win32-2.7\\library.zip\\moviepy\\video\\fx/*.*'
```

kerstin:
Linking #263 as it looks like it's about the same error.

bw4sz:
Is the #275 supposed to fix this error? I have the same issue, should I pip install from git branch?

PR #275, opened 2016-03-15
Fix for #274 

nad2000:
Fix for #274 

the list of modules build using the built in **pkgutil**, that support zipped/egged libs/modules.


earney:
I've ran a few test cases, and this seems to work okay.  Since it is a cleaner solution, I'm going to merge it with the repo.

PR #295, opened 2016-05-16
Fix for python 3

mikhuang:
Fix for `TypeError: 'str' does not support the buffer interface` when there's an ffmpeg error


mbeacom:
@mikhuang Thank you for your contribution! This looks like a good solution to resolve the exception, but I'm going to move ahead with #403, as it is addressing the py3 byte issue.

PR #301, opened 2016-06-11
change deprecated ToString function to ToBytes in resize

soops:


pydo:
This should get accepted soon @Zulko . Currently resizing with pillow is broken...

soops:
I've found a temporary fix, you can downgrade Pillow to an earlier version
and it will work. As I am away from my laptop, I don't remember which
version, but trial and error should work it all out.
-- 
- Ethan

Zulko:
I'm not sure I understand this commit, you are replacing a commented line with another commented line?

pydo:
@Zulko it is uncommented in the current release on pipy. Not sure why it is commented now. But either way resizing with pillow >= 3.0 is broken since `tostring()` has been deprecated.

CrankyYankee:
Folks, any idea when this will be merged?  I need resize.

craiglytle:
Just confirming that this pull works with the latest Pillow (with the tostring() commented out).  Thanks! 
But the current release available via pip doesn't include this change and resize still fails with Pillow>=3.0.
Any plans to add this change to the current release?
(By the way, moviepy is awesome)

PR #302, opened 2016-06-16
fix deprecation message

mgaitan:
currently, the docstring is nonsense. 

```
The function ``concatenate_videoclips`` is deprecated and is kept temporarily for backwards compatibility.
Please use the new name, ``concatenate_videoclips``, instead.
```


Gloin1313:
The change is already implemented

Gloin1313:
Oops

PR #315, opened 2016-09-21
Memory compress

AaronYALai:


earney:
this PR relies on scipy, which only supports python 2.7 and 3.4+.  is that okay with people, or do we need to support lower versions of 3.x ?

earney:
I pulled this into my local repo and all the examples work (except for test_ImageSequenceClips, and that one doesn't produce correct output in the master repo either..)..

Zulko:
Sorry for not answering before but i am hold back b the sparse matrices and scipy dependancy. It looks like a narrow/specific usecase.

earney:
@Zulko   I agree,  I'd rather keep the dependencies as minimal as possible.    I don't know much about sparse.csr_matrix.  How hard would it be to remove that dependency from this PR?

earney:
@Zulko   should we reject this PR since it requires scipy?

tburrows13:
Closing due to significantly increasing the complexity of the code, and increasing the project dependencies for little benefit. Also, there are a significant number of merge conflicts because it is so old.

PR #316, opened 2016-09-23
Changed .tostring() over to .tobyte()

I2obiN:
Running moviepy on debian with my own compiled ffmpeg and python 2.7.12

Have tested it and moviepy successfully exported the composite video with the new code.
I was taking a Simpsons clip and simply putting it in a clips_array x3, then resizing the array before writing the file.

`from moviepy.editor import *`
`lennyclip = VideoFileClip("./lenny.mp4").margin(10)`
`memes = clips_array([[lennyclip, lennyclip, lennyclip]])`
`memes.resize(width=480).write_videofile("lennymeme.mp4")`

Can confirm above works with tobytes() at least


earney:
@I2obiN ,  this seems to work with tostring.   why should we change to tobyte?    Does it have something to do with python version 3.x ?

PR #328, opened 2016-10-12
FIX Ffmpeg writer error display in py3

caspervdw:
This fixes a `TypeError` due to the bytes <-> unicode transition in Python 3, at the error display of the ffmpeg writer.


jeromegrosse:
Duplciate of https://github.com/Zulko/moviepy/pull/295

PR #339, opened 2016-10-17
Add a test case.

drewm1980:
On my Ubuntu 14.04 machine, TextClip crashes in ImageMagick's convert. Perhaps this is a permissions issue?

```
E           IOError: MoviePy Error: creation of None failed because of the following error:
E           
E           convert.im6: not authorized `@/tmp/tmpXz0_GE.txt' @ error/property.c/InterpretImageProperties/2959.
E           .
E           
E           .This error can be due to the fact that ImageMagick is not installed on your computer, or (for Windows users) that you didn't specify the path to the ImageMagick binary in file conf.py, or.that the path you specified is incorrect

```

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -1,12 +1,15 @@
 """
-Tests meant to be run with pytest
+To run the tests:
+    py.test tests.py
 """
-
-import pytest
 
 from moviepy.editor import *
 
-
-@pytest.fixture
-def example_video1():
-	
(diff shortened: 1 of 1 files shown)


earney:
will merge this since it only modifies a file not in use at this time,  will move the contents of test.py to test_issues.py as function test_issue_xxx  where xxx is the issue number.

Issue #344, opened 2016-10-25
AAC support for mp4


jeromegrosse:
In this [pull request](https://github.com/Zulko/moviepy/pull/26), the AAC compatibility was added to MoviePy by passing `-strict -2` to ffmpeg.

Between then and now, some changes have been done. The `-struct -2` is still being passed to ffmpeg but it is not possible to pass the kwarg `audio_codec="aac"` while writing an mp4 to VideoClip without getting the error `The audio_codec you chose is unknown by MoviePy.`

mbeacom:
Resolved in #345.

PR #345, opened 2016-10-25
Added AAC codec to mp4

jeromegrosse:
Fix of the issue #344

### DIFF
--- a/moviepy/tools.py
+++ b/moviepy/tools.py
@@ -137,7 +137,7 @@ def deprecated_version_of(f, oldname, newname=None):
 # Note that 'gif' is complicated to place. From a VideoFileClip point of view,
 # it is a video, but from a HTML5 point of view, it is an image.
 
-extensions_dict = { "mp4":  {'type':'video', 'codec':['libx264','libmpeg4']},
+extensions_dict = { "mp4":  {'type':'video', 'codec':['libx264','libmpeg4', 'aac']},
                     'ogv':  {'type':'video', 'codec':['libtheora']},
                     'webm': {'type':'video', 'codec':['libvpx']},
                     'avi':  {'type':'video'},


mbeacom:
@jeromegrosse Thanks for the PR.

[AAC](https://trac.ffmpeg.org/wiki/Encode/AAC) is an formatting/codec type for Audio (this extensions dict defines `video` codec types.  Just looking for some clarification.
What issue does this resolve?

ping:
I'm not the OP, but the problem is specified in #344
Without this PR, I cannot specify aac as the audio codec in ``write_videofile()``

```python
vidclip.write_videofile(output_file, codec='libx264', audio=True, audio_codec='aac')
```

earney:
@mbeacom,  should we merge this pull request or should we look at adding an audio_codec to the dictionary?

jeromegrosse:
@mbeacom
The `extention_dict` dictionnary is used by [find_extention](https://github.com/jeromegrosse/moviepy/blob/00e911a617eaff63f8e87c6e069d087a8ff6186e/moviepy/tools.py#L155) which is itself used by [write_videofile](https://github.com/jeromegrosse/moviepy/blob/00e911a617eaff63f8e87c6e069d087a8ff6186e/moviepy/video/VideoClip.py#L309) when looking for the right codec to use for the audio file. This PR allows the possibility to use AAC and MP4.

The way I understood it, `extention_dict` is used to instinctively look up the audio and video codecs. Maybe it is semantically wrong though, and we should use two different dictionary, one for video codecs and one for audio codecs?

mbeacom:
@jeromegrosse From the look of it, we should probably split up the dictionaries.  This will require additional refactoring throughout the codebase to split up the dictionaries. @earney Maybe we merge this and open a separate issue to track potentially splitting up the codec dictionaries?

earney:
will do.

PR #355, opened 2016-11-29
Fix issue #354

### LINKED ISSUE #354: crossfadeout "Attribute 'duration' not set"
When trying to crossfadeout a clip with defined duration `ValueError: Attribute 'duration' not set ` raised.

groundflyer:
This should fix #354

### DIFF
--- a/moviepy/video/compositing/transitions.py
+++ b/moviepy/video/compositing/transitions.py
@@ -25,6 +25,7 @@ def crossfadeout(clip, duration):
     """ Makes the clip disappear progressively, over ``duration`` seconds.
     Only works when the clip is included in a CompositeVideoClip.
     """
+    clip.mask.duration = clip.duration
     newclip = clip.copy()
     newclip.mask = clip.mask.fx(fadeout, duration)
     return newclip


earney:
Fellow developers..  I see no difference in the video output when this PR is applied.  
here is the code I'm using..

```
_clip = VideoFileClip("media/knights.mp4")
_clip1=CompositeVideoClip([_clip.set_duration(20)]).crossfadeout(1)
_clip1.write_videofile("test.mp4", codec="mpeg4")
```
Am I overlooking something?  Can someone else replicate my results?

kerstin:
@earney I haven't tested this, but are you saying you're still getting the error originally described by @groundflyer in #354?

earney:
I'm not sure if the fix does anything.  I do not see a difference in the output file "test.mp4".   I thought I'd let @Zulko take a look.  Maybe I'm overlooking something.

kerstin:
@earney I think the commit wasn't supposed to change the output file, but to get rid of an error. If the error can't be reproduced with the current code (minus the fix), maybe the issue is not valid anymore.

earney:
@kerstin,  no I'm not getting the error.

earney:
My bad guys..  I posted the wrong set of code here.  The pull request does solve the issue for the problem in #354 .  should we look at merging this?

earney:
Or is the @required_decorator for the function crossfadeout unnecessary?  crossfadein does not have a @required_decorator ?

earney:
the @required_decorator on crossfadeout isn't necessary.  The duration value error is checked further down in the fadeout function.   But I believe this issue, brings up an issue with that the mask.duration is not being set somewhere.  There are other issues on this board that bring this up, and I fill that this pull request addresses the issue, and not the real problem (ie, the mask.duration is not being set when it should).

kerstin:
@earney At first I thought you were referring to someone because of the @ sign.

Would you consider using backticks for code bits so they can be better differentiated from normal text? I know it's a bit more work to type messages that way but it improves their readability a lot.

earney:
so you mean ```@required_decorator``` vs @required_decorator ?

kerstin:
@earney Yes!! (:

earney:
I've finally figured out what is going on.

Because TextClip inherits from ImageClip, and by default transparent = true,  ImageClip will create a mask at class instantiation.  Since the duration is not known at this time,  the duration is set to none.   There are a few solutions.  One, when someone sets the duration of the clip, we could also set the duration of the mask.  but if the mask gets updated (or replaced) we would need to reset the duration as well.  The second option is to check that the masked duration is correct before we use it, which is similar to what this PR is doing.  What method do people recommend?

earney:
I'm currently leaning toward doing what this pull request does.  In each fx, setting the ``` mask.duration = c.duration ```.   That way, the mask duration is guaranteed to match the clips duration.  if there is no objection, in the next week or so, I'll start working on this.

PR #367, opened 2016-12-19
Fix issue 313

arlyon:
reduce has been moved to functools since python 3


mbeacom:
@arlyon Thank you for your contribution! Ultimately, I merged #410, as it incorporated the py3 check to avoid issues on python 2.7.

Thank you again for the contribution!

Issue #246, opened 2016-01-06
IndexError when converting audio to_soundarray()


pinae:
When converting audio to a numpy array with `.to_soundarray()` I get a `IndexError` with some files. This occured with audion from video clips and with audio clips

The error occurs with this File: https://github.com/pinae/Audiosyncer/blob/master/sample_mono_short.flac
This code reproduces the error:

```
from moviepy.editor import *
sound = AudioFileClip("sample_mono_short.flac")
samples = sound.to_soundarray()
```

The whole error message:

```
IndexError                                Traceback (most recent call last)
/home/xxx/venv/lib/python3.4/site-packages/moviepy/audio/io/readers.py in get_frame(self, tt)
    186                 indices = frames - self.buffer_startframe
--> 187                 result[in_time] = self.buffer[indices]
    188                 return result

IndexError: index 120575 is out of bounds for axis 0 with size 120575

During handling of the above exception, another exception occurred:

OSError                                   Traceback (most recent call last)
<ipython-input-2-7b8ec45b76f0> in <module>()
      2 
      3 sound = AudioFileClip("sample_mono_short.flac")
----> 4 samples = sound.to_soundarray()

<decorator-gen-194> in to_soundarray(self, tt, fps, quantize, nbytes, buffersize)

/home/xxx/venv/lib/python3.4/site-packages/moviepy/decorators.py in requires_duration(f, clip, *a, **k)
     52         raise ValueError("Attribute 'duration' not set")
     53     else:
---> 54         return f(clip, *a, **k)
     55 
     56 

/home/xxx/venv/lib/python3.4/site-packages/moviepy/audio/AudioClip.py in to_soundarray(self, tt, fps, quantize, nbytes, buffersize)
    112             if self.duration>max_duration:
    113                 return stacker(self.iter_chunks(fps=fps, quantize=quantize, nbytes=2,
--> 114                                                  chunksize=buffersize))
    115             else:
    116                 tt = np.arange(0, self.duration, 1.0/fps)

/home/xxx/venv/lib/python3.4/site-packages/numpy/core/shape_base.py in vstack(tup)
    228 
    229     """
--> 230     return _nx.concatenate([atleast_2d(_m) for _m in tup], 0)
    231 
    232 def hstack(tup):

/home/xxx/venv/lib/python3.4/site-packages/numpy/core/shape_base.py in <listcomp>(.0)
    228 
    229     """
[...]

Issue #265, opened 2016-02-16
IndexError: index is out of bounds for axis 0 with size XXXX


fortunto2:
Hi,

Error 
```
OSError: Error in file /ubuntu/conv/150757/conv/baaadilz/vid20160215102221_010.mp4, At time t=0.11-0.15 seconds, indices wanted: 4806-6805, but len(buffer)=5645
index 5645 is out of bounds for axis 0 with size 5645
```
How i can skip like this video file when rendering?
O solve problem...with many same files
```
[MoviePy] >>>> Building video /ubuntu/conv/150757/film/160360/baaadilz.mp4
[MoviePy] Writing audio in baaadilzTEMP_MPY_wvf_snd.mp3
 73%|███████▎  | 464/634 [00:01<00:00, 427.40it/s]Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/moviepy-0.2.2.11-py3.4.egg/moviepy/audio/io/readers.py", line 188, in get_frame
    result[in_time] = self.buffer[indices]
IndexError: index 5645 is out of bounds for axis 0 with size 5645

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./1job.py", line 246, in <module>
    duration = template.fast.render(allParam, data, videos, audios)
  File "/ubuntu/newconv/template/fast.py", line 400, in render
    ffmpeg_params=ffmpeg_params, threads=8, write_logfile=True)  
  File "<decorator-gen-51>", line 2, in write_videofile
  File "/usr/local/lib/python3.4/dist-packages/moviepy-0.2.2.11-py3.4.egg/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, _a, *_k)
  File "<decorator-gen-50>", line 2, in write_videofile
  File "/usr/local/lib/python3.4/dist-packages/moviepy-0.2.2.11-py3.4.egg/moviepy/decorators.py", line 137, in use_clip_fps_by_default
    return f(clip, _new_a, *_new_kw)
  File "<decorator-gen-49>", line 2, in write_videofile
  File "/usr/local/lib/python3.4/dist-packages/moviepy-0.2.2.11-py3.4.egg/moviepy/decorators.py", line 22, in convert_masks_to_RGB
    return f(clip, _a, *_k)
  File "/usr/local/lib/python3.4/dist-packages/moviepy-0.2.2.11-py3.4.egg/moviepy/video/VideoClip.py", line 330, in write_videofile
[...]

PR #377, opened 2017-01-03
Audio error handling made failsafe

gyglim:
This PR is to improve audio reading robustness.

In case there are audio frames missing, this repeats the last frame, as is done in the case for video frames:
https://github.com/Zulko/moviepy/blob/master/moviepy/video/io/ffmpeg_reader.py#L109

Thus it makes the code more consistent and robust.
Fixing https://github.com/Zulko/moviepy/issues/246

Cheers,
Michael


moviepy/video/io/ffmpeg_reader.py
```diff
@@ -8,9 +8,6 @@
 import subprocess as sp
 import re
 import warnings
-import logging
-logging.captureWarnings(True)
```
gyglim: This would lead to the warning not being shown. Not sure what the logic is here? Does the code require to instantiate a logger now?

gyglim:
The PR now only shows a warning when not enough frames can be read (happens when there are less frames than should be, according to fps*duration). Because of that, it was possible to go even out of the estimated valid range. Thus, it now needs a check that the required times are valid according to the estimated range, as otherwise the list of frames to be read is empty in https://github.com/gyglim/moviepy/blob/68df86c880ec1df0f9f3e0d4b36efa02f8c2b8eb/moviepy/audio/io/readers.py#L173

gyglim:
Hey. Just added a test for this. Is there anything else needed to merge it? @earney

earney:
@gyglim. Thanks for the pr
  Lets see what some of the other project contributors thinks.  @zulko. Is it ok to replace the error message with this code?

gyglim:
New issues on the same:
https://github.com/Zulko/moviepy/issues/265
https://github.com/Zulko/moviepy/issues/521

@Zulko Any opposition against merging this PR? Anything I can do to improve it?

Zulko:
Sorry i thought this was merged. Id say yes to this pr. It's a bit of a hack, but it wont be the first in moviepy, and if it solves that audio problem it's good for me.

gyglim:
Ok, great :). I agree, but it makes sense and is consistent with what happens for video

earney:
@gyglim   lets merge this in.. unless someone objects?

gyglim:
@earney Yes, let's do it

Issue #376, opened 2017-01-02
TypeError: 'float' object cannot be interpreted as an integer


WDBell:
I'm new to moviepy.

`clip = VideoFileClip("Rough Guide to Dublin (54).mp4")`

I tried three mp4 files, all with the same result. Fuller information:

```
>>> clip = VideoFileClip("Sailing Mediterranean - 7 Sardinia.mp4")
Traceback (most recent call last):
  File "<interactive input>", line 1, in <module>
  File "C:\Python34\lib\site-packages\moviepy\video\io\VideoFileClip.py", line 82, in __init__
    nbytes = audio_nbytes)
  File "C:\Python34\lib\site-packages\moviepy\audio\io\AudioFileClip.py", line 63, in __init__
    buffersize=buffersize)
  File "C:\Python34\lib\site-packages\moviepy\audio\io\readers.py", line 70, in __init__
    self.buffer_around(1)
  File "C:\Python34\lib\site-packages\moviepy\audio\io\readers.py", line 234, in buffer_around
    self.buffer =  self.read_chunk(self.buffersize)
  File "C:\Python34\lib\site-packages\moviepy\audio\io\readers.py", line 123, in read_chunk
    self.nchannels))
TypeError: 'float' object cannot be interpreted as an integer
```

Suggest this change to code (and hope that this is a reasonable place to post this). In `readers.py`, substitute the second `result =` statement.

```
    def read_chunk(self,chunksize):
        #~ chunksize = int(round(chunksize))
        L = self.nchannels*chunksize*self.nbytes
        s = self.proc.stdout.read(L)
        dt = {1: 'int8',2:'int16',4:'int32'}[self.nbytes]
        result = np.fromstring(s, dtype=dt)
        #~ result = (1.0*result / 2**(8*self.nbytes-1)).\
                                 #~ reshape((len(result)/self.nchannels,
                                          #~ self.nchannels))

        result = (1.0*result / 2**(8*self.nbytes-1)).\
                                 reshape((int(len(result)/self.nchannels),
                                          self.nchannels))

        #self.proc.stdout.flush()
        self.pos = self.pos+chunksize
        return result

```

adelavega:
I'm having the same problem. Looks like reshape is expecting an integer, but it gets a float because Python 3 does not auto convert floats to integers in division unless you use `//`. 

I may open a PR

mbeacom:
This will be resolved when #384 is merged.

laventura:
[...]

Issue #279, opened 2016-03-17
TypeError in Adding Soundtrack


MaxLFarrell:
When running this line:
`movie = VideoFileClip(image).set_audio(AudioFileClip(audio))`
I get the following error:

```
File "C:\Python34\lib\site-packages\moviepy\audio\io\AudioFileClip.py", line 63, in __init__
    buffersize=buffersize)
  File "C:\Python34\lib\site-packages\moviepy\audio\io\readers.py", line 70, in __init__
    self.buffer_around(1)
  File "C:\Python34\lib\site-packages\moviepy\audio\io\readers.py", line 234, in buffer_around
    self.buffer =  self.read_chunk(self.buffersize)
  File "C:\Python34\lib\site-packages\moviepy\audio\io\readers.py", line 123, in read_chunk
    self.nchannels))
TypeError: 'float' object cannot be interpreted as an integer
```

The audio file is a 1411 kbps wav.

Zulko:
Does it work if you first transcode your file into some other format using e.g. ffmpeg ?

PR #384, opened 2017-01-16
ensures int arguments to np.reshape; closes #383

tyarkoni:
See #383 for discussion. This just patches `readers.py` to ensure compatibility with numpy 1.12.0.

PR #386, opened 2017-01-18
Windows Python3 incomatability fix in np reshape expecting float

adelavega:
In #279 and #376, two users had issues in Python3 on Windows reading audio. Looks like this is due to np reshape expecting an int but getting a float, due to python 3 not "flooring" float division with a single `/` operator. 

I had the same error and fixed it by changing the operator to `//`. Some minor testing suggests the fix works and I don't immediately see any downstream issues from this change.


adelavega:
Oops! Duplicate of #384

PR #390, opened 2017-01-25
bug presented

TonyChen0724:
This is a bug I encountered after I try the example in readme file, I think others might encounter it in the future


earney:
Any reason why we should not merge this PR?  I believe it is a much clearer description then what we current have for finding ImageMagick.  If we replace ImageMagick, then we will just have to remove this portion of the readme.rst, but at this point we are not sure how long it will take to replace ImageMagick.

Issue #373, opened 2016-12-26
manual_tracking format issue


apiszcz:
Possible format issue in interpolators.py

```
  File "tracktest.py", line 5, in <module>
    trajectories = manual_tracking(clip, t1=5, t2=8, fps=1, nobjects=1, savefile="track.txt")
  File "<decorator-gen-101>", line 2, in manual_tracking
  File "lib\moviepy\decorators.py", line 89, in wrapper
    return f(*new_a, **new_kw)
  File "<decorator-gen-100>", line 2, in manual_tracking
  File "lib\moviepy\decorators.py", line 137, in use_clip_fps_by_default
    return f(clip, *new_a, **new_kw)
  File "lib\moviepy\video\tools\tracking.py", line 132, in manual_tracking
    Trajectory.save_list(result, savefile)
  File "lib\moviepy\video\tools\interpolators.py", line 68, in save_list
    header = "\t".join(N*['t (ms)', 'x', 'y']))
  File "lib\numpy\lib\npyio.py", line 1162, in savetxt
    % (str(X.dtype), format))
TypeError: Mismatch between array dtype ('object') and format specifier ('%d')
```

dermusikman:
I have a similar problem:
`>>> tracking.manual_tracking(clip.subclip(0,4.5), nobjects=2, savefile="left-right-hat.txt")
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/numpy/lib/npyio.py", line 1158, in savetxt
    fh.write(asbytes(format % tuple(row) + newline))
TypeError: %d format: a number is required, not zip

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<decorator-gen-212>", line 2, in manual_tracking
  File "/usr/lib/python3.5/site-packages/moviepy/decorators.py", line 89, in wrapper
    return f(*new_a, **new_kw)
  File "<decorator-gen-211>", line 2, in manual_tracking
  File "/usr/lib/python3.5/site-packages/moviepy/decorators.py", line 137, in use_clip_fps_by_default
    return f(clip, *new_a, **new_kw)
  File "/usr/lib/python3.5/site-packages/moviepy/video/tools/tracking.py", line 132, in manual_tracking
    Trajectory.save_list(result, savefile)
  File "/usr/lib/python3.5/site-packages/moviepy/video/tools/interpolators.py", line 68, in save_list
[...]

PR #394, opened 2017-01-29
Fix Issue #373 Trajectory.save_list

dermusikman:
This PR fixes the issue reported in #373 (confirmed on Python 3.6).


earney:
just ran a simple test, and for python 3, we need this PR for a trajectory.save_list call to work correctly..  Merging!

earney:
I will create a PR that contains a test to show that this PR #394 is correct.  Will post it shortly.

Issue #401, opened 2017-02-08
Unable to run the sample code! Getting errors. Plase Help 


batmanrising:
Hi there, I tried to run the sample/example code for the moviePy but I'm getting this error. I don't know how to resolve. I'm an novice in the field so don't have an experience like you guys.

from moviepy.editor import *
clip = VideoFileClip("myHolidays.mp4").subclip(5,15)
clip = clip.volumex(0.8)
txt_clip = TextClip("My Holidays 2013",fontsize=70,color='white')
txt_clip = txt_clip.set_pos('center').set_duration(10)
video = CompositeVideoClip([clip, txt_clip])
video.write_videofile("myHolidays_edited.avi")

Error message: 

[MoviePy] This command returned an error !
Traceback (most recent call last):
  File "/home/pi/clip_example.py", line 11, in <module>
    txt_clip = TextClip("My Holidays 2013",fontsize=70,color='white')
  File "/usr/local/lib/python2.7/dist-packages/moviepy/video/VideoClip.py", line 1173, in __init__
    raise IOError(error)
IOError: MoviePy Error: creation of None failed because of the following error:

convert: not authorized `@/tmp/tmpIJUQvL.txt' @ error/property.c/InterpretImageProperties/3405.
convert: no images defined `PNG32:/tmp/tmpt9CCfR.png' @ error/convert.c/ConvertImageCommand/3210.
.

.This error can be due to the fact that ImageMagick is not installed on your computer, or (for Windows users) that you didn't specify the path to the ImageMagick binary in file conf.py, or.that the path you specified is incorrect

mbeacom:
@bhavin20feb You'll need to install imagemagick.  If you're on OSX, you can install it via `macports` or `brew`.  Also, the code you're referencing is merely an example.  You will need to modify the `myHolidays.mp4` filename reference to an existing MP4 file for this example to work.
Let me know if that resolves your issue!

bhavin20feb:
>Let me know if that resolves your issue!
Hi, @mbeacom thanks a lot for responding on my post. I'm on Raspbian Jessie. 
I have already installed FFmpeg, imagemagic, as well as moviepy. But I'm still getting this error. 
Also, before executing the code I renamed one mp4 video to myHolidays.mp4

I also tried executing this code,
from moviepy.editor import *
video = VideoFileClip("vidnew.mpg").subclip(5,15)
video.write_videofile("vidnew_edited.mpg",fps=24, codec='mpeg4')
[...]

Issue #335, opened 2016-10-14
Bug in ffmpeg_audiowriter.py for python 3


waylander47:
Hi, 

In ffmpeg_audiowriter.py if there is a codec error that raise an exception, in the exception another exception arrises because self.proc.stderr.read() return bytes in python 3 so it needs .decode('utf-8') to be happened in the error message line 91 like that : ffmpeg_error.decode('utf-8')

Best regards,
Waylander.

mbeacom:
Fixed in #403

PR #403, opened 2017-02-10
fix issue #401

earney:
this fixes a python 3 error dealing with strings and bytes.   See issue #401 for more details of this error.

### DIFF
--- a/moviepy/audio/io/ffmpeg_audiowriter.py
+++ b/moviepy/audio/io/ffmpeg_audiowriter.py
@@ -2,6 +2,9 @@
 import numpy as np
 import subprocess as sp
 
+import sys
+PY3=sys.version_info.major >= 3
+
 import os
 try:
     from subprocess import DEVNULL # py3k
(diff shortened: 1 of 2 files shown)


Gloin1313:
This does the same as https://github.com/Zulko/moviepy/pull/295/files ?

earney:
Gloin1313..    I didn't see  pull 295 before I created my version.  Sorry.   I like my solution better, but I'll let the maintainers to choose what version they like best. :)

Gloin1313:
I'm not experienced so I can't comment which I like better. I'm not even sure the other one works on Python 2, but obviously it works fine with Python 3.

earney:
Looking at this more closely, there is a difference..   
Line 85, will also through an error on python 3

   self.proc.stdin.write(frames_array.tostring())
(for python 3) needs to be
   self.proc.stdin.write(frames_array.tobytes())

earney:
I tried my version on python 2 and python 3 and it seems to work.  At least
for the couple of test cases I used.

Gloin1313:
The tobytes()/tostring() difference causes an imageio crash from somewhere else when I run moviepy until I fix it in the code. However, this difference makes no difference either way.

earney:
that is funny.. I don't get the crash..

I'm running Ubuntu 16.10
My version of imageio  is 2.1.2   (for python 3 and for python 2).

What are your versions?

earney:
@Zulko  @mbeacom   anyway of getting this merged in?
Thanks!
Billy

mbeacom:
@earney Thanks for the contribution!  I prefer your approach to this. 
However, It appears this might be an issue in the [ffmpeg_audiowriter](https://github.com/Zulko/moviepy/blob/master/moviepy/audio/io/ffmpeg_audiowriter.py#L92) as well based on the issue #335 and the resolution presented in #295.

I believe this is the appropriate way to go and it resolves #401. We'll want to resolve #335 before pushing a release.

earney:
Thanks. I will look into the fixing the other issue you mentioned


On Feb 14, 2017 5:53 PM, "Mark Beacom" <notifications@github.com> wrote:

*@mbeacom* approved this pull request.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Zulko/moviepy/pull/403#pullrequestreview-21895397>, or mute
the thread

Gloin1313:
It should be extremely easy to add this to ffmpeg_audiowriter as well

earney:
@mbeacom now, this pull also fixes issue #335. :)

PR #405, opened 2017-02-14
Add `self.filename` as a `VideoFileClip` attribute

Gloin1313:
Docstring on lines 39-40 state that `VideoFileClip` has an object `self.filename`. This is not currently correct, as only its `reader` has that attribute. This PR adds `filename` as an attribute of the actual clip as well.
Current usage:
`my_clip.reader.filename``
Proposed usage:
`my_clip.filename

Also added line 45 to explain that a `VideoFileClip` does actually have more than 2 useful attributes.


Zulko:
Thanks !

Issue #313, opened 2016-09-10
concatenate.py - Python3 incompatible


AlexanderNickolsky:
concatenate_videoclips function uses reduce, which is changed in Python 3.
The fix is to use functools:

clips = functools.reduce(lambda x, y: x + y, l) + [clips[-1]]

arlyon:
Solution is to import reduce from functools.

`from functools import reduce`

mbeacom:
Resolved with the merger of #410.

PR #410, opened 2017-02-15
fix issue #313, make concatenate_videoclips python 3 compatible.

earney:


earney:
@mbeacom ..  I was starting to think the same thing..   I'll work on creating a compat.py file with contents:

import sys
PY3 = sys.version_info.major >=3

then I can look at replacing all the updates I've made with

from compat import PY3

where do you recommend that we place the compat.py file ?  maybe moviepy/compat.py, right next to config.py ?

earney:
Can this issue be closed?

PR #411, opened 2017-02-15
move PY3 to new compat.py file

earney:


Gloin1313:
Good job! Shouldn't the files change in [#403](https://github.com/Zulko/moviepy/pull/403/files) be updated as part of this PR?

Also, just a small thing: should we maybe check for python 2 instead of 3? Just that when python 4 comes out (which probably will still be backwards compatible with 3), all these checks will need to be updated. I know that that is looking _really_ far into the future, but it seems that we might as well put this in now...?

earney:
Yes, that is something we need to think about.  Also, how long do we plan on supporting python 2?

Zulko:
@earney: please leave some time before a PR and a merge ! This one changes the API and should therefore be discussed more.

earney:
okay.. will do.

PR #413, opened 2017-02-15
fixes #357.  confusing error about coreader, when media file does not exist

### LINKED ISSUE #357: VideoFileClip instance has no attribute 'coreader'
Hi everyone,
I'm newbie with moviepy and I try to do some example code in documents. But when I run code below
`
"""
Description of the video:
The screen is split in two parts showing Carry and Audrey at the phone,
talking at the same time, because it is actually two scenes of a same
movie put together.
"""

from moviepy.editor import *
from moviepy.video.tools.drawing import color_split

duration = 6 # duration of the final clip

# LOAD THE MAIN SCENE
# this small video contains the two scenes that we will put together.

main_clip = VideoFileClip("../../videos/charadePhone.mp4")
W,H = main_clip.size

# MAKE THE LEFT CLIP : cut, crop, add a mask 
                            
mask = color_split((2*W/3,H),
                   p1=(W/3,H), p2=(2*W/3,0),
                   col1=1, col2=0,
                   grad_width=2)
                   
mask_clip = ImageClip(mask, ismask=True)
                   
clip_left = (main_clip.coreader()
                      .subclip(0,duration)
                      .crop( x1=60, x2=60 + 2*W/3)
                      .set_mask(mask_clip))

# MAKE THE RIGHT CLIP : cut, crop, add a mask 
                   
mask = color_split((2*W/3,H),
                   p1=(2,H), p2=(W/3+2,0),
                   col1=0, col2=1,
                   grad_width=2)

mask_clip = ImageClip(mask, ismask=True)

clip_right = (main_clip.coreader()
                       .subclip(21,21+duration)
                       .crop(x1=70, x2=70+2*W/3)
                       .set_mask(mask_clip))

# ASSEMBLE AND WRITE THE MOVIE TO A FILE

cc = CompositeVideoClip([clip_right.set_pos('right').volumex(0.4),
                         clip_left.set_pos('left').volumex(0.4)],
                         size = (W,H))
#cc.preview()
cc.write_videofile("../../biphone3.avi",fps=24, codec='mpeg4')
`
I got error as: AttributeError: VideoFileClip instance has no attribute 'coreader' could everyone can help me?

earney:
del function creates a messy error that hides the real issue.  The media file does not exist.   I set self.reader = None so that if ffmpeg has issues,  del function will execute without complaining.  Then real error message is easier to read.

### DIFF
--- a/moviepy/video/io/VideoFileClip.py
+++ b/moviepy/video/io/VideoFileClip.py
@@ -54,8 +54,8 @@ class VideoFileClip(VideoClip):
         
         # Make a reader
         pix_fmt= "rgba" if has_mask else "rgb24"
-        reader = FFMPEG_VideoReader(filename, pix_fmt=pix_fmt)
-        self.reader = reader
+        self.reader = None #need this just in case FFMPEG has issues (__del__ complains)
+        self.reader = FFMPEG_VideoReader(filename, pix_fmt=pix_fmt)
         # Make some of the reader's attributes accessible from the clip
         self.duration = self.reader.duration
         self.end = self.reader.duration


Gloin1313:
I've done a different fix that definitely solves some of this one, if not all of it: #419

earney:
@Zulko , @mbeacom   if we can get one or both of you to view Gloin1313's solution and my solution and choose which one we want to use (if we want to use either of them).   This seems to be a very popular issue.  There are 3 or 4 issues that these solutions would fix.

Gloin1313:
We could use both for extra redundancy

earney:
I'm not real crazy about the try/except for this because it could hid some other issue.  I'd rather when it fails , it complains, and it will be much easier to track down that way.

Gloin1313:
@earney You're probably right, but if it is explicitly an `AttributeError`, then the only problem that it doesn't fail on is if `clip.reader` doesn't exist.  I can't really see how this could be a problem.  If `clip.reader` doesn't exist, but is needed somewhere else, then that will cause an error anyway.

earney:
I agree.. that will work for me..  Do you want to modify your pull, or do you want me to modify mine?

Gloin1313:
Why don't we just do both?

earney:
sure.. we can do that too.

Gloin1313:
Sorry, I don't understand. What did you mean by "I agree.. that will work for me.. Do you want to modify your pull, or do you want me to modify mine?"? Modify what changes?

earney:
I was thinking we could merge the two pulls into one pull request, but you
are correct, they can still be separate pulls.

earney:
these pull requests do not affect much, so if I do not get a response from @mbeacom or @Zulko  by Monday morning, I'm going to push this PR.

Gloin1313:
Sounds fine to me, I'll do the same

earney:
I've got no response since Friday, so I'm going to push this PR.

Issue #145, opened 2015-03-12
UnboundLocalError in video/compositing/concatenate.py


mgaitan:
Hey @Zulko, I'm using master ('0.2.2.11') and got this 

```
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-80-263194d08e84> in <module>()
----> 1 supercutter('la gente')

<ipython-input-79-437c51178aac> in supercutter(word, pattern, sub_ext, output, shuffle)
      9     if shuffle:
     10         random.shuffle(subclips)
---> 11     final = concatenate(subclips, method='composite')
     12     final.write_videofile(output, codec='libvpx', fps=24)
     13 

/usr/local/lib/python2.7/dist-packages/moviepy/tools.pyc in fdepr(*a, **kw)
    127     def fdepr(*a, **kw):
    128         warnings.warn("MoviePy: " + warning, PendingDeprecationWarning)
--> 129         return f(*a, **kw)
    130     fdepr.__doc__ = warning
    131 

/usr/local/lib/python2.7/dist-packages/moviepy/video/compositing/concatenate.pyc in concatenate_videoclips(clips, method, transition, bg_color, ismask, padding)
     99                size = (w, h), bg_color=bg_color, ismask=ismask)
    100 
--> 101     result.tt = tt
    102 
    103     result.start_times = tt[:-1]

UnboundLocalError: local variable 'result' referenced before assignment
```

mgaitan:
hey, I've just catch it. I've passed `method='composite'` instead of `method='compose'`. but, shouldn't raise  a more clear error if the given method parameter is unknown?

earney:
a solution for this issue has been merged into the repo.. 
closing this issue.

PR #415, opened 2017-02-15
Issue 145

earney:
it was suggested in issue #145, to output some type of error if the method argument was not valid (ie, 'chain' or 'compose').  The modification allow for an exception to be raised.


Gloin1313:
I'd recommend putting the variable 'transition' into the error message somewhere so the user can know what he/she put in wrong (helpful if the code calling it is heavily automatic). Also, is the a more standard moviepy error message format? ie begins with something like 'Moviepy Error:'

earney:
I've modified the pull so that the function name is retrieved at runtime so that the error message is more descriptive.

earney:
will resubmit this pull request once pull request #413 is complete, and I can refork the zulko/moviepy repo.  I modified the earney/moviepy master branch for pull request #413, instead of creating a special branch.

Issue #407, opened 2017-02-15
VideoClip has no attribute fps error when trying to concatenate


BrianLee608:
Hello, new to moviepy! I ran into an issue when trying to concatenate some mp4 files. 

My code: 

```
dir = '.'
files = glob.glob(os.path.join(dir, '*.mp4'))

clips = [VideoFileClip(f) for f in files]
final_clip = concatenate_videoclips(clips)
final_clip.write_videofile("my_concatenation.mp4")
```

Error: 

> Traceback (most recent call last):
>   File "autocomp.py", line 38, in <module>
>     final_clip.write_videofile("my_concatenation.mp4")
>   File "<decorator-gen-51>", line 2, in write_videofile
>   File "/Users/brianlee/Envs/autocompenv/lib/python3.5/site-packages/moviepy/decorators.py", line 54, in requires_duration
>     return f(clip, *a, **k)
>   File "<decorator-gen-50>", line 2, in write_videofile
>   File "/Users/brianlee/Envs/autocompenv/lib/python3.5/site-packages/moviepy/decorators.py", line 133, in use_clip_fps_by_default
>     for (arg, name) in zip(a, names)]
>   File "/Users/brianlee/Envs/autocompenv/lib/python3.5/site-packages/moviepy/decorators.py", line 133, in <listcomp>
>     for (arg, name) in zip(a, names)]
>   File "/Users/brianlee/Envs/autocompenv/lib/python3.5/site-packages/moviepy/decorators.py", line 122, in fun
>     " the clip's fps with `clip.fps=24`"%f.__name__)
> AttributeError: No 'fps' (frames per second) attribute specified for function write_videofile and the clip has no 'fps' attribute. Either provide e.g. fps=24 in the arguments of the function, or define the clip's fps with `clip.fps=24`

It looks like each clip in [clips] has an fps attribute (I can print them with)...
But when I tried to access the final_clip's fps attribute, it does not have one. 

I have to write
`final_clip.fps = 30`
before calling 
`final_clip.write_videofile("my_concatenation.mp4")`
in order to make it work. Is this a bug?

Zulko:
``concatenate`` will only attribute a fps to the final clip if all clips in the concatenation have the same fps, otherwise it's considered an ambiguous case and the user has to provide the fps. That's implemented in [this line](https://github.com/Zulko/moviepy/blob/master/moviepy/video/compositing/concatenate.py#L112).

[...]

PR #416, opened 2017-02-15
pick highest fps when concatenating

BrianLee608:
These changes resolve issue [407](https://github.com/Zulko/moviepy/issues/407)


Zulko:
This forgets the case where all the fps may be None (this happens if you are concatenating a series of still pictures), at which case the concatenation's fps should be None. Here you'll get a bug with the max. Can we add something like this ?
```
if all([clip.fps is None for clip in clips]):
    self.fps=None
```

BrianLee608:
I think something went weird with my revised PR. 

But I just changed the logic to:
```
fpss = [c.fps for c in clips if hasattr(c,'fps')]
    if len(fpss) == 0:
        result.fps = None
    else:
        result.fps = max(fpss)
```

It avoids looping through the clips twice.

earney:
can we commit this PR?  It passes the 4 or 5 tests, I have created so far.  I also have a test for this issue that needs to be uncommented once this PR is committed.    I'll commit it if no one objects by Wednesday March 1st.

Zulko:
Sure go ahead. If it ever go causes trouble in some scenarios it will be easy to revert/refine later.

Zulko:
Wait actually my remark above hasn't been fully addressed. The line should read like this to take into account that some clips have a fps equal to None:

```
fpss = [c.fps for c in clips if hasattr(c,'fps') and c.fps is not None]
```

PR #418, opened 2017-02-15
adding six for compatibility with py2 in ImageClip

### LINKED ISSUE #417: error 'unicode' object has no attribute 'shape'
i found this error with the python 2

leonardoo:
this fix #417

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -9,6 +9,8 @@ import os
 import subprocess as sp
 import multiprocessing
 import tempfile
+import six
+
 from copy import copy
 
 from tqdm import tqdm
@@ -919,7 +921,7 @@ class ImageClip(VideoClip):
 
         VideoClip.__init__(self, ismask=ismask, duration=duration)
 
-        if isinstance(img, str):
+        if isinstance(img, six.string_types):
             img = imread(img)
 
         if len(img.shape) == 3:  # img is (now) a RGB(a) numpy array
--- a/setup.py
+++ b/setup.py
@@ -16,13 +16,13 @@ except ImportError:
 from setuptools import setup, find_packages
 
 
-#exec(open('moviepy/version.py').read()) # loads __version__
+exec(open('moviepy/version.py').read()) # loads __version__
 
 setup(name='moviepy',
-    version="0.2.2.13",
+    version=__version__,
     author='Zulko 2017',
     description='Video editing with Python',
-    long_description="",
+    long_description=open('README.rst').read(),
     url='http://zulko.github.io/moviepy/',
     license='MIT License',
     keywords="video editing audio compositing ffmpeg",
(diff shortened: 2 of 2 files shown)


Gloin1313:
Is there a way to do this without adding another dependency? I don't think we want to add a whole required module just for one small fix.

earney:
I agree.  I would like to keep the dependencies as minimal as possible.
There are howtos of how to write code that is py2 and py3 compatible.

Zulko:
@Gloin @earney I would also agree with no more dependencies, but here is a good example where ``six`` is the most elegant solution.

Have a look at this stackoverflow thread over this problem. All solutions not using six are much longer and sometimes cryptic.

http://stackoverflow.com/questions/11301138/how-to-check-if-variable-is-string-with-python-2-and-3-compatibility

I've never used six before as I prefer to use compatibilioty recipes, but ``six`` is only 10kb zipped, and has been downloaded more than 100 million times, so we can trust it. As we advance towards python2 and 3 compatibility and without a test suite, it may save us lots of headaches.

mbeacom:
@Zulko @earney @Gloin1313 Maybe we can consider a solution like the [requests package](https://github.com/kennethreitz/requests/blob/master/requests/compat.py)? I was trying to avoid merging too much before we get a base test suite in place.

Zulko:
@mbeacom  Six is basically the same kind of file, but put behind a library, and with a test suite.
Got you with the merging, there is too much enthusiasm and too early, lets take the necessary time to make the library maintainable (test suite, rules, how-tos, etc.).

Are you working on a test suite now or do you want me to start one ?

earney:
I agree with the merging..  we need to create a basic test suite.  And add extra tests as issues arise.    Now, we might be breaking something, when we fix something else.

earney:
PR #121 seems to address some testing issues, and how to set it up with github, etc.

earney:
this is an easy fix, and we don't need six to do it..
Here is the basic code to do it:

change:

if isinstance(img, str):
   img = imread(img)

to
# add from ..compat import PY3    at the top of the file
if PY3:
           if isinstance(img, str):
               img = imread(img)
        else:
           if isinstance(img, (str, unicode,)):
               img = imread(img)

leonardoo:
this will be merge?

earney:
@leonardoo   it seems we have a couple contributors that wants to add six, but we have a couple that do not, so for now, I think we are holding off, until we can figure out how to proceed.   Please keep your Pull Request, and we will let you know ASAP.

Zulko:
I'll fix that one with a merge later on. In that particular instance, we check that the provided image is a filename and not a numpy array. So instead of checking if the image is a string/unicode/bytes... we can just check that it is not a numpy array (``not isinstance(img, np.ndarray)``), which works fine in Py2 and Py3.

earney:
@zulko   I can working on adding the not isinstance(img, np.ndarray), if you are too busy to add it in the next few days.

earney:
the issue resolved in this PR is also fixed in PR #438

earney:
PR #438 has been merged so this PR will not be used.  sorry..

PR #419, opened 2017-02-16
Fix `AttributeError: reader`

Gloin1313:
I get the error

```
Exception ignored in: <bound method VideoFileClip.del of <moviepy.video.io.VideoFileClip.VideoFileClip object at 0x10d347d30>>
    File "/anaconda/lib/python3.5/site-packages/moviepy/video/io/VideoFileClip.py", line 89, in del
      del self.reader
AttributeError: reader
```

When I run:

```
try:
    clip = VideoFileClip("/Users/path/to/movie.mp4")  # When this movie can't be found, it outputs the error above
except OSError:
    print("Video not found")
```

The error doesn't actually stop the program, but it is an annoying output.

### DIFF
--- a/moviepy/video/io/VideoFileClip.py
+++ b/moviepy/video/io/VideoFileClip.py
@@ -40,7 +40,9 @@ class VideoFileClip(VideoClip):
       Name of the original video file.
     
     fps:
-      Frames per second in the original file. 
+      Frames per second in the original file.
+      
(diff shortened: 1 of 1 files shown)


earney:
pull request #413 also addresses this issue.

Gloin1313:
Ah yes, I hadn't worked out what #413 did. We do need one of them merged though.

Gloin1313:
I messed up the branches I pushed from. See #423 for the correct PR

Issue #366, opened 2016-12-19
Parameter color in ColorClip


furas:
Documentation shows example with `color` parameter

    clip = ColorClip(size=(460,380), color=[R,G,B])

http://zulko.github.io/moviepy/getting_started/clips.html

but code expects `col` parameter

    def __init__(self, size, col=(0, 0, 0), ismask=False, duration=None):

https://github.com/Zulko/moviepy/blob/master/moviepy/video/VideoClip.py#L1036

Gloin1313:
Can confirm. Can we change col to color in the code please? Readability counts

kerstin:
@Gloin1313 I think I'd rather `col` were kept and the documentation adapted because people spell the word differently (colour vs. color). Using the shortcut gets rid of the problem.

Gloin1313:
Well, I would naturally spell it colour, but I am well accustomed to spelling it color for anything like this, where I'd expect American spellings to prevail. Still, I don't really mind either way.

kerstin:
I'd also spell it colour but I think being able to switch effortlessly is something that's easier for some than it is for others (could depend on everyday use, or even general level of English). I have definitely got this wrong in CSS despite knowing how I'm supposed to spell it!

I'm also not sure if the spelling is consistent throughout the entire codebase and documentation (I only know I've seen what looked like Frenchisms to me).

earney:
Why not color and colour as parameters?  We would complain if both vars are set and they are not equal.

Gloin1313:
If no-one objects, I'll write this. I'll also include a DeprecationWarning if `col` is used.

Zulko:
Can we go for ``color`` (and color only) ? ``col`` is not "verbose" enough. ``colour`` is less used. Having both color and colour would be confusing.

Gloin1313:
Sure. But I think we need a Deprecation error is `col` is used, because we are changing the API here.

earney:
closing..   PR 424 was merged into repo to fix this issue.

PR #424, opened 2017-02-18
`Colorclip` changed `col`>`color`

Gloin1313:
Changed the `col` paramater of `ColorClip` to `color` and added a DeprecationWarning if the users tries to use `col`.


kerstin:
Does this fix #366? Dropping the no. in here so this PR is linked from the issue.

Gloin1313:
Sorry, yes, it does.

earney:
Gloin1313.  this looks good to me.  Should we go ahead and merge?   Anyone object?

Zulko:
Just two remarks: please enforce PEP8, no line with more than 80 characters, it's really important.
Also -less important- maybe in a first time we could raise a warning instead of an error, i don't really like to have a parameter there with no other purpose than throwing an error.

earney:
@zulko..  is there a way we can add PEP 8 testing of the source code?  I've heard of this, but I haven't tried it myself.  Maybe one else has some experience with PEP 8 testing and github.

Zulko:
[There are](https://pypi.python.org/pypi/pep8) but I've never been really confortable with these, in some cases you have to disregard PEP8 for one line or two, and that makes your tests fail. On moviepy you'll certainly have a lot of red flags. I'd suggest we just keep the basics of PEP8 in mind (line length, naming, etc.) and just eyeball one another's code.

earney:
good points..  We don't want to get too anal about this.. ;)

kerstin:
@earney @Gloin1313 Any decent code editor will allow you to use a word wrap column/ruler, which can be set to 78-80 characters (78 being preferred because of the extra spaces just in case).

It's not just useful if you want to follow PEP8, but also on sites like GitHub, where longer lines have to be scrolled horizontally, which is a pain. For the rest, a read-through of the [PEP8](https://www.python.org/dev/peps/pep-0008/) page as a refresher should suffice (+ maybe checking back with it a few times for those completely new to it).

Gloin1313:
@Zulko I'll add those suggestions. Sorry about the line length thing, I'll make sure it doesn't happen again! When you say "warning", do you mean just `print("Moviepy WARNING: message here)` or do you mean something like [the standard library warnings module](https://docs.python.org/3/library/warnings.html#module-warnings)?
Whichever we do, I'll then add `color = col`, unless both are specified, in which case I'll use `color` and have another warning that `col` is being ignored.

earney:
@Gloin1313 ,  we are using the warnings module in /moviepy/moviepy/video/io/ffmpeg_reader.py.  So that is probably what @Zulko  is referring to.
Can you also create a test or two for this PR and place it in  test/test_PR.py ?  Let me know if you need some help.

Gloin1313:
This is ready to go. Is it OK if I place the test in a separate PR? The `colorclip-update` branch doesn't have the `test_PR.py` file as it was added since I created the branch.

earney:
Sure..  we can merge the two files together later.

Gloin1313:
@earney About the tests: are they just testing for proper usage, or shall I put one in using `col` to check the  warnings work ok as well?

Gloin1313:
Tests are in this one after all. Ready to go

earney:
Nice work..  unless someone objects, I'd merge this into the repo.  Maybe give others a few days to respond (maybe till friday?)

Gloin1313:
There are a lot of commits in the PR. Is the right thing to do "rebase and merge" instead of just "Merge pull request"? Or do we "squash and merge"?

earney:
A squash and merge will do what u want.

PR #428, opened 2017-02-20
fix issue 145.  raise Exception when concatenate method != chain or c…

earney:
fix issue #145     give a clear error message when we pass an incorrect value to concatenate_videoclip, method argument.


Zulko:
Can you remind me what problem this solves ?

earney:
@zulko.     The method argument of concatenate (and concatenate_videoclips) should only be set to chain or compose.   It was requested that if method was set to something else, an error should be raised.

Zulko:
Ok, but the name detection is unnecessary in my opinion, a simple message would do.
Please use PEP8 standards (no line with more than 80 columns) and make imports
at the top of files (not in functions) when possible.

earney:
Thanks, @Zulko ,  I've make those changes.

Zulko:
Thanks !

PR #430, opened 2017-02-20
Test for issue 145

earney:
This branch does contain the knights.mp4 file, which is about 50 mb in size.  We could get having this in the repo by placing a few media files at zulko.github.io, and then use the download_file function to download these if they are missing..   For simpility sake, I just added the media file here.

Issue 145 is a request to make sure that the method argument of concatenation_videoclips gives an error if method != chain or compose.   This seemed like an easy test to get the test suite started.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -9,4 +9,11 @@ from moviepy.editor import *
 
 @pytest.fixture
 def example_video1():
-	
+    pass
+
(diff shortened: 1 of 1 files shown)


Zulko:
lets not place video files in this repo, it will make every cloning much heavier.
The best for examples is to download files directly from youtube, Moviepy has a function for this if i remember well

Issue #285, opened 2016-04-20
concatenate ImageClip 


vparitskiy:
Create several ImageClips:

```
clip_1 = ImageClip(imread('path_to_img.png'), duration=10)
clip_2 = ImageClip(imread('path_to_img.png'), duration=10)
clip_3 = ImageClip(imread('path_to_img.png'), duration=10)

merged_clip = concatenate_videoclips([clip_1, clip_2, clip_3])
```

It will fail with error:

```
Traceback (most recent call last):
  File "/home/vitaliy/projects/quickorama-site/quickorama/../lib/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/vitaliy/projects/quickorama-site/quickorama/../lib/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/vitaliy/projects/quickorama-site/quickorama/media/tasks.py", line 93, in merge_video
    clip = concatenate_videoclips(clips)
  File "/home/vitaliy/projects/quickorama-site/quickorama/../lib/moviepy/video/compositing/concatenate.py", line 95, in concatenate_videoclips
    result.mask = concatenate_videoclips(masks, method="chain", ismask=True)
  File "/home/vitaliy/projects/quickorama-site/quickorama/../lib/moviepy/video/compositing/concatenate.py", line 72, in concatenate_videoclips
    tt = np.cumsum([0] + [c.duration for c in clips])
  File "/home/vitaliy/.virtualenvs/1.9/local/lib/python2.7/site-packages/numpy/core/fromnumeric.py", line 2114, in cumsum
    return _wrapit(a, 'cumsum', axis, dtype, out)
  File "/home/vitaliy/.virtualenvs/1.9/local/lib/python2.7/site-packages/numpy/core/fromnumeric.py", line 48, in _wrapit
    result = getattr(asarray(obj), method)(*args, **kwds)
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

I discovered that while creating ImageClip instance with `transparent=True`, it sets self.mask to another ImageClip, but without duration.

And in `concatenate_videoclips` function it checks if any clip has mask and calculate duration. But duration is None, that's why TypeError.

Maybe, it should set mask.duration to clip.duration?

earney:
[...]

PR #287, opened 2016-04-20
Maybe fixes #285. Set mask duration to clip duration to avoid error.

vparitskiy:

PR #435, opened 2017-02-21
Issue 285

earney:
sets duration of masks to value of clips.
One question, should we fix ImageClip, so that when the duration argument is set, the mask is also set?  I had trouble finding where/how mask was being set, so I went the route of the PR to fix this.  Warning.. this fix could be seen as more of a hack, then a real fix...

### DIFF
--- a/moviepy/video/compositing/concatenate.py
+++ b/moviepy/video/compositing/concatenate.py
@@ -72,7 +72,6 @@ def concatenate_videoclips(clips, method="chain", transition=None,
         clips = reduce(lambda x, y: x + y, l) + [clips[-1]]
         transition = None
 
-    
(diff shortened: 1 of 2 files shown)


earney:
@Zulko , you want to take a look?   I was having trouble figuring out how the mask is being set from the following set of example code...
```
clip_1 = ImageClip('media/python_logo.png', duration=10)
clip_2 = ImageClip('media/python_logo.png', duration=10)
clip_3 = ImageClip('media/python_logo.png', duration=10)

merged_clip = concatenate_videoclips([clip_1, clip_2, clip_3])
```

earney:
PR #287 may also fix issue #285.

PR #441, opened 2017-02-22
Mod versions

earney:
create a simple way to figure out what python modules are installed, what python version is running, and what OS, etc.


earney:
this will be great for those users who are having issues, and they may not know how to figure out what modules they have installed, what python version they are running ,etc.

earney:
we could also add where things like imagemagick is located, and its version, to help with debugging things.  I was thinking about adding those, but decided to start with this simple script to get a discussion going.

PR #443, opened 2017-02-23
Test issue 407 (video has a valid fps after concatenate function)

earney:
check that fps is set to an appropriate value when we concatenate video files.  Some of the clips may not have fps, etc.


earney:
this also just adds a test so merging..

PR #447, opened 2017-02-26
add testing with travis-ci

earney:


earney:
removed youtube-dl because travis-ci doesn't like it.  youtube-dl is a python script and it isn't compatible with all the versions of python we are testing..

PR #451, opened 2017-02-28
add test_compositing.py

earney:
move distribution to Trusty..
get a recent version of ffmpeg since trusty has very old version, which moviepy complains about..


tests/test_compositing.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+from moviepy.editor import *
+
+def test_clips_array():
+    _red = ColorClip((1024,800), col=(255,0,0))
```
mbeacom: Internal vars?

earney: @mbeacom .. sorry, I'm not following you.  What do you mean?

mbeacom: @earney https://shahriar.svbtle.com/underscores-in-python#single-underscore-before-a-name-eg-code-class_1

mbeacom: And referencing PEP-8 http://stackoverflow.com/a/6930223/2560922

earney:
@mbeacom.  I always append my local variables with an underscore.  Is that discouraged?
As the article says, I also do that for variables I consider private, and shouldn't be used externally, but that doesn't apply here.

mbeacom:
I was curious about the reasoning behind the naming convention. All good.

earney:
if we decide on a naming convention, I'm fine with trying to follow it.  It may take some time for me to break old habits, but I can try. :)

Zulko:
I have never seen such a systematic use of underscores before, are there projects that use this ? There are some rare cases where it helps understanding which variables are 'permanent' and which are throw-away, but I'm not sure they help here.

kerstin:
I've never seen underscores used in front of variable names in Python, just as separators between several words making up one name.

PR #446, opened 2017-02-24
fix YouTube embeds in docs

kerstin:
Fixes #304, renders #342 + #343 invalid.

Issue #421, opened 2017-02-17
Add matplotlib example to the user guide?


stefanv:
Thanks for MoviePy!

I was just showing it to a colleague, and realised that the matplotlib example is hidden in a blog post. The common use cases shown here:

http://zulko.github.io/blog/2014/11/29/data-animations-with-python-and-moviepy/

would make for a great addition to the user guide.

kerstin:
Thanks for the suggestion, @stefanv!

@Zulko is pretty busy, so not sure when (if) he'll find time to add his own examples. The project accepts pull requests though, so if you wanted to add examples to the docs, that'd be great!

stefanv:
I'll wait for #368 to be resolved before adding the examples.

flothesof:
Hi @kerstin and @stefanv 

Since I often use moviepy to do matplotlib animations, I'd like to look into adding matplotlib to documentation.
This is where I need some input. The documentation is, so far, organized like this:

- getting started
- gallery 
- example scripts

It turns out that the "gallery" section links to the matplotlib examples from Zulko's blog post. 
So I could expand the section in the gallery part and add matplotlib examples to it. 
However, we could also add a special section since matplotlib has a large user base.
To not disrupt things, I'll go with the first version. Let me know if you think it's better to make a separate page.

On another note, when I open [http://zulko.github.io/moviepy/gallery.html](http://zulko.github.io/moviepy/gallery.html) in my browser, the page downloads lot of small files that look like YouTube links. Do you observe the same thing? Does anyone know how to fix this?

Best regards,
Florian

(pull request to be followed later, hopefully)

kerstin:
@flothesof Just a quick note on your browser prompting you to download stuff: looks like I only fixed video links with #446 but there's still mixed (secure + insecure) content because of images prefixed with `http` instead of `https`. Will fix image links next.

Regarding adding examples: I think I'd also just extend the existing sections for now. If one particular kind becomes so big it could get its own separate section, I guess we could still split it off.

kerstin:
@flothesof The image embeds should be fixed now.

kerstin:
Well, this is weird. It looks like none of my changes are reflected in the docs... hmm.

kerstin:
My bad, I thought GitHub autogenerates documentation like Gitlab does. I'm getting lots of errors and warnings when trying to generate the docs though, and end up with pages missing content or not being there at all.

[...]

PR #462, opened 2017-03-05
Fix mixed content

kerstin:
Fixes remaining insecure content embedded in secure pages (images embedded in MoviePy docs) which gets rejected by browsers. (Continuation of PR #446 prompted by issue described in #421.)

Additionally replaces *all other* insecure links with secure versions for websites that offer `https` (almost all of them).

### DIFF
--- a/docs/FAQ.rst
+++ b/docs/FAQ.rst
@@ -1,30 +1,30 @@
 FAQ and troubleshooting
 =========================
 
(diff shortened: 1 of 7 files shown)


kerstin:
Merging this as it is just a follow-up to a previous PR. All changed URLs were manually tested to make sure they can be opened.

PR #463, opened 2017-03-05
Small improvements to docs pages, docs usage

kerstin:
`makehtml.sh` can now also be used on Mac (automatically opens the generated HTML in Firefox).


kerstin:
I think it should be fine to merge this...

Issue #482, opened 2017-03-11
Fix documentation


keikoro:
The documentation needs to be fixed.

Currently, Sphinx throws lots of errors and warnings and the generated HTML is missing parts and even whole pages. It also looks like files are getting duplicated now (maybe due to changes in how Sphinx does things?): newly generated sources have `.rst.txt` as file extension, but files with the same contents already exist with a `.txt` extension. (These files are saved in the `_sources` dir of the generated docs.)

As far as I can tell, Sphinx is supposed to incorporate the docstrings of certain classes/methods on some pages, and that is currently failing, so I'm guessing these "includes" aren't referenced correctly (anymore).

kerstin:
### Short docs building how-to
To build the docs, a few additional packages (like Sphinx) have to be `pip` installed.

Then the Sphinx `$ make html` command has to be run in the `docs` directory on `master`, which contains the source files of the documentation.

The resulting HTML (etc.) files are saved *outside* of the directory that MoviePy sits in, to a directory called `docs` *on the same level* as one's local copy of the MoviePy repo. Within this `docs` dir, there's an `html` dir and the `index.html` file contained in there is the "start page" of the documentation (same as what's seen on [zulko.github.io/moviepy](https://zulko.github.io/moviepy/)).

To update the online documenation, one has to switch to the `gh-pages` branch of one's local copy of the repo and then copy over the contents of the `html` directory containing the newly generated files to the root of `gh-pages`.

kerstin:
The linked PR #483 contains a file with the list of packages needed for generating the HTML files that make up the documentation.

To install them, run
```$ pip install -r requirements_docs.txt```

flothesof:
Cool, looks like you managed to find out how it works. The how-to-build-the-docs would definitely be worth archiving (but where?...). 
Best,
Florian

kerstin:
@flothesof Oh, I knew how to do it alright. But the docs are broken.

And yeah, the idea of adding a `CONTRIBUTING.md` file to the repo in the future is floating around, and there's also a Wiki (or the beginnings of a Wiki). Though I think we should fix the docs first before making people aware of how they could *theoretically* contribute.

tburrows13:
The instructions at https://github.com/Zulko/moviepy#documentation are:

> Running build_docs has additional dependencies that require installation.
> 
>     $ (sudo) pip install moviepy[docs]
> The documentation can be generated and viewed via:
> 
>     $ python setup.py build_docs

[...]

PR #483, opened 2017-03-11
include requirements file for docs

kerstin:
As mentioned in #482, building the documentation requires some additional packages to be installed, which are excluded when using `setup.py`.

This requirements file contains the packages needed for doc building.
They can be installed by running `$ pip install -r requirements_docs.txt`


earney:
this is awesome..  Does anyone see a reason we cannot merge this (maybe an additional needed dependency, etc)?

PR #485, opened 2017-03-12
Change `verbose_print`

Gloin1313:
I don't see why we need the fancy `sys.stout` when `print` prints to `stout` by default, and flushes automatically. This also means we can use `verbose_print` just like `print` except with the `verbose` parameter before anything to be printed. e.g. `verbose_print(True, var1, var2) which would have failed with the old `verbose_print`.

I tried to use `verbose_print` in my application, but acted very weirdly, hence the change.

### DIFF
--- a/moviepy/tools.py
+++ b/moviepy/tools.py
@@ -11,16 +11,10 @@ import os
 from .compat import DEVNULL
 
 
-def sys_write_flush(s):
-    """ Writes and flushes without delay a text in the console """
-    sys.stdout.write(s)
-    sys.stdout.flush()
-
-
-def verbose_print(verbose, s):
-    """ Only prints s (with sys_write_flush) if verbose is True."""
+def verbose_print(verbose, *args):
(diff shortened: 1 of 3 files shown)


Zulko:
I should have documented this one, but basically the issue was that in some consoles (and in the IPython notebook, at least at the time), "print" commands get delayed, while stdout.flush are instantaneous, so my method was better at providing feedback.

I would be against changing it to print unless it solves some known issues.

Gloin1313:
I posted this as a question on Stack Overflow here: http://stackoverflow.com/questions/42752218/does-print-delay-in-some-consoles-where-stdout-flush-doesnt

earney:
@Gloin1313   do you think you can add @Zulko 's  comment in the source code so that if someone tries to change this (like you did), we have an explanation of why it needs to stay as is?  Thanks!

Gloin1313:
Yep, will do!

Gloin1313:
Done in #514

PR #489, opened 2017-03-13
Allow resizing frames in ffmpeg when reading

gyglim:
This PR is adds the possibility to read videos frames that have been resized in ffmpeg.
This is much faster than streaming in high-res and then resizing in python and important for applications where speed is crucial

### DIFF
--- a/moviepy/video/io/VideoFileClip.py
+++ b/moviepy/video/io/VideoFileClip.py
@@ -32,6 +32,17 @@ class VideoFileClip(VideoClip):
     audio:
       Set to `False` if the clip doesn't have any audio or if you do not
       wish to read the audio.
+
+    target_resolution:
+      Set to (desired_height, desired_width) to have ffmpeg resize the frames before return them.
+      This is much faster than streaming in high-res and then resizing.
+      if either size is None, the frames are resized by keeping the existing aspect ratio.
+
+    resize_algorithm:
+      The algorithm used for resizing. Default: "bicubic", other popular options include "bilinear"
+      and "fast_bilinear".
+      For more information, see https://ffmpeg.org/ffmpeg-scaler.html
+
       
     Attributes
     -----------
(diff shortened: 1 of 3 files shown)


Zulko:
Interesting. Ideally, I would be for using ffmpeg just for decoding the video, and then do all the logics inside moviepy. Can you give numbers on how much you time your PR saves ? How do you explain the speed difference between resizing on ffmpeg side versus on moviepy's side ? It's C routines on both sides.

gyglim:
quite substantial, but it was part of a pipeline, so I don't have exact numbers
I will update this comment with a comparison in a few minutes

earney:
@zulko  Since you have contributed to this conversation, I want to give you a chance to approve/disapprove this PR.

Zulko:
I'd say approve. I am a bit against creating another way of resizing and against using ffmpeg for more than read/write, but in this case it seems that there is much to gain (and I cant think of a simple way to get the same performance).

gyglim:
Awesome. Another advantage I see is that it is the same for everyone. The resizing in moviepy tries OpenCV, the PIL, then scipy:
https://github.com/Zulko/moviepy/blob/32b700726508bf26d1d6d8f81f7820df958ce1dc/moviepy/video/fx/resize.py#L4

But everybody needs to have ffmpeg if they are using moviepy. Not necessarily OpenCV. Resizing with PIL is much slower

bilinear:
`1 loop, best of 3: 12.1 s per loop`
ANTIALIAS (as used in moviepy):
`1 loop, best of 3: 21.7 s per loop`
so double the time compared to OpenCV. I know scipy is also slower from earlier tests.

gyglim:
give me a minute before merging.

Zulko:
@gyglim I know, I use Pillow by default in moviepy because it's lightweight and easy to install everywhere. It's not difficult to create a custom filter using opencv's resize.

Thanks for the PR, it's a solid one. For future contributions, please make sure to use PEP8 (all lines <80 chars) and consistent names (resize_algo vs resize algorithms).

gyglim:
@Zulko Thank you.
re PEP8 and variable names: Will do.

fixed that you can do also up-scaling incl. test (was an artifact from our own version).
From my side it is all good now

Issue #464, opened 2017-03-05
ImageSequenceClip repeats frames depending on fps


PhillipGH:
I was putting together a sequence of frames with ImageSequenceClip and saw that my animation seemed to hang at a certain point as if there were a repeated frame. However, I noticed the problem went away when I switched from 24 to 30 fps.

I wrote a program to reproduce the issue that makes ImageSequenceClips out of the same set of frames at different fps's and looks for repeated frames.

```
from moviepy.editor import *

frames_filenames = []
# set of frames that are all different
for i in range(20):
	frames_filenames.append('frames2/processing/' + str(i) + 'frame.png')

for frames_per_second in range(1,31):
	print("\nfps: " + str(frames_per_second))
	movie = ImageSequenceClip(frames_filenames, fps=frames_per_second)
	lastPixels = []
	count = 0
	for frame in movie.iter_frames():
		# band of pixels that is different in every frame
		pixels = frame[306,560:600,0]
		if len(pixels) == len(lastPixels):
			tuple = zip(pixels, lastPixels)
			difference = [x for x in tuple if x[0] == x[1]]
			if len(difference) == 600 - 560:
				print("frame " + str(count))
				print(difference)
				print("\n")
		lastPixels = pixels
		count += 1
```

Here is the output of this program:

```
fps: 1

fps: 2

fps: 3
frame 10
[(184, 184), (184, 184), (182, 182), (182, 182), (180, 180), (180, 180), (178, 178), (178, 178), (177, 177), (177, 177), (174, 174), (171, 171), (169, 169), (166, 166), (161, 161), (158, 158), (157, 157), (146, 146), (136, 136), (131, 131), (128, 128), (121, 121), (117, 117), (115, 115), (114, 114), (117, 117), (127, 127), (138, 138), (147, 147), (153, 153), (159, 159), (162, 162), (166, 166), (166, 166), (170, 170), (170, 170), (171, 171), (175, 175), (177, 177), (181, 181)]



fps: 4

fps: 5
frame 15
[...]

PR #494, opened 2017-03-14
Fix issue #464, repeated/skipped frames in ImageSequenceClip

neitzal:
This PR fixes the occasional repeating and skipping of frames mentioned in issue #464 due to floating point inaccuracies in images_starts.

### DIFF
--- a/moviepy/video/io/ImageSequenceClip.py
+++ b/moviepy/video/io/ImageSequenceClip.py
@@ -78,8 +78,10 @@ class ImageSequenceClip(VideoClip):
         self.fps = fps
         if fps is not None:
             durations = [1.0/fps for image in sequence]
+            self.images_starts = [1.0*i/fps-np.finfo(np.float32).eps for i in range(len(sequence))]
+        else:
+            self.images_starts = [0]+list(np.cumsum(durations))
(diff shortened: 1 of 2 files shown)


earney:
@neitzal, would it be possible for you to add a test for this that would fail before your fix, but succeeds after your fix?   You can post it here or you can add it to tests/test_issues if you want.  Is that okay?

neitzal:
@earney Sure, I added a test for one failure case that now succeeds. The test is not exhaustive though, in the sense that it does not verify that no fps-number (in a reasonable range) ever produces a frame mismatch. I guess that would be a little too expensive to run every time.

earney:
Awesome..  Thanks!

earney:
@Zulko   do you think this would be okay to merge into the repo?  I'm not an expert on ImageSequenceClip so I'm a bit hesitant.

earney:
scipy does not support python 3.3 anymore (but supports 2.7, 3.4, 3.5 and 3.6).  if we accept this PR we will lose python 3.3.  is that okay with everyone?

earney:
any objections to this PR?

Zulko:
LGTM

PR #503, opened 2017-03-18
Add aspect_ratio @property to VideoClip

scherroman:


Zulko:
That's a useful one, thanks.

PR #509, opened 2017-03-25
Added another project using MoviePy

justswim:
Hi! 

I hacked together an online video meme maker called Kapwing using MoviePy and I was hoping you could feature it in the documentation? It's basically a web UI to create video memes for Facebook. 

MoviePy is awesome! Thanks for your hard work on it. 

-Eric

PR #510, opened 2017-03-25
Correcting text

gpantelis:
Change the 'au' to 'and'

### DIFF
--- a/docs/getting_started/videoclips.rst
+++ b/docs/getting_started/videoclips.rst
@@ -3,7 +3,7 @@
 Creating and exporting video clips
 ===================================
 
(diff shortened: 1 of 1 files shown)


Zulko:
thanks !

gpantelis:
What coverage represents? sorry for this question.

kerstin:
@gpantelis I asked basically the same question in #509, so will redirect you to there.

I *think* it tests for actual testing, so new contents/lines that are not covered by tests (which documentation and doc fixes etc. naturally aren't) will always decrease the existing "coverage" ("are these lines of code/stuff covered by tests?").

Zulko:
[The wikipedia article](https://en.wikipedia.org/wiki/Code_coverage). Now I don't know exactly how coveralls.io decides what is code and what is not, so I woulnd't attach too much importance to the small changes it announces (+-1%) and use it more as a general indicator.

I can see two main uses: (1) if the test suite gets refactored, ensure that it still covers as much features as before. (2) if a contributor adds code introducing many new features, ensure that the pull request is contains tests that properly cover the new code.

PR #528, opened 2017-04-05
Add int() wrapper to scroll to prevent floats

### LINKED ISSUE #527: vfx.scroll giving TypeError: slice indices must be integers or None or have an __index__ method
I run:
```clip = ImageClip("panorama.jpg")
new_clip = vfx.scroll(clip, w=1000, x_speed=5)
new_clip = new_clip.set_duration(float(clip.w - 1000) / 5)
new_clip.write_videofile("pano.mp4", fps=24)
```
as recommended by Zulko in #240, but it gives the following error:

```Traceback (most recent call last):
  File "/Users/tomburrows/Dropbox/Python Programming/QF_Editor/panorama_scroll.py", line 8, in <module>
    new_clip.write_videofile("pano.mp4")
  File "<decorator-gen-51>", line 2, in write_videofile
  File "/anaconda/lib/python3.5/site-packages/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, *a, **k)
  File "<decorator-gen-50>", line 2, in write_videofile
  File "/anaconda/lib/python3.5/site-packages/moviepy/decorators.py", line 137, in use_clip_fps_by_default
    return f(clip, *new_a, **new_kw)
  File "<decorator-gen-49>", line 2, in write_videofile
  File "/anaconda/lib/python3.5/site-packages/moviepy/decorators.py", line 22, in convert_masks_to_RGB
    return f(clip, *a, **k)
  File "/anaconda/lib/python3.5/site-packages/moviepy/video/VideoClip.py", line 345, in write_videofile
    progress_bar=progress_bar)
  File "/anaconda/lib/python3.5/site-packages/moviepy/video/io/ffmpeg_writer.py", line 209, in ffmpeg_write_video
    fps=fps, dtype="uint8"):
  File "/anaconda/lib/python3.5/site-packages/tqdm/_tqdm.py", line 833, in __iter__
    for obj in iterable:
  File "/anaconda/lib/python3.5/site-packages/moviepy/Clip.py", line 475, in generator
    frame = self.get_frame(t)
  File "<decorator-gen-14>", line 2, in get_frame
  File "/anaconda/lib/python3.5/site-packages/moviepy/decorators.py", line 89, in wrapper
    return f(*new_a, **new_kw)
  File "/anaconda/lib/python3.5/site-packages/moviepy/Clip.py", line 95, in get_frame
    return self.make_frame(t)
  File "/anaconda/lib/python3.5/site-packages/moviepy/Clip.py", line 136, in <lambda>
    newclip = self.set_make_frame(lambda t: fun(self.get_frame, t))
  File "/anaconda/lib/python3.5/site-packages/moviepy/video/fx/scroll.py", line 17, in f
    return gf(t)[y:y+h, x:x+w]
TypeError: slice indices must be integers or None or have an __index__ method
```
Doing some research, this is because `x` is  0 when you first call it, but then when it writes the video to a file, it iterates for each frame, and the second time, x is 1.0. This is a float, not an int so it fails.
A solution that worked for me is to change lines 14 and 15 of `scroll.py` to change the float into a int.
Can anyone repro this, and shall we make the suggested changes?

Gloin1313:
Fixes #527. I'll add a test for this soon. EDIT Ready to go


earney:
the current code base (without this PR), produces the correct output for me.  Below is my code, does it match yours?   In the example below, I modified the speed and the duration, but I used your original values and they worked too.

```
clip = ImageClip("media/panorama.jpg")
    new_clip = vfx.scroll(clip, w=1000, x_speed=50)
    new_clip = new_clip.set_duration(20)
    new_clip.fps = 24
    new_clip.write_videofile("pano.mp4")
```

earney:
By the way, I'm using python versions 2.7.13 and 3.5.3.  How about you?

earney:
the code does work on my machine, but I cannot figure out why.. (it should not work).   x is a numpy.float64, which does NOT have a __index__ function, but the array references still work.     I say we merge this PR.  converting to an int is probably the best way to guarantee this works.

earney:
another suggestion.   since scroll only uses numpy for the round function, we could probably remove the numpy dependency.  How does that sound?

Gloin1313:
Summary of discussion:
Scroll works with numpy 1.11, but not 1.12. No-one knows why, as `x` is still a float, that doesn't not appear to have a `__index__()` method.
`x` is always a float of the form `z.0`, so converting it to `int()` does no harm. I'm happy to proceed.

@earney are you saying we should change `np.round` to `round`, then we can remove the `import numpy` at the top?

earney:
Yes, lets remove the numpy import

Gloin1313:
Will `np.round` do the same as `round`? Because there is a slight difference.

earney:
That is a good point.  I believe in this case the difference does not matter.

Gloin1313:
@Zulko? Do you agree?
If so, I'll edit and merge.

mbeacom:
@Gloin1313 Re-assigning this to the next release milestone since this branch is currently conflicting with `master` and is awaiting revisions.

earney:
conflict fixed.. :)

earney:
@Gloin1313   lets merge this in if you want.

PR #537, opened 2017-04-11
Fix youtube video rendering in documentation

mbeacom:
This PR resolves youtube video rendering in the example documentation.


Gloin1313:
Thanks for doing this; I'm afraid I have no idea how all this works, though!

earney:
I'm glad to see the docs fixed.   We also need to verify that the examples still work with this version of moviepy.  Certain examples such as headblur do not work.  We really need to fix these in another PR.

PR #540, opened 2017-04-12
add opencv dependency since headblur effect depends on it.

earney:


Zulko:
won't that multiply the setup time of everything ? I mean, opencv is a heavy dependency.

earney:
it seems that all the dependencies for opencv are already installed (numpy, etc)

PR #541, opened 2017-04-12
add freeze, freeze_region, and headblur tests

earney:


earney:
I'll delay adding the headblur test, since it looks like some might think the opencv dependencies maybe on the heavy side.  I don't think so, but I'll delay a bit on this PR.

earney:
The time to run the tests have increased a couple of minutes, but I have added 3 or 4 additional tests.  I'm not sure how much of that time is because of the headblur test.  Something I'll need to investigate if people think that 9 minutes is too long for these tests.

tburrows13:
Closed as now redundant due to #1124.

PR #552, opened 2017-04-20
attempt to fix Issue 190 (use b/w images with ImageSequenceClip

earney:


earney:
guys.. this produces no errors, but the output is not correct.  can someone take a look and see what I'm missing with this fix?

PR #558, opened 2017-05-04
FX: Allows modify the volume by arrays

agalera:
Probably the name "levels" is not the most correct

```python
from moviepy.audio.fx.audio_levels import audio_levels


clip = VideoFileClip('example.mp4').subclip(0, 10)
clip = clip.fx(audio_levels, [(0, 1), (5, 0), (10, 1)], fast=True)
clip.write_videofile("result.mp4")
```

Example videos:
Normal:
https://youtu.be/tRsvKTv-v9Q

Fast:
https://youtu.be/1kfyBJ58sN0

Explain differences (blue is normal, green is fast):

http://imgur.com/04L4ohP

This happens because in the fast version, multiply the array of 2000 values and it can happen.

The normal version uses the "fast" part only when it detects that it does not have any point that affects.

In the last commit, the speed difference is very small, I think you could eliminate the fast option

### DIFF
--- a/moviepy/audio/fx/audio_levels.py
+++ b/moviepy/audio/fx/audio_levels.py
@@ -56,12 +56,12 @@ def audio_levels(clip, levels, fast=False):
             return np.vstack([factor, factor]).T * gft
 
         def not_scalar(t):
-            gft = gf(t)
             prv, nxt, final = get_range_levels(t[0])
 
             if t[-1] < nxt[0] or final:
                 return not_scalar_fast(t, prv, nxt)
 
+            gft = gf(t)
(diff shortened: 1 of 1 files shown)


Gloin1313:
Hi!
As I said in your other PR, would you mind creating a test in `tests/test_PR` to demonstrate the usage of this feature?

mbeacom:
@kianxineki Can you rebase this with `master` as well?

Thank you!

mondeja:
This should be implemented easily passing a function as `factor` argument in `multiply_volume` and `multiply_stereo_volume`.

PR #561, opened 2017-05-05
Fix travis build and enable pip caching

mbeacom:
The goal of this PR is to:

- Enable `pip` caching on Travis builds.
- Silence the `ffmpeg` install via `-qq`.
- Lowercase `dist` name to conform to travis docs.
- Add missing `sudo: required` entry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 dist: trusty
 sudo: required
 language: python
+cache: pip
 python:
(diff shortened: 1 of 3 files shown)


Gloin1313:
I'm afraid I don't know much about this stuff, but it looks good. When we've sorted out the problem with Travis, I think we can merge.

Issue #281, opened 2016-04-03
write_audiofile raises IndexError/IOError


xjcl:
This happened while writing a large (159MB, almost 3 hours) audio file.

**Related issues**

See #246, #262 (?), #265 

**Temporary Fix**

write_audiofile completed successfully after I replaced each call of the form `AudioFileClip(s)` in my script by `AudioFileClip(s, buffersize=20000)` (i.e. setting buffersize to 20k). At that time I mistakenly believed the default buffersize for AudioFileClips to be 10k, but it is in fact 200k, so lowering(!) buffersize is what made it work properly. The call to write_audiofile remained unmodified (mentioned because this function also takes a buffersize as argument). 

I have no idea whether this fix works in general or just for my specific file.

**Traceback**

(Note that IOError is aliased to OSError in Python3)

```
[MoviePy] Writing audio in ops01to40.mp3
 96%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▋           | 219372/228977 [03:17<00:07, 1323.03it/s]Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/moviepy/audio/io/readers.py", line 187, in get_frame
    result[in_time] = self.buffer[indices]
IndexError: index 100001 is out of bounds for axis 0 with size 100001

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "conan2.py", line 54, in <module>
    res.write_audiofile('ops01to40.mp3')
  File "<decorator-gen-73>", line 2, in write_audiofile
  File "/usr/lib/python3.5/site-packages/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, *a, **k)
  File "/usr/lib/python3.5/site-packages/moviepy/audio/AudioClip.py", line 204, in write_audiofile
    verbose=verbose, ffmpeg_params=ffmpeg_params)
  File "<decorator-gen-70>", line 2, in ffmpeg_audiowrite
  File "/usr/lib/python3.5/site-packages/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, *a, **k)
  File "/usr/lib/python3.5/site-packages/moviepy/audio/io/ffmpeg_audiowriter.py", line 161, in ffmpeg_audiowrite
    nbytes= nbytes, fps=fps):
[...]

Issue #269, opened 2016-02-27
zero-size array to reduction operation minimum which has no identity


fortunto2:
hi, a common mistake in the files when creating audio. 
I add to the video, a music mp3 track.
What could be the problem?

`[MoviePy] Writing audio in baaadmot_15-19-1-1TEMP_MPY_wvf_snd.mp3
 38%|███▊      | 2045/5356 [00:03<00:05, 646.43it/s]----------
zero-size array to reduction operation minimum which has no identity
----------
Traceback (most recent call last):
  File "./render_last.py", line 318, in renderJob
    duration = template.fast2.render(allParam, data, concatPath)
  File "/ubuntu/newconv/template/fast2.py", line 301, in render
    ffmpeg_params=ffmpeg_params, threads=8, write_logfile=True)
  File "<string>", line 2, in write_videofile
  File "/home/ubuntu/anaconda/lib/python2.7/site-packages/moviepy-0.2.2.11-py2.7.egg/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, *a, **k)
  File "<string>", line 2, in write_videofile
  File "/home/ubuntu/anaconda/lib/python2.7/site-packages/moviepy-0.2.2.11-py2.7.egg/moviepy/decorators.py", line 137, in use_clip_fps_by_default
    return f(clip, *new_a, **new_kw)
  File "<string>", line 2, in write_videofile
  File "/home/ubuntu/anaconda/lib/python2.7/site-packages/moviepy-0.2.2.11-py2.7.egg/moviepy/decorators.py", line 22, in convert_masks_to_RGB
    return f(clip, *a, **k)
  File "/home/ubuntu/anaconda/lib/python2.7/site-packages/moviepy-0.2.2.11-py2.7.egg/moviepy/video/VideoClip.py", line 330, in write_videofile
    verbose=verbose)
  File "<string>", line 2, in write_audiofile
  File "/home/ubuntu/anaconda/lib/python2.7/site-packages/moviepy-0.2.2.11-py2.7.egg/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, *a, **k)
  File "/home/ubuntu/anaconda/lib/python2.7/site-packages/moviepy-0.2.2.11-py2.7.egg/moviepy/audio/AudioClip.py", line 204, in write_audiofile
    verbose=verbose, ffmpeg_params=ffmpeg_params)
  File "<string>", line 2, in ffmpeg_audiowrite
[...]

PR #570, opened 2017-05-25
Fix out of bounds error

### LINKED ISSUE #19: Unclosed processes.
When I'm trying to get video duration

video_clip = VideoFileClip(video_file_path)
print video_clip.duration

I see 2 processes, that are never closed
_ ffmpeg -i /tmp/9ea28576d6154d568b755f4cd9718c77.mp4 -f image2pipe -pix_fmt rgb24 -vcodec rawvideo -
_ ffmpeg -i /tmp/9ea28576d6154d568b755f4cd9718c77.mp4 -vn -f s16le -acodec pcm_s16le -ar 44100 -ac 2 -

Can you advise?

shawwn:
Closes #19, #246, #269, #281.

`iter_chunks` now uses `np.linspace` to generate indices. In addition
to fixing the out of bounds error, this also ensures each chunk
generates a uniform number of indices.

### DIFF
--- a/moviepy/audio/AudioClip.py
+++ b/moviepy/audio/AudioClip.py
@@ -67,15 +67,14 @@ class AudioClip(Clip):
         
         totalsize = int(fps*self.duration)
 
-        if (totalsize % chunksize == 0):
-            nchunks = totalsize // chunksize
-        else:
-            nchunks = totalsize // chunksize + 1
+        nchunks = totalsize // chunksize + 1
 
-        pospos = list(range(0, totalsize,  chunksize))+[totalsize]
+        pospos = np.linspace(0, totalsize, nchunks + 1, endpoint=True, dtype=int)
 
         def generator():
             for i in range(nchunks):
+                size = pospos[i+1] - pospos[i]
+                assert(size <= chunksize)
                 tt = (1.0/fps)*np.arange(pospos[i],pospos[i+1])
                 yield self.to_soundarray(tt, nbytes= nbytes, quantize=quantize, fps=fps,
                                          buffersize=chunksize)
--- a/moviepy/audio/io/readers.py
+++ b/moviepy/audio/io/readers.py
@@ -185,6 +185,8 @@ class FFMPEG_AudioReader:
             try:
                 result = np.zeros((len(tt),self.nchannels))
                 indices = frames - self.buffer_startframe
+                if len(self.buffer) < self.buffersize // 2:
+                    indices = indices - (self.buffersize // 2 - len(self.buffer) + 1)
                 result[in_time] = self.buffer[indices]
                 return result
 


shawwn:
Hmm, it looks like this PR doesn't solve all the cases.  I'll keep investigating.

moviepy/audio/AudioClip.py
```diff
@@ -67,15 +67,14 @@ def iter_chunks(self, chunksize=None, chunk_duration=None, fps=None,
...
         totalsize = int(fps*self.duration)
 
-        if (totalsize % chunksize == 0):
-            nchunks = totalsize // chunksize
-        else:
-            nchunks = totalsize // chunksize + 1
+        nchunks = totalsize // chunksize + 1
 
-        pospos = list(range(0, totalsize,  chunksize))+[totalsize]
+        pospos = np.linspace(0, totalsize, nchunks + 1, endpoint=True, dtype=int)
```
mbeacom: Is it intended to add `+ 1` to nchunks on L70 and add `+ 1` again in `pospos` ?

shawwn: Yes. Since we're generating `endpoint=True`, we need to pass in one more than the number of chunks. I.e. if `nchunks` is 1,  we tell `linespace` to return two points (a start and an end).

shawwn:
As far as I can tell, that last commit fixes all remaining errors on my machine. I've been making heavy use of `moviepy` to generate spectrograms of a large number of videos, and I kept running into the `IndexError` problem. After these commits, I haven't seen any errors after processing ~30 videos.

shawwn:
Ok, I've successfully stress-tested this PR by processing several hour-long videos without encountering any errors at all. I believe this PR is ready to merge.

The test suite seems to pass, but on my machine `python3 setup.py test` results in six errors like:

```
E           OSError: [Errno 32] Broken pipe
E           
E           MoviePy error: FFMPEG encountered the following error while writing file /tmp/crop1.webm:
E           
E            crop1TEMP_MPY_wvf_snd.ogg: No such file or directory

moviepy/video/io/ffmpeg_writer.py:178: OSError
```

I'm not quite sure why that's happening, but I don't think it's related to this PR.

If you have any questions or concerns, let me know.

Gloin1313:
@shawwn I don't think those errors are relevant. As long as it passes on Travis, I think it should be ok.

shawwn:
Ok, cool. In that case, this PR is ready to merge.

shawwn:
After heavy usage, I still haven't run into any out of bounds errors (or any other problems).

stonebig:
merging soon ?

kevinmcalear:
Any idea when this can be merged? This would fix an issue that I'm currently dealing with.

shawwn:
@mbeacom I still haven't run into any out of bounds errors, and @stonebig @kevinmcalear have requested this be merged. Since this PR is now four months old, can it be accepted?

shawwn:
(Or @Zulko.)

shawwn:
@Zulko @mbeacom, could this PR be merged? Several users have asked for this, and after heavy usage I still haven't run into any out of bounds errors or any other problems.

I suspect this PR has languished because it changes the bounds checking. In general, it's good to be suspicious of doing this carelessly. But the math is correct now; the reason users have been running into errors is that the math was off in certain corner cases.

Another reason is probably that this seems like a magical, unexplained change in `moviepy/audio/io/readers.py`:

```py
             try:
                  result = np.zeros((len(tt),self.nchannels))
                  indices = frames - self.buffer_startframe
 +                if len(self.buffer) < self.buffersize // 2:
 +                    indices = indices - (self.buffersize // 2 - len(self.buffer) + 1)
                  result[in_time] = self.buffer[indices]
                  return result
```

Again, it's good to be hesitant of changes like that, but FWIW I arrived at this solution by running moviepy on a video that breaks, then sat in `pdb.trace()` and looked at the actual data. The above snippet fixes the out of bounds errors.

If other users are running into out of bounds errors with moviepy and just need a fix, I suppose you can apply the above changes manually. But it'd be nice to get this merged.

Zulko:
Sorry for the delay. I believe most maintainers are too busy right now to give this much thought so we'll have to trust you on this (I hope that was the right decision :) ).

PR #578, opened 2017-06-03
Correctiong ffmpeg_resize parameters

gpantelis:
The previous res[0] and res[1] they are not defined anywhere , and in this command in ffmpeg needs  to define the new width and the  new height. Check the url above: https://trac.ffmpeg.org/wiki/Scaling%20(resizing)%20with%20ffmpeg


tburrows13:
This is fixed in master.

Issue #574, opened 2017-05-30
wrong video duration value 


earney:
It seems that subclip is producing the wrong duration at times.   The duration is 1100.17, but the 4th line in the output shows 497.16 instead of 1100.17.   I wonder if this is the mask, or something about one of the decorators?  This causes issues later when you try to do a video.subclip("0:10:00"), which the t_start > duration.   I will assign myself to this issue, but if someone knows the answer, let me know.. Thanks!
Billy

```
video=concatenate([title, info_arrives1, part2, part3, part4], method="compose")

print(video.duration)
video1=video.subclip(0,17)
print("video2")
video2=video.subclip(23,"0:01:23")
print("video3")
video3=video.subclip("0:03:36", "0:05:22")
```

```
1100.17
<moviepy.video.compositing.CompositeVideoClip.CompositeVideoClip object at 0x7f43b183ceb8> 1100.17 0
<moviepy.audio.AudioClip.CompositeAudioClip object at 0x7f43b18518d0> 1100.17 0
<moviepy.video.compositing.CompositeVideoClip.CompositeVideoClip object at 0x7f43b1851908> 497.16 0
video2
<moviepy.video.compositing.CompositeVideoClip.CompositeVideoClip object at 0x7f43b183ceb8> 1100.17 23
<moviepy.audio.AudioClip.CompositeAudioClip object at 0x7f43b1851b00> 1100.17 23
<moviepy.video.compositing.CompositeVideoClip.CompositeVideoClip object at 0x7f43b1851b70> 497.16 23
video3
```

earney:
FYI..  this bug is present when concatenates method argument is set to "compose").

earney:
A question for all repo maintainers..  What is the difference between the two statements below?
@mbeacom 
@Zulko 
@Gloin1313 

```
concatenate_video([video1, video2, video3], method="compose")
```
and
```
CompositeVideoClip([video1, video2, video3])
```

Zulko:
The first statement will set the clip starts so they play one after another, the second one will start all 3 clips at the same time (if they are the same size, only video3 will be displayed since it is on the top).

earney:
@zulko  thanks..   That is what I thought..  So the duration of the resulting video from the first statement should be the sum of the videos that make up the first argument, correct? 
 If so, this issue is valid, and I'll need to fix this issue.

earney:
PR 585 should fix this issue.

PR #585, opened 2017-06-14
Issue #574, fix duration of masks when using concatenate(.., method="compose")

earney:


Gloin1313:
Sorry, I thought you had the wrong issue number, so I changed the title. I was wrong, #574 is correct!

kencochrane:
What is the status of this one, more work needed, or can we merge?

PR #603, opened 2017-07-02
Exceptions do not have a .message attribute.

Julian-O:
These exception handles interrogate the exception for the ``message`` attribute. These exceptions don't have that attribute.

### DIFF
--- a/moviepy/config.py
+++ b/moviepy/config.py
@@ -45,10 +45,9 @@ elif FFMPEG_BINARY=='auto-detect':
 else:
     success, err = try_cmd([FFMPEG_BINARY])
     if not success:
-        raise IOError(err.message +
-                 "The path specified for the ffmpeg binary might be wrong")
-
-
+        raise IOError(
+            str(err) +
+            " - The path specified for the ffmpeg binary might be wrong")
 
 if IMAGEMAGICK_BINARY=='auto-detect':
     if os.name == 'nt':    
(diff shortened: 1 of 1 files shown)


moviepy/config.py
```diff
@@ -45,10 +45,9 @@ def try_cmd(cmd):
 else:
     success, err = try_cmd([FFMPEG_BINARY])
     if not success:
-        raise IOError(err.message +
-                 "The path specified for the ffmpeg binary might be wrong")
-
-
+        raise IOError(
+            str(err) +
+            " - The path specified for the ffmpeg binary might be wrong")
```
kencochrane: Could this be because of the version of Python used? Maybe one version has it, and another doesn't? Or is it just a bug?

Julian-O: According to [this](https://stackoverflow.com/questions/33239308/how-to-get-exception-message-in-python-properly), some built-in exceptions have a msg attribute, some don't.

IOError's definition is platform dependent. When I look at builtins.py on Windows, Python 3.6, it isn't present. Maybe it is on Linux?

Julian-O: Wait, some have the `msg` attribute. I am not seeing any (on Windows, Python 3.6) with a `message` attribute. In any case, the `message` attribute isn't standard, and this change is still recommended.

kencochrane: Thanks for looking that up. I agree, best to make the change.

PR #597, opened 2017-06-30
Support for running tests on Windows

Julian-O:
I have been trying to run the unit-tests on a Windows 10 machine.

Here are some minor fixes to help get past some stumbling blocks.

(Currently: 44 of 53 tests passing.)

PR #604, opened 2017-07-02
Help tests run on Windows - don't assume temp dir or fonts.

Julian-O:
This is a re-attempt at PR #597.

* Python already has a feature for finding the temp dir. Changed
test helper to take advantage of it.

* Still outstanding: Several hard-coded references to /tmp appear in
the tests.

* Liberation-Mono is not commonly installed on Windows, and even when
it is, the font has a different name. Provide a fall-back for Windows
fonts. (Considered the use of a 3rd party tool to help select, but
seemed overkill.)


tburrows13:
This doesn't appear to be need anymore... one change was adding a `close()` call, which I don't think is implemented yet. The other appears to have been fixed elsewhere

Issue #598, opened 2017-06-30
test_duration() fails in test_TextClip()


Julian-O:
I am trying to get the tests to run on my Windows machine. One of the tests that fails is `test_TextClip::test_duration()`.

However, I don't think it is a problem with my set-up. The test *should* fail. 

At first glance, the test seems reasonable: 
```
    clip = TextClip('hello world', size=(1280,720), color='white')
    clip.set_duration(5)
    assert clip.duration == 5
```

However, `set_duration()` has an `@outplace` decorator, which means it doesn't act on the clip you pass in, but instead makes a copy and acts on that. So, the original clip's duration is not affected, and the test fails.

Either it should not have the `@outplace` decorator, or the second line should read:

```
    clip = clip.set_duration()
```

(I am afraid I can't tell which is more appropriate; I am still new to the code.)

The test is skipped by Travis (because of what sounds like an ImageMagick configuration error).

Does it work for Linux users?

earney:
PR #585 fixes this.  It just has not been merged into the repo yet.   Since I authored the PR, it is probably best for someone to merge (and someone elses eyes get a chance to look at it before it is merged).

Julian-O:
Great. Thanks. Will close and wait. (This merge won't be fun.)

Issue #596, opened 2017-06-29
__del__ is insufficient - need a close()


Julian-O:
I am working on refactoring some code where the original author uses moviepy. So, I am not too familiar with the code usage, sorry.

I am having some problems that echo the problems described in #101, #110, #164, and  #518. Examining the code, I believe they all have the same root cause.

Various classes (e.g. VideoFileClip) follow the same pattern:

* In the constructor they seize some resources (e.g. creating FFMPEG_VideoReader, which creates a process which locks the file.
* In the \_\_del\_\_ method, they release these resources (by calling del on sub-resources, and by terminating any subprocesses.).

I believe there are two misunderstandings here.

* Python's \_\_del\_\_ is not like C++'s destructors. There is no guarantee that they run immediately - or that they run at all. In practice, even though my code is deleting the VideoFileClip instance, the file remains open until the garbage collector gets around to it.
* On Windows, unlike Linux, you cannot delete a file that is in use by another process. So when Windows users complain that the file is still locked, the Linux developers are saying "It seems to work fine."

As a result, you are seeing Windows users complain that temporary files can't be deleted, and other users complaining that the number of "zombie" subprocesses is climbing as they continue processes.

The solution is:
* Don't rely on garbage-collection to manage the lifetime of non-memory resources. Add a close() method to each of the items that has resources, so clients can call them. By all means, continue to support \_\_del\_\_for when clients forget and the garbage collector notices.
* Please, while doing this, be even more pythonic: Support the context manager interface (\_\_enter\_\_ and \_\_exit\_\_), so people can write code in a way that close is always called, and the resources are always cleaned up, and it happens when they expect it.

---

The symptoms I am seeing, broadly:

Running Windows 10, Python 3.6.0. moviepy==0.2.3.2

* I create a file and put video data in it.
* I pass it to VideoFileClip(), and then use the instance to extract dimension data.
* I delete the VideoFileClip instance.
* I try to delete the file, but Windows complains it is still being used by another process.
* Then the \_\_del\_\_ methods are called by the garbage collector, during shut-down. (I know this, because I added some print statements to the moviepy code.

gyglim:
I second that. I'd love to have a context manager that takes care of the cleanup.
Current solution is to explicitly calling `__del__()`, which is not very nice

Zulko:
Thanks very much for the insight @Julian-O. Let's maybe wait for one or two other opinions, but if it was all good, would you be able to clean the code and submit a PR ?

[...]

PR #608, opened 2017-07-03
Issue #596: Add support for closing clips.

Julian-O:
* Add key support for close()

   * FFMPEG_VideoWriter and FFMPEG_AudioWriter: Support close() and context managers.
   * Clip: support close() and context manager. Doesn't do anything itself. The work is done in the subclasses that need it.
   * Clip subclasses: Overrride close.
       * Move away from depending on clients calling__del__(). Deleting can be left to Garbage Collector.
   * CompositeVideoClip: Note: Don't close anything that wasn't constructed here. The client needs to be able to control the component clips.
   * AudioFileClip:  Was concerned that lambda might include a reference to reader that wasn't cleaned up by close, so changed it over to an equivalent self.reader. Probably has no effect, but feels safer.

* Update tests to use close().

   * Note: While many tests pass on Linux either way, a large proportion of the existing unit tests fail on Windows without these changes.
   * Include changes to many doctest examples - Demonstrate good practice in the examples.
   * Also, migrate tests to use TEMPDIR where they were not using it.
   * test_duration(): also corrected a bug in the test (described in #598). This bug is also been addressed in #585, so a merge will be required.

* Add two new test files:

   *  test_resourcereleasedemo exercises the path where close is not called and demonstrates that there is a consistent problem on Windows. Even after this fix, it remains a problem that if you don't call close, moviepg will leak locked files and subprocesses. Because the problem remains until the process ends, this is included in a separate test file.]
   * test_resourcerelease demonstrates that when close() is called, the problem goes away.

* Update documentation to include usage tips for close()

Not included:

*  Example code has not been updated to use close().

### DIFF
--- a/moviepy/audio/io/ffmpeg_audiowriter.py
+++ b/moviepy/audio/io/ffmpeg_audiowriter.py
@@ -125,11 +125,27 @@ class FFMPEG_AudioWriter:
 
 
     def close(self):
-        self.proc.stdin.close()
-        if self.proc.stderr is not None:
-            self.proc.stderr.close()
-        self.proc.wait()
-        del self.proc
+        if self.proc:
+            self.proc.stdin.close()
+            self.proc.stdin = None
+            if self.proc.stderr is not None:
+                self.proc.stderr.close()
+                self.proc.stdee = None
+            # If this causes deadlocks, consider terminating instead.
+            self.proc.wait()
+            self.proc = None
+
+    def __del__(self):
+        # If the garbage collector comes, make sure the subprocess is terminated.
+        self.close()
+
+    # Support the Context Manager protocol, to ensure that resources are cleaned up.
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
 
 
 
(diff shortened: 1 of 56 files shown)


kencochrane:
Thank you for adding this, the code looks good to me, but I'm not a contributor, so we will need a contributor to review as well.

moviepy/audio/io/ffmpeg_audiowriter.py
```diff
@@ -135,12 +135,32 @@ def __del__(self):
...
+
+    # Support the Context Manager protocol, to ensure that resources are cleaned up.
+
     def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
         self.close()
 
+    def __exit__(self, exc_type, exc_value, traceback):
```
mgaitan: this es repeated, right?

Julian-O: You are right. Well spotted. That is an error. (It is of no consequence, but one should be deleted.)

Gronis:
Hi,

Why is this issue closed? Is it fixed somewhere else?

PR #589, opened 2017-06-16
Add test for detect_scenes

scherroman:

PR #587, opened 2017-06-14
Fixed Optional Progress Bar in cuts/detect_scenes

scherroman:
Progress bar was previously hard-coded to True.

PR #610, opened 2017-07-05
Use max fps for CompositeVideoClip

scherroman:
As of commit c0f6925, concatenate_videoclips uses the max fps of the
video clips.

This commit adds the same functionality for CompositeVideoClip.

### DIFF
--- a/tests/test_PR.py
+++ b/tests/test_PR.py
@@ -8,6 +8,7 @@ from moviepy.video.fx.scroll import scroll
 from moviepy.video.io.VideoFileClip import VideoFileClip
 from moviepy.video.tools.interpolators import Trajectory
 from moviepy.video.VideoClip import ColorClip, ImageClip, TextClip
+from moviepy.video.compositing.CompositeVideoClip import CompositeVideoClip
 
 sys.path.append("tests")
 from test_helper import TMP_DIR, TRAVIS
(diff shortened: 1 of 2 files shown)


tests/test_PR.py
```diff
@@ -112,5 +113,18 @@ def test_PR_529():
     assert video_clip.rotation == 180
 
 
+def test_PR_610():
```
kencochrane: I wonder if we should  have a better name and file name. Something more descriptive. In the future no one will remember what PR 610 was. Better to name if after what is also tested. Maybe composite_video_clip_max_fps?

scherroman: Good call

For a test like this, as I did in PR (#589), I think the best move is to name the test `test_composite_video_clip_should_use_max_fps` and put it under `tests/video/compositing/test_CompositeVideoClip`, which mirrors the folder structure of the library.

I think the underlying issues causing `tests/test_PR` to exist are
1. moviepy's tests aren't structured to mirror the folder structure of the library, so it can be difficult for a contributor to tell where to put a new test.
2. Sometimes PRs are very minor changes/bugfixes that one wouldn't normally write a test for (or are hard to write a test for), like fixing the optional progress bar for detect_scenes (#587) . However, because the PR is coming from an outside contributor and not a maintainer addressing a github issue, even if the changes look simple, there needs to be some level of assurance that the change A) does what it says it does and B) doesn't break the current, intended behavior of detect_scenes (especially when there's no existing test for detect_scenes yet, and its a hassle for a maintainer to go and test it manually).

PR #612, opened 2017-07-14
Fixed typo in the slide_out transition

kencochrane:
Fixed a typo in the slide_out transition, also cleaned up a little bit of whitespace in the file. I'm not sure if this actually works, but figured it was an easy fix.

Signed-off-by: Ken Cochrane <kencochrane@gmail.com>


kencochrane:
Not sure why it failed on travis for just one python version, I didn't change anything that should effect that.

PR #626, opened 2017-08-08
Support Windows platform in MoviePy

Julian-O:
## Refactor of setup.py and continuous integration.

This is a significant change to the way Travis and `setup.py` work.

The primary driving force for these changes: MoviePy is buggy on Windows and very difficult to install on Windows. I think the root cause is that it is not *tested* on Windows, so OS-specific code is added without any feedback to the developer.

These changes not only help MoviePy run on Windows and install more easily on Windows (still messy!) but more importantly allow it to be automatically tested on Windows, even for changes submitted by Linux developers.

## Changes

* **Includes several outstanding PRs**.
    * PRs #608,  #604, and #603 are incorporated into this PR, to give allow tests to pass on Windows.
* **Improve setup.py**
    * **Dependency Clean Up**
       * Support a *range* of versions in setup.py - particularly taking advantage of [Semantic Versioning](http://semver.org/), where available, to know which future versions will be compatible. This is especially useful when people are installing many packages or from binaries (which may have limited availability) - it prevents version clashes.
       * Remove unnecessary dependencies. For example, optional installs are not required for testing, so don't make it hard to test.
       * Add missing dependencies. Some dependencies were not recorded (e.g. youtube_dl)
       * Add Environment Markers. Not all packages support all platforms. Avoid trying to install them where it won't work.
   * Further integrate pytest into setup tests.
* **Improve Travis**
    * Move heavy lifting to `setup.py`, following DRY principle
         - Let `setup.py` do as much installation as possible.
             - This separation isn't perfect. A few packages need to be manually upgraded because Setuptools isn't good at resolving it needs to upgrade an existing installation.
             - `numpy` remains problematic.
         - Let `setup.py` run the tests themselves.
    * Exercise the installer in tests. Test not only the code, but the installer.
* **Introduction of Appveyor.**
    * Travis-CI doesn't support Windows, but Appveyor serves a similar role in the Windows space.
    * Unfortunately, the attempts to simplify Travis don't always work in the Windows world.
        * PIP won't install numpy, so conda is used.
        * Neither PIP nor conda will install ImageMagick (for ffmpeg)
* **Incidental fixes to platform-specific problems**
    * references to Python 3 exceptions.
    * references to Linux's /tmp directory.
    * upgraded downloader to use `requests` package, because old versions of `urllib` had outdated certificates, and were rejecting one of the downloads.

## Work to do

* Register the main branch for Appveyor, so every pull-request and check-in is tested on Windows automatically.
* Update the version number to a legal one, supported by setuptools. See [Semantic Versioning](http://semver.org/) for recommendations.
    - Then, change script to build and distribute Windows binaries to make installation simpler on Windows platforms.
* `Dockerfile` and `ez_setup` should not have been affected by this change, but re-testing them would be a good idea.
    - Automatically testing them would be a great idea.
* Add OSX support to the Travis script.
* Submit Windows testing to coveralls data. (Windows percentage should be higher because several tests are skipped in Travis.)


setup.py
```diff
@@ -57,15 +57,45 @@ def run_tests(self):
...
+    'decorator>=4.0.2,<5.0',
+    'imageio>=2.1.2,<3.0',
+    'tqdm>=4.11.2,<5.0',
+    'numpy',
+    ]
+
+optional_reqs = [
+        "opencv-python>=3.0,<4.0; python_version!='2.7'",
+        "scikit-image>=0.13.0,<1.0; python_version>='3.4'",
+        "scikit-learn; python_version>='3.4'",
```
jakirkham: Haven't looked at all of the changes here. So could be missing something. Is this suppose to be a test requirement still? Did it become an optional requirement as well?

Julian-O: I'm not sure I've understood what you are referring to, but I'll explain what I have done, and hope I answer your question through a scattergun!

MoviePy has a number of optional requirements - if we look at a feature like `resize()`, it will look for any of several different packages to provide the functionality it needs.

Previously all the optional requirements were considered part of the test requirements. If you asked `setup.py` to execute the tests, they were all installed. However, this was (a) unnecessary, and (b) painful to get to work on all platforms.

Now, the test_reqs doesn't contain as many requirements - it should only include requirements necessary to run the tests successfully. The tests are all succeeding.

The `.travis.yml` file does another test: that all the optional requirements can be installed via pip, so we know the `setup.py` doesn't name imaginary packages. I don't attempt that on the Windows version, as setuptools is incapable of installing all of the optional packages, such as scipy.

tburrows13:
These were all merged in with #630.

Issue #632, opened 2017-08-16
make_frame(t) after concatenation


timpauli:
First of all I wanted to say what a cool project moviepy is and thank everyone who participated in it.

I wrote an algorithm using moviepy which cuts a clip in very tiny bits and applies different effects to it and then concatenates them back together. 

Sometimes I get this error message while rendering in the end:

```
  File "<decorator-gen-51>", line 2, in write_videofile
  File "/usr/local/lib/python3.5/dist-packages/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, *a, **k)
  File "<decorator-gen-50>", line 2, in write_videofile
  File "/usr/local/lib/python3.5/dist-packages/moviepy/decorators.py", line 137, in use_clip_fps_by_default
    return f(clip, *new_a, **new_kw)
  File "<decorator-gen-49>", line 2, in write_videofile
  File "/usr/local/lib/python3.5/dist-packages/moviepy/decorators.py", line 22, in convert_masks_to_RGB
    return f(clip, *a, **k)
  File "/usr/local/lib/python3.5/dist-packages/moviepy/video/VideoClip.py", line 349, in write_videofile
    progress_bar=progress_bar)
  File "/usr/local/lib/python3.5/dist-packages/moviepy/video/io/ffmpeg_writer.py", line 209, in ffmpeg_write_video
    fps=fps, dtype="uint8"):
  File "/usr/local/lib/python3.5/dist-packages/tqdm/_tqdm.py", line 833, in __iter__
    for obj in iterable:
  File "/usr/local/lib/python3.5/dist-packages/moviepy/Clip.py", line 475, in generator
    frame = self.get_frame(t)
  File "<decorator-gen-14>", line 2, in get_frame
  File "/usr/local/lib/python3.5/dist-packages/moviepy/decorators.py", line 89, in wrapper
    return f(*new_a, **new_kw)
  File "/usr/local/lib/python3.5/dist-packages/moviepy/Clip.py", line 95, in get_frame
    return self.make_frame(t)
  File "/usr/local/lib/python3.5/dist-packages/moviepy/video/compositing/concatenate.py", line 83, in make_frame
    return clips[i].get_frame(t - tt[i])
  File "<decorator-gen-14>", line 2, in get_frame
  File "/usr/local/lib/python3.5/dist-packages/moviepy/decorators.py", line 89, in wrapper
    return f(*new_a, **new_kw)
[...]

Issue #631, opened 2017-08-11
Supersample Concatenated Clip


timpauli:
If you supersample a concatenated Clip you get the following error:

> File "/usr/local/lib/python3.5/dist-packages/moviepy/video/fx/supersample.py", line 12, in supersample
>     return clip.fl(fl)
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/Clip.py", line 136, in fl
>     newclip = self.set_make_frame(lambda t: fun(self.get_frame, t))
>   File "<decorator-gen-57>", line 2, in set_make_frame
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/decorators.py", line 14, in outplace
>     f(newclip, *a, **k)
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/video/VideoClip.py", line 694, in set_make_frame
>     self.size = self.get_frame(0).shape[:2][::-1]
>   File "<decorator-gen-14>", line 2, in get_frame
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/decorators.py", line 89, in wrapper
>     return f(*new_a, **new_kw)
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/Clip.py", line 95, in get_frame
>     return self.make_frame(t)
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/Clip.py", line 136, in <lambda>
>     newclip = self.set_make_frame(lambda t: fun(self.get_frame, t))
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/video/fx/supersample.py", line 9, in fl
>     avg = np.mean(1.0*np.array([gf(t_) for t_ in tt], dtype='uint16'), axis=0)
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/video/fx/supersample.py", line 9, in <listcomp>
>     avg = np.mean(1.0*np.array([gf(t_) for t_ in tt], dtype='uint16'), axis=0)
>   File "<decorator-gen-14>", line 2, in get_frame
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/decorators.py", line 89, in wrapper
>     return f(*new_a, **new_kw)
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/Clip.py", line 95, in get_frame
>     return self.make_frame(t)
>   File "/usr/local/lib/python3.5/dist-packages/moviepy/video/compositing/concatenate.py", line 82, in make_frame
[...]

PR #634, opened 2017-08-18
Issues #631 #632

inkeye:
Still needs to be discussed. I can't oversee all the implications of these changes.

### DIFF
--- a/moviepy/video/compositing/concatenate.py
+++ b/moviepy/video/compositing/concatenate.py
@@ -79,7 +79,7 @@ def concatenate_videoclips(clips, method="chain", transition=None,
 
     if method == "chain":
         def make_frame(t):
-            i = max([i for i, e in enumerate(tt) if e <= t])
+            i = max([i for i, e in zip(range(len(tt) - 1), tt) if e <= t])
             return clips[i].get_frame(t - tt[i])
 
         def get_mask(c):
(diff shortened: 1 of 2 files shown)


inkeye:
@uummoo 

running into 

> TypeError: only integer scalar arrays can be converted to a scalar index

using your solution.

tests/test_fx.py
```diff
@@ -206,7 +207,15 @@ def test_speedx():
...
-    pass
+    clip = VideoFileClip("media/big_buck_bunny_432_433.webm")
+
+    clip1 = supersample(clip, clip.duration, 1)
+    assert clip1.duration == clip.duration
+    clip1.write_videofile(os.path.join(TMP_DIR, "supersample1.webm"))
+
+    clip2 = supersample(clip, clip.duration / 2, 2)
+    assert clip2.duration == clip.duration
+    clip2.write_videofile(os.path.join(TMP_DIR, "supersample2.webm"))
```
koutoftimer, line 218: I don't know, maybe it is just me. But this test passes even without your changes. Which leads to question: what the reason of this PR.

I mean there is definitely a problem has been solve, but tests didn't reflect this fact. In other words there is no tests to reproduce an issue you have tried to solve.

koutoftimer (changes requested):

OsaAjani:

PR #648, opened 2017-09-27
Fix FFMPEG_VideoReader.get_frame()

MatthewShotton:
```self.pos``` in ```FFMPEG_VideoReader``` isn't updated when a frame is read with ```FFMPEG_VideoReader.read_frame()```. This causes ```FFMPEG_VideoReader.get_frame()``` to fail if seeking backwards because ```self.pos``` is never greater than the calculated new position (if seeking backwards to any other time than 0.)


tburrows13:
Superseded by #1220.

PR #652, opened 2017-10-05
fixed typo in library include

Goddard:
skimage.filters works for me not skimage.filter


tburrows13:
Thanks for this, I can confirm that the [docs](http://scikit-image.org/docs/dev/api/skimage.filters.html) definitely use `skimage.filters`.

PR #665, opened 2017-11-14
Removed Hz from audio_fps match in ffmpeg_parse_infos

qmac:
Cuts the ' Hz' off of the regex match when trying to find the fps info of an audio stream.

Previously, this was always setting `result['audio_fps']` to `'unknown'` because the casting to an integer would fail.

### DIFF
--- a/moviepy/video/io/ffmpeg_reader.py
+++ b/moviepy/video/io/ffmpeg_reader.py
@@ -385,7 +385,7 @@ def ffmpeg_parse_infos(filename, print_infos=False, check_duration=True,
         line = lines_audio[0]
         try:
             match = re.search(" [0-9]* Hz", line)
-            hz_string = line[match.start()+1:match.end()-3]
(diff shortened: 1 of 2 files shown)


qmac:
Coverage is only decreasing because https://github.com/Zulko/moviepy/pull/665/files#diff-a49583100a493b6fe1c05b48824e4a3cL389 is no longer executed on any of the test media files.

tburrows13:
Thank you very much for this!  I'll merge it right away

tburrows13:
Haha, just found this PR again. I've investigated, and you're definitely right, this is a bug. However, it doesn't appear that this `['audio_fps']` is used again by moviepy. Maybe it should be?

qmac:
Yes that may be true. Seems like `VideoFileClip`s get the `fps` from the meta-info, but `AudioFileClip`s rely on a user-provided `fps`, defaulting to 44100.

PR #666, opened 2017-11-17
Convert named lambda functions into defined functions

yuvallanger:
http://pep8.org/ says:

> Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier.
> 
> Yes:
> 
> `def f(x): return 2*x`
> 
> No:
> 
> `f = lambda x: 2*x`
> 
> The first form means that the name of the resulting function object is specifically `f` instead of the generic `<lambda>`. This is more useful for tracebacks and string representations in general. The use of the assignment statement eliminates the sole benefit a lambda expression can offer over an explicit def statement (i.e. that it can be embedded inside a larger expression)


tburrows13:
Thank you for the submission. However, this has been superseded by #1151.

PR #667, opened 2017-11-18
Removed wrong example from AudioFileClip documentation

v0idpwn:
There's an example using a bitrate argument. Which isnt supported, so I removed it.


tburrows13:
Thanks for this, but I'm about to do some further changes to AudioFileClip, so I'm closing this

PR #670, opened 2017-11-20
Use input seeking in ffmpeg_extract_subclip, not output seeking

scrpy:
`ffmpeg_extract_subclip` should not decode the frames preceding the start of the desired subclip, as this is slow. Input seeking makes more sense here than output seeking. See [seeking (ffmpeg wiki)](https://trac.ffmpeg.org/wiki/Seeking).


dhthompson:
Superseded by #882

scrpy:
Superseded by #882

PR #671, opened 2017-11-21
Update Readme.rst

rlphillips:
Updated readme to point out the name of the ImageMagick executable changed in the last version.


tburrows13:
Thanks for spotting this! I realised just after merging that it could be improved, so I've done that [in this commit](https://github.com/Zulko/moviepy/commit/f444f8cb6695272014b5b312eafeec6d6145d1b6).

Issue #640, opened 2017-08-29
bug in resize method. Mask is handled incorrectly when resize is called with lambda function.


TejasBob:
Hi,

The method for resizing clips located in  [resize.py](https://github.com/Zulko/moviepy/blob/master/moviepy/video/fx/resize.py) is buggy ( lines 103 - 113). I came across this bug while working on a video, I tried fixing it by looking at the source code.

`if hasattr(newsize, "__call__"):     

            newsize2 = lambda t : trans_newsize(newsize(t))

            if clip.ismask:

                fun = lambda gf,t: (1.0*resizer((255 * gf(t)).astype('uint8'),
                                                 newsize2(t))/255)

            else:

                fun = lambda gf,t: resizer(gf(t).astype('uint8'),
                                          newsize2(t))

            return clip.fl(fun, keep_duration=True,
                           apply_to= (["mask"] if apply_to_mask else []))
`

Take a look at the else part. The "fun" function code works correctly for image frames but fails for mask frames as mask frame values are in the range [0,1]. The transparency values less than 1 are converted to 0 (because of uint8 type casting). 

The fix to this problem is adding condition within lambda function to incorporate mask frames. I'm attaching code to fix this. 

My bug fix:
`        
        if hasattr(newsize, "__call__"):
            
            newsize2 = lambda t : trans_newsize(newsize(t))
            
            if clip.ismask:
                
                fun = lambda gf,t: (1.0*resizer((255 * gf(t)).astype('uint8'),
                                                 newsize2(t))/255)

            else:
                
                fun = lambda gf,t: resizer(gf(t).astype('uint8'),
                                      newsize2(t)) if len(gf(t).shape) == 3 else 1.0*resizer((gf(t)*255).astype('uint8'),newsize2(t))/255

            return clip.fl(fun, keep_duration=True,
                           apply_to= (["mask"] if apply_to_mask else []))`

Note the if else condition inside lambda function.

Tejas

kerstin:
Hi @TejasBob, just a general note about GitHub issues vs. code: the point of issues is to report problems; problem fixes, on the other hand, should be submitted as pull requests.

I.e. could you create a pull request containing your fix and reference this issue when creating it?

TejasBob:
I just created a pull request. 
Thanks!

alexzach:
[...]

PR #674, opened 2017-11-25
bug fix in resize method for a clip with mask

TejasBob:
I had previously reported an issue regarding the **bug in resize method for clips with mask** #640 .
I figured out the fix by reading fl method in Clip.py. 

Lambda function ( in original code ) on line 113 in resize.py is treating RGB image frame and mask frame in same manner. RGB image frame has it's values in range [0,255] while for a mask frame they are in range [0.0,1.0]. Applying uint8() on mask frame floors all values to 0 (and that is exactly the bug). 

I tried fixing the bug by checking the shape of frame returned by **get_frame** method. 
1. (gf(t)*255).astype('uint8')  makes it a valid opencv image (numpy uint8 image)
2. Then it is resized to newsize2.
3. Dividing the result by 255 brings the values back in range [0.0, 1.0]. 

Hope this helps!

Tejas

### DIFF
--- a/moviepy/video/fx/resize.py
+++ b/moviepy/video/fx/resize.py
@@ -111,7 +111,7 @@ def resize(clip, newsize=None, height=None, width=None, apply_to_mask=True):
             else:
                 
                 fun = lambda gf,t: resizer(gf(t).astype('uint8'),
-                                          newsize2(t))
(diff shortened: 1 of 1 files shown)


mondeja:
Closed due to lack of reproducibility. See [comment](https://github.com/Zulko/moviepy/issues/640#issuecomment-820551788) for further information. Feel free to reopen it if you can provide a minimal example showing the problem.

PR #684, opened 2017-12-12
Use popen.communicate() to prevent subprocess hang

tehKost:
Broken video files caused ffmpeg subprocess to hang


mgaitan:
could we merge this?

tburrows13:
It looks like this has already mostly been implemented elsewhere, however without the `terminate()`.

tburrows13:
It looks like this has already mostly been implemented elsewhere, however without the `terminate()`.

tburrows13:
Largely superseded by #926, but I'm going to merge this change which now adds `proc.terminate()`. I doubt that it is necessary, but I don't think that it can hurt.

PR #606, opened 2017-07-02
#600: Several YouTube examples in Gallery page won't load.

Julian-O:
* Used YouTube to get new iframes HTML.
* Replaced embedded image with link for private video.

Issue #600, opened 2017-07-01
Several YouTube examples in Gallery page are unable to load.


Julian-O:
* The embedded video for Cup Song Covers Mix reports "An error occurred. Please try again later." with a playback id. The [underlying YouTube video](https://www.youtube.com/rIehsqqYFEM) is fine.

* Similarly, the videos for:
   * The (old) MoviePy reel video.
   * Piano rolls transcription to sheet music
   * Videogrep

* Soccer cuts won't show because embedding has been banned. It can only be displayed in YouTube.

kerstin:
@Julian-O The video you linked results in a 404 on my end.

Zulko:
@Julian-O Thanks for the report. 4 videos are down for me too. Maybe there was a change in the youtube API ? Btw the address of the video you linked is:

 https://www.youtube.com/watch?v=rIehsqqYFEM

Julian-O:
Whoops. Sorry for broken link. The point I was hoping to convey was that the first few videos themselves remain up and working on YouTube. They haven't been taken down - it seems an API change is the most likely cause.

kerstin:
Looks like this can be closed? (Changes merged in #606?)

krishraghuram:
Has the bug resurfaced?

![screenshot from 2018-01-21 10-22-37](https://user-images.githubusercontent.com/10024339/35191042-22d47a82-fe97-11e7-815f-20de1d329dd1.PNG)

kerstin:
Hey @krishraghuram, maybe not all examples were fixed - did you check if this one in particular was included in the linked PR or earlier ones dealing with the same issue?

Julian-O:
@krishraghuram: Yes, several YouTube files are no longer displaying, when they used to. The code hasn't been reverted, so it seems likely that YouTube has changed the rules again. I recommend a new Issue. Would you like to raise it?

@kerstin: No, it isn't the case that some examples were missed. The "Cup Song Covers Mix" was fixed in #600, and the new code is still present, but it isn't displaying. Looks like a new problem has emerged.

Julian-O:
The merge does not seem to have been completely carried out.

For example, the changes to `gallery.rst` I proposed in #600 has [this embedding](https://github.com/Julian-O/moviepy/commit/7e3a0861da86c008ddaee28588b862027eaf5cc2) for the Cup Song:

[...]

PR #700, opened 2018-01-22
Make gallery fixes go live on github.io web-site.

Julian-O:
There are bugs in the Gallery code that were merged in #606 to the Master branch, but never made it to the gh-pages branch, so never became official. See #600 for more info.


tburrows13:
From what I can tell, this isn't what's needed to push the docs updates, and the PR is just getting larger with every commit so I'm going to close it until we have a definite solution.  @Zulko, any ideas?

PR #705, opened 2018-02-04
flake8 test to find syntax errors, undefined names

cclauss:


tburrows13:
Looks good to me.  I'm just merging the PRs you suggested, then I'll rerun the build here.

tburrows13:
After rerunning it, there are still errors: 5 in Python 3.6 and 2 in Python 2.  3 of the Python 3.6 ones are because they are Python 2 builtins (`cmp`, `execfile` and `unicode`).  Not sure what can be done about this...

The 2 that are common to py2 and py3 are a reference to `newfile` in `Clip.py` and `time_to_string` in `subtitles.py`.  The `newfile` one appears to still be legitimate due to the `@outplace` decorator (I have no idea how decorators work, so I may be wrong).  The `time_to_string` appears to be a genuine bug that I'll have a look at later.

So, is there a way of suppressing the py2/3 errors, because obviously we can't merge this as it stands right now.  It is a useful tool that I think would be good to have.

cclauss:
Give me 24 hours and I will submit more PRs.

tburrows13:
Great, thank you! No rush though

cclauss:
Please review this one because it would be good to get it merged in.

tburrows13:
Yep, this is really useful! Thanks

PR #729, opened 2018-02-28
Delete ez_setup.py

cclauss:
__ez_setup__ is no longer a recommended way to install Python code.


tburrows13:
I have zero experience with building libraries using `ez_setup.py`, `setup.py` etc, so I can't merge this one.  If we do go ahead with it though, we need to remove all references from `ez_setup.py` from the docs and readme.

Zulko:
@cclauss Can you provide more information on this ? From what I just found, Python3 still encourages the use of ``setuptools``, and ``setuptools``-dependant libraries still need an ``ez_setup.py`` file when installed on machines which don't have ``setuptools``.

Issue #493, opened 2017-03-13
FFMPEG not downloaded


kartikarora:
Until two days ago, imageio was downloading ffmpeg correctly. It is not doing so anymore.
```
ERR Traceback (most recent call last):
ERR   File "restapi.py", line 9, in <module>
ERR     from modules.location_from_media_module import location_from_media
ERR     from helpers.flicksys_audio_video import get_video, get_frame, make_url, delete_file
ERR   File "/home/vcap/app/helpers/flicksys_audio_video.py", line 5, in <module>
ERR   File "/app/.heroku/python/lib/python2.7/site-packages/moviepy/editor.py", line 22, in <module>
ERR     from .video.io.VideoFileClip import VideoFileClip
File "/app/.heroku/python/lib/python2.7/site-packages/moviepy/video/io/VideoFileClip.py", line 3, in <module>
ERR     from moviepy.video.VideoClip import VideoClip
ERR   File "/app/.heroku/python/lib/python2.7/site-packages/moviepy/video/VideoClip.py", line 20, in <module>
ERR     from .io.ffmpeg_writer import ffmpeg_write_image, ffmpeg_write_video
ERR   File "/app/.heroku/python/lib/python2.7/site-packages/moviepy/video/io/ffmpeg_writer.py", line 19, in <module>
ERR   File "/app/.heroku/python/lib/python2.7/site-packages/moviepy/config.py", line 38, in <module>
ERR     FFMPEG_BINARY = get_exe()
ERR   File "/app/.heroku/python/lib/python2.7/site-packages/imageio/plugins/ffmpeg.py", line 86, in get_exe
ERR     raise NeedDownloadError('Need ffmpeg exe. '
ERR imageio.core.fetching.NeedDownloadError: Need ffmpeg exe. You can download it by calling:
ERR   imageio.plugins.ffmpeg.download()
```

Details of deployed server:

IBM Bluemix CloudFoundry App
Runs Ubuntu 14.04 LTS Server
Moviepy installed via pip.

Adding the line `imagio.plugins.ffmpeg.download()` to my python code has no effect. Also, I'm not sure why it is looking for ffmpeg exe when it is running on a linux server.

Zulko:
[...]

PR #731, opened 2018-02-28
Added ffmpeg download when importing moviepy.editor

tburrows13:
See discussion in #493. Calling `imagio.plugins.ffmpeg.downlad()` doesn't install ffmpeg if it is already installed somewhere on the system.

### DIFF
--- a/moviepy/editor.py
+++ b/moviepy/editor.py
@@ -17,6 +17,8 @@ clip.preview().
 # Note that these imports could have been performed in the __init__.py
 # file, but this would make the loading of moviepy slower.
 
+import os
+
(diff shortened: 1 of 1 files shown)


Zulko:
Definitely necessary, but would this cause a problem for people running moviepy without direct internet access ? Such users would certainly have their own FFMPEG_BINARY, so maybe add something like this:

```
if os.getenv('FFMPEG_BINARY', 'ffmpeg-imageio') == 'ffmpeg-imageio':
    magio.plugins.ffmpeg.downlad()
```

mgaitan:
could we merge this?

tburrows13:
Yep thanks for the reminder.

PR #747, opened 2018-03-18
Make TextClip work on Travis CI

tburrows13:
I found a line in our Dockerfile that fixes the problem that ImageMagick has with TextClips.  This PR intends to run that command on Travis as well, so that all the tests using TextClips can be run.

The first commit just pushes the changes to the tests, to confirm that it still fails on Travis.


tburrows13:
I think I've got it working, thanks @earney for the [one-liner](https://github.com/Zulko/moviepy/blob/master/Dockerfile#L32) to fix ImageMagick!

PR #749, opened 2018-03-19
Fix tracking example

guhur:
The head blurring example was not working because it was using fxfy instead of trajectories

### DIFF
--- a/examples/headblur.py
+++ b/examples/headblur.py
@@ -1,36 +1,35 @@
-import pickle
-
 from moviepy.editor import *
-from moviepy.video.tools.tracking import manual_tracking, to_fxfy
+from moviepy.video.tools.tracking import manual_tracking
 
 
 # LOAD THE CLIP (subclip 6'51 - 7'01 of a chaplin movie)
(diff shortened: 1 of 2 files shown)


examples/headblur.py
```diff
@@ -1,45 +1,44 @@
...
+#                                 fps=6,
+#                                 nobjects=1,
+#                                 savefile="../../videos/chaplin_traj.txt")
 
 
 # IF THE MANUAL TRACKING HAS BEEN PREVIOUSLY DONE,
-# LOAD THE TRACKING DATA AND CONVERT IT TO FUNCTIONS x(t),fy(t)
+# RECOVER THESE TRAJECTORIES
+
+from moviepy.video.tools.tracking import Trajectory
```
mgaitan: we should import this at the top of the moduler, right?

moviepy/video/fx/headblur.py
```diff
@@ -12,7 +12,7 @@
 #-----------------------------------------------------------------------
 
 
-def headblur(clip,fx,fy,r_zone,r_blur=None):
+def headblur(clip,traj,r_zone,r_blur=None):
```
mgaitan: minor thing, but pep8 recommends space after each function argument,

Issue #745, opened 2018-03-16
defunct ffmpeg processes left even after clip.close()


TejasBob:
I tried following snippets in ipython terminal.

**Snippet 1**:

```
from moviepy.editor import *

clip = VideoFileClip(_video_path_).subclip(10,20).resize((200,100))

clip.write_videofile("result.mov", codec = "libx264", fps= 25)

clip.close()
```

After doing clip.close(),  I checked 
```
$ps -ef | grep ffmpeg
501 56157   343   0 12:55PM ttys000    0:00.00 grep ffmpeg
501 56134 56091   0 12:52PM ttys001    0:00.00 (ffmpeg)
```
There is one ffmpeg process left in memory.
I thought this happened because of loss of reference to ffmpeg process due to 2 outplace operations (subclip and resize), so I tried following script.

**Snippet2**:

```
from moviepy.editor import *
proc_list = []
clip = VideoFileClip(_video_path_)
proc_list.append(clip)

clip = clip.subclip(10,20)
proc_list.append(clip)

clip = clip.resize((200,100))
proc_list.append(clip)

clip.write_videofile("result.mov", codec = "libx264", fps= 25)

for ele in proc_list:
    ele.close()
```

Again I did,
`ps -ef | grep ffmpeg`

It gave similar output. There is one defunct process left in memory.

What is the right way of removing the ffmpeg instances?

Thanks!

tburrows13:
This is an problem with the main branch, that doesn't have a fix yet. It is discussed in #596. There is a fix in #608 that you can try using.

Hope this helps, feel free to ask again if you need more assistance.

TejasBob:
Thank you for references.

I noticed that, defunct ffmpeg processes are left only when the video being read has an audio associated with it. I managed to figure out that subclip() on the audio was causing the problem. 

I added outplace decorator to subclip method in Clip.py and then it stopped leaving zombie processes behind.

Original Clip.py:
```
@convert_to_seconds(['t_start', 't_end'])
@apply_to_mask
@apply_to_audio
def subclip(self, t_start=0, t_end=None):
    ...
```

After adding **outplace** decorator
```
@convert_to_seconds(['t_start', 't_end'])
@apply_to_mask
@apply_to_audio
@outplace
def subclip(self, t_start=0, t_end=None):
    ...
```

I'm not sure if it is a legit fix or just co-incidence.

TejasBob:
**Update:**

[...]

PR #750, opened 2018-03-20
defunct ffmpeg processes removed after clip.close()

TejasBob:
Defunct ffmpeg processes which were previously being left behind (even after doing clip.close() ) were associated with the audio of the file.

I had mentioned this issue at #745 

After comparing code in [ffmpeg_reader.py](https://github.com/Zulko/moviepy/blob/master/moviepy/video/io/ffmpeg_reader.py) and [readers.py](https://github.com/Zulko/moviepy/blob/master/moviepy/audio/io/readers.py), I noticed `self.proc.wait()` was missing in `close_proc` method in [readers.py](https://github.com/Zulko/moviepy/blob/master/moviepy/audio/io/readers.py)

So, I Added `self.proc.wait()` in `close_proc` method and that helped me get rid of defunct ffmpeg processes.

I'm not sure if someone else has already done it before.

Thanks!

### DIFF
--- a/moviepy/audio/io/readers.py
+++ b/moviepy/audio/io/readers.py
@@ -148,6 +148,7 @@ class FFMPEG_AudioReader:
             for std in [ self.proc.stdout,
                          self.proc.stderr]:
                 std.close()
+            self.proc.wait()
             self.proc = None
 
     def get_frame(self, tt):


do-mas:
This is some fix. Is there a plan to merge this PR?

tburrows13:
Thanks for the fix. However, this has been implemented in #1185.

Issue #649, opened 2017-10-01
Trying to use slide_in transition


cquintini:
I'm trying to use slide in and out transitions between  videoclips with no luck I was only available to make it work using crossfadein like this:

     from moviepy.editor import VideoFileClip, concatenate_videoclips, vfx, afx, transfx, concatenate, 
     CompositeVideoClip

     output = "out.mp4"
     delay = 1
     clip1 = VideoFileClip("174_3.mp4")
     clip2 = VideoFileClip("174_4.mp4")
     clip3 = VideoFileClip("174_5.mp4")
     clip4 = VideoFileClip("174_6.mp4")

	final = concatenate([clip1,
                     clip2.crossfadein(delay),
                     clip3.crossfadein(delay),
                     clip4.crossfadein(delay)],
             padding=-delay, method="compose")
	final.write_videofile(output)

But when I use the example on  https://github.com/Zulko/moviepy/blob/3d86a2c4cba6769528112765d964eaec273d7636/moviepy/video/compositing/transitions.py

     from moviepy.editor import VideoFileClip, concatenate_videoclips, vfx, afx, transfx, concatenate, 
     CompositeVideoClip

     output = "out.mp4"
     clip1 = VideoFileClip("174_3.mp4")
     clip2 = VideoFileClip("174_4.mp4")
     clip3 = VideoFileClip("174_5.mp4")
     clip4 = VideoFileClip("174_6.mp4")

     clips = [clip1, clip2, clip3, clip4];
     CompositeVideoClip(clips)

     slided_clips = [clip.fx( transfx.slide_out, 1, 'bottom') for clip in clips]
     final_clip = concatenate( slided_clips, padding=-1)
     final_clip.write_videofile(output) 

This result is videos concatenated but not transition at all, please let me know what I'm doing wrong

mivale:
This would indeed be very helpful. Is there really nobody who knows how to execute this functionality?

The docs say nothing about this, and the source is, ehrm, not too helpful in this instance ;-)

knezi:
I'm copying my answer from https://stackoverflow.com/questions/46517510/moviepy-slidein-and-slideout-transitions/49617291#49617291:

Remember that most functions in moviepy do not modify the object, but return a modified shallow copy of the object - so the line CompositeVideoclip(clips) does not do anything.

Instead, you want to do:

```
slided_clips = [CompositeVideoClip([clip.fx( transfx.slide_out, 1, 'bottom')]) for clip in clips]
```

[...]

PR #753, opened 2018-03-25
Bug-fix: Slide_out: fixed computation of position

knezi:
Slide_out didn't do anything before.

Can you please estimate when this gets into a release?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -42,7 +42,7 @@ In this example we open a video file, select the subclip between t=50s and t=60s
 Installation
 ------------
 
-MoviePy depends on the Python modules Numpy_, imageio_, Decorator_, and tqdm_, which will be automatically installed during MoviePy's installation. The software FFMPEG should be automatically downloaded/installed (by imageio) during your first use of MoviePy (installation will take a few seconds). If you want to use a specific version of FFMPEG, follow the instructions in ``config_defaults.py``. In case of trouble, provide feedback.
+MoviePy depends on the Python modules Numpy_, imageio_, Decorator_, tqdm_, and requests_, which will be automatically installed during MoviePy's installation. The software FFMPEG should be automatically downloaded/installed (by imageio) during your first use of MoviePy (installation will take a few seconds). If you want to use a specific version of FFMPEG, follow the instructions in ``config_defaults.py``. In case of trouble, provide feedback.
 
 **Installation by hand:** download the sources, either from PyPI_ or, if you want the development version, from GitHub_, unzip everything into one folder, open a terminal and type:
 
--- a/moviepy/video/compositing/concatenate.py
+++ b/moviepy/video/compositing/concatenate.py
@@ -76,6 +76,7 @@ def concatenate_videoclips(clips, method="chain", transition=None,
     h = max([r[1] for r in sizes])
 
     tt = np.maximum(0, tt + padding*np.arange(len(tt)))
+    tt[-1]-=padding # last element is the duration of the whole
 
     if method == "chain":
         def make_frame(t):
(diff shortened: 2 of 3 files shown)


tburrows13:
Hi, thanks for this. Do you think this is related to #649? Is there a similar problem in `slide_in`?

tburrows13:
I'm just trying this fix out and I don't think it works, but I may be misunderstanding the function.

I'm using this code, taken largely from the example in the docstring:

```from moviepy.editor import *
c1 = ColorClip((1500, 1000), color=(255, 0, 0))
c2 = ColorClip((1500, 1000), color=(255, 255, 0))
c3 = ColorClip((1500, 1000), color=(0, 255, 0))
clips = [c1, c2, c3]
sclips = [clip.set_duration(4).fx(transfx.slide_out, 1, 'bottom') for clip in clips]
c = concatenate_videoclips(sclips, padding=-1)

c.write_videofile("test.mp4", fps=24)
```

Can I see your code please?

knezi:
In the doc string there's written you have to use it inside CompositeVideoClip. My code:

```
mask=ColorClip(clip2.size, color=1, ismask=True, duration=clip2.duration)\
                        .fx(transfx.slide_in, \
                                duration=tr_time, side=fr)
mask=CompositeVideoClip([mask], ismask=True, size=clip2.size)
clip2=clip2.set_mask(mask)
```


But you don't have to use it in mask obviously. So if you do something like: [CompositeVideoClip([clip.....]) for clip in clips], it should work.

No, #649 is not related to this. Slide_in works, but again, you have to include it in CompositeVideoClip.

knezi:
I added the fixed examples. (And missing required dependency).

tburrows13:
Thank you for editing the examples, that’s really helpful. Why have you mentioned requests as a dependency? According to setup.py, it’s only needed when running the test suite.

knezi:
When I cloned the git repository and ran `import moviepy` in python, I got an error that requests is missing. However, I'm not getting this anymore. Sorry, please disregard this commit.

knezi:
I'm still trying to understand how pull requests work on github. Sorry for confusion. My last commit was automatically added here. So, commits `slide_out` and `fix examples` are correct and they're fixing what the PR is about.
The commit `Added dependency` is wrong and shall be disregarded.
My last commit `Fix: concatenate` is correct and fixes different part of code.

mgaitan:
hey guys. IMHO this patch is very important and almost done. would you like I reopen it squashing the correct commits? or could you @tburrows13 do it?

knezi:
@mgaitan Yes, I feel the same. I can also do it if you can guide me what to do.

mgaitan:
@knezi what you could do is to rebase your branch against the parent of your first commit. 

```
 git rebase -i 6cbd4f34773
```

this will open your editor with the list of commits and you can pick / squash / drop them following the intructions commented in the same screen.

As this operation will assigns new hashes for the resulting commits (i.e it rewrites the history) you will need to force the push with "-f".  

Alternatively, create a new branch from `master`, and cherry-pick the right commits. Then close this PR and reopen a new one from the clean branch. 

(Btw, it's always a good practice to create a branch for your feature/bugfix and avoid to open prs from your master to the official's master branch.)

tburrows13:
Yep, we need to clean up this PR before merging, or start anew. What @mgaitan sugggested looks good. @knezi, if you need a hand with working with git, feel free to post on the [moviepy Gitter channel](https://gitter.im/movie-py/Lobby) (though it’s currently down for maintenance 😔). We try to have a separate PR for each feature/bugfix.

PR #766, opened 2018-04-16
add a decorator to quit pygame

### LINKED ISSUE #575: On windows, a preview doesn't close when you quit.
I'm using windows, Python 3.6 and moviepy 0.2.3.2, run through jupyter notebook.

clip.ipython_display() runs but doesn't produce any visible result.

clip.preview() pops up a PyGame window. The command ends on ESC. But the window doesn't pop down. I can close it by importing pygame and calling pygame.quit()

mgaitan:
This fix #575


tburrows13:
I'm running:

```
clip = VideoFileClip("media/big_buck_bunny_0_30.webm").subclip(0,1)
clip.preview()
time.sleep(10)
```

with this PR and, after playing the video for a second, it crashes (doesn't sleep for 10 secs):

```
2018-04-22 18:58:17.403 python[2873:132770] WARNING: nextEventMatchingMask should only be called from the Main Thread! This will throw an exception in the future.
  File "tom_t.py", line 7, in <module>
    clip.preview()
  File "<decorator-gen-102>", line 2, in preview
  File "/Users/tomburrows/Python/moviepy-dev/moviepy/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, *a, **k)
  File "<decorator-gen-101>", line 2, in preview
  File "/Users/tomburrows/Python/moviepy-dev/moviepy/moviepy/decorators.py", line 22, in convert_masks_to_RGB
    return f(clip, *a, **k)
  File "<decorator-gen-100>", line 2, in preview
  File "/Users/tomburrows/Python/moviepy-dev/moviepy/moviepy/decorators.py", line 142, in pygame_quit
    value = f(*a, **k)
  File "/Users/tomburrows/Python/moviepy-dev/moviepy/moviepy/video/io/preview.py", line 135, in preview
    for event in pg.event.get():
pygame.error: video system not initialized
```

When run on the master branch, after a second of video, the last frame just sits there for 10 seconds, then disappears.

mgaitan:
I will check, thanks for test it.

tburrows13:
It appears that in the past we've never imported pygame in the test environment, so here its raising an error because its not installed.

tburrows13:
Ok, I've removed the tests for now, pygame will require some setting up with SDL on Travis, and I have no idea if that would ever work.

mondeja:
This could be a good idea, but:

- Doesn't solves the problem fixed by #1537 of quitting pygame when a key is pressed.
- Catching all whatever pygame error passing it seems to me very risky.
- The code and the testing environment is conflicting a lot with current codebase.

For these reasons, I'm closing this to reduce the future reviewing work. Feel free to update this code and reopen the pull if you think that could be interesting :+1:

PR #785, opened 2018-05-02
Add graphics blend modes

abeedvisram:
*Zerothly, thanks so much to @Zulko et al. for all the amazing work that's gone into MoviePy, it really is a fantastic library.

Firstly, apologies for a pull request out of the blue. I appreciate that there should be some discussion and acceptance prior to adding a feature, but this is something that we've added and are using internally so figured it would be worth pushing upstream to see if the community was interested in adopting it. Secondly, I couldn't see a contributing guide, so let me know what we've missed / how to go about this in the future.

### Motivation

Recently we ran into a situation where we wanted to reproduce blend / mix modes in a video, see e.g. https://en.wikipedia.org/wiki/Blend_modes . We came across the [blend_modes package](https://github.com/flrs/blend_modes) which seemed to have taken care of the details, but wanted to integrate it into MoviePy to take into account a varying background video.

#### New class -- BlendedCompositeVideoClip

We didn't want to risk breaking any of the existing workflows that users might have had, so decided to isolate all of the functionality into a new clip class. This sub-classes from `video.compositing.CompositeVideoClip` but modifies the `make_frame` and `playing_clips` methods to account for the blend mode.

#### Optional dependency

As this feature requires an additional dependency then we wanted to make it optional and not force users to have to install an extra library if they didn't need it. As such, an error is only raised when trying to import the new class without blend_modes being installed.

Blend modes is MIT licensed.

#### Modifying blit_on and blit functions

In order to expose the blend_modes functionality with the minimum of impact it was necessary to change the above-mentioned method signatures. We have added in:
- blend_mode -- this corresponds to either "normal", i.e. the existing MoviePy behaviour, or to any of the blend_modes defined in the package.
- blend_opacity -- opacity of the layer, passed through to blend_modes
- blend_weight -- allows for weighting between the specified blend_mode and the "normal" mode

### Output

Forgive the gif quality, but below is a sample output using the soft_light and hard_light blending modes. This is the output of the test that we've added. Note that as far as I could tell, there was no way to actually check that the output of the tests matches what is expected, only that the files render. Is this sufficient?

blended_composition

The background video was taken from https://videos.pexels.com/videos/blurred-bokeh-video-855204 and is CC0 licensed. This is required for the test and is uploaded below:

[blend_modes.zip](https://github.com/Zulko/moviepy/files/1967962/blend_modes.zip)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -92,6 +92,7 @@ For advanced image processing, you will need one or several of the following pac
 - `Scikit Image`_ may be needed for some advanced image manipulation.
 - `OpenCV 2.4.6`_ or a more recent version (one that provides the package ``cv2``) may be needed for some advanced image manipulation.
 - `Matplotlib`_
+- `Blend Modes`_ for using the BlendedCompositeVideoClip for [blend modes](https://en.wikipedia.org/wiki/Blend_modes)
 
 Once you have installed it, ImageMagick will be automatically detected by MoviePy, (except for windows users and Ubuntu 16.04LTS users).
 
@@ -198,6 +199,7 @@ Maintainers
 .. _ImageMagick: http://www.imagemagick.org/script/index.php
 .. _`Matplotlib`: https://matplotlib.org/
 .. _`Sphinx`: http://www.sphinx-doc.org/en/master/setuptools.html
+.. _`Blend Modes`: https://github.com/flrs/blend_modes
 
 .. People
 .. _Zulko: https://github.com/Zulko
--- a/moviepy/editor.py
+++ b/moviepy/editor.py
@@ -31,6 +31,10 @@ from .video.io.ImageSequenceClip import ImageSequenceClip
 from .video.io.downloader import download_webfile
 from .video.VideoClip import VideoClip, ImageClip, ColorClip, TextClip
 from .video.compositing.CompositeVideoClip import CompositeVideoClip, clips_array
+try:
+    from .video.compositing.BlendedCompositeVideoClip import BlendedCompositeVideoClip
+except ImportError:
+    pass
 from .video.compositing.concatenate import concatenate_videoclips, concatenate # concatenate=deprecated
 
 from .audio.AudioClip import AudioClip, CompositeAudioClip, concatenate_audioclips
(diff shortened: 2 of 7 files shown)


tburrows13:
Wow! Thanks for doing this. I’m no video expert, so I won’t pretend to understand everything that this does from that perspective. The code looks well written, with PEP8 and suitable comments; well done.

I think we need more documentation on this first (For `BlendedCompositeVideoClip`, and consider adding an example in the examples section.

Zulko:
That's a great feature, I think it should be either completely fused to CompositeVideoClip, or completely separated, in its own file, without touching any core file, like a "plug-in".

At least in a first time, because it is not a mainstream use case, I would be for the later option, which would consist of a self-contained ``BlendedCompositeVideoClip.py`` file, located in ``moviepy.video.compositing`` and which would contain the BlendedCompositeVideoClip class, and a special function blended_blit (instead of making the central ``blit`` method more complex).

Users would call it like this:
```
from moviepy.video.compositing import BlendedCompositeVideoClip
...
```
The class would raise an error at run time if blend_mode is not installed.

Some remarks:

- As Tom said, it needs a bit more documentation. Could you also help for documentation in Sphinx ?
-  Avoid ``*args, **kwargs``, it makes the library difficult to understand. Even though it is a chore, you will need to copy the parameters and docstrings of CompositeVideoClip.
- I like it when import error messages also suggest a pip command, like "install xxx with 'pip install blend_modes' and try again".

abeedvisram:
Thanks for taking the time to look over it and for the feedback:

- [ ] Will definitely add more documentation and an example.
- [ ] Will ensure that it's completely separated out and look into a separate `blended_blit` function, might be slightly involved to ensure that it works with any `Clip` sub-class that can be thrown at it but should be fine.
- [ ] Will copy parameters from the parent `CompositeVideoClip`.
- [ ] Will update the error message on import to advise how to fix.

We'll make these changes and get back to you shortly.

@tburrows13 regarding the video for the test suite I saw that they were being pulled from @earney 's moviepy_media repo. I've included the video file in the zip archive at the end of the first comment (https://github.com/Zulko/moviepy/files/1967962/blend_modes.zip) but can create a separate PR to add it in there if that's the done thing.

kyb3r:
When will this get merged?

tburrows13:
Closing in favour of #1152. It is a neat feature, but likely not high priority for the time being.

PR #791, opened 2018-05-06
Removed import of download_webfile so that requests is not required

tburrows13:
I just tried installing moviepy on a fresh system, and upon importing `moviepy.editor` I got an error saying that `requests` was required. It looks like @knezi ran into the same problem in #753. This PR removes the import that caused it.

### DIFF
--- a/moviepy/editor.py
+++ b/moviepy/editor.py
@@ -6,7 +6,7 @@ MoviePy by simply typing:
 
 In particular it will load many effects from the video.fx and audio.fx
 folders and turn them into VideoClip methods, so that instead of
->>> clip.fx( vfx.resize, 2 ) # or equivalently vfx.resize(clip, 2)
+>>> clip.fx(vfx.resize, 2) # or equivalently vfx.resize(clip, 2)
 we can write
 >>> clip.resize(2)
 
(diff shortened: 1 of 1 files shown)


Zulko:
That is very concerning, why would the tests work if a fresh install gives a buggy moviepy ?
I'd be more for adding request as a dependency in that case, download_webfile is super useful for examples.

tburrows13:
The tests work because requests is listed as a dependency for testing, so it is installed by travis ([Travis line](https://github.com/Zulko/moviepy/blob/master/.travis.yml#L39), [setup.py line](https://github.com/Zulko/moviepy/blob/master/setup.py#L91). When a user installs moviepy, the test dependencies aren't installed, so if they don't already have requests installed, importing moviepy.editor throws an error.

So, we either need to add reqests as a dependency (which I am fine with; most people would already have it installed) or merge this PR which just means that in order to use `download_webfile`, the user has to write `from moviepy.video.io.downloader import download_webfile`. This change would not be backwards compatible.

Zulko:
Wow so Moviepy has been broken for the last month ! Thanks for catching that. I vote for adding request to the dependencies. Also many thanks for all the efforts keeping Moviepy afloat btw (I am swamped between work and open source projects which i need on a daily basis, leaves me no brain space for non-trivial moviepy issues).

PR #795, opened 2018-05-11
Fix slide out

knezi:
See #753 - first part of the PR
Slide_out didn't do anything before.


knezi:
I think it's not possible to merge this because the changes in comit slide_out were commited and in the next commit reverted. Hence it should be alright to merge this (I checked it in the code).

keikoro:
@tburrows13 Can this be merged? I saw you had a look at this in the past.

Overdrivr:
Thanks for your contribution, could you add some more information regarding this issue and also some test cases for reproducing the bug and showing fix works as intended ?

knezi:
@Overdrivr Well, there's an example in the dosctring (commit https://github.com/Zulko/moviepy/pull/753/commits/50ce30ad4c52c9e6a12a824ce07dfa323a8768b3) plus in the last PR https://github.com/Zulko/moviepy/pull/753#issuecomment-378014303 (but of course replace slide_int with slide_out.

Overdrivr:
It's not entirely clear what the problem is, to me at least, even with that information.

Did using slide_out before had zero impact on the video ?
Also, why is slide_in used in the slide_out function example ?

tburrows13:
I've given this a test and I can confirm that before the fix it didn't do anything, but with the fix it does work. However, I can't get the example that you've added into the docstring to work. They do slide in/out but there is no overlap between the two as you'd expect given that there is negative padding. I got it to work by putting the list of clips inside a CompositeVideoClip:

    v1 = VideoFileClip("test.mp4").fx(transfx.slide_out, duration=2, side='left')
    v2 = VideoFileClip("test2.mp4").set_start(2)

    final = CompositeVideoClip([v2, v1])  # The other way round so that v1 appears on top

    final.write_videofile(f"slideout_left.mp4")

However, this requires manually setting the start time for the second clip based on the length of the first clip minus the duration of the slide out. Any suggestions?

tburrows13:
I'm going to merge this as it is. The documentation is no worse than it was before and it appears that we are going to somewhat rethink the fx api for v2.0 anyway.

PR #796, opened 2018-05-11
FIX: concatenate - duration of the result

knezi:
see #753 
second part of the bad PR.


tburrows13:
Would you mind giving an example piece of code that only works correctly with this code please? Thanks

knezi:
Well, just this:
```
a=VideoFileClip('something')
b=VideoFileClip('something')
concatenate_videoclips([a,b]).write_videofile('something')
```

Or do you mean into the code?

tburrows13:
What I mean is a piece of code that doesn't do what it should unless this patch has been applied. I've looked into this one, and I agree that there is a bug. Here are 2 pieces of code that demonstrate the bug:

```
from moviepy.editor import *
a = ColorClip((500, 750), color=(0, 255, 0)).set_fps(24).set_duration(10)
b = ColorClip((750, 500), color=(255, 255, 0)).set_fps(24).set_duration(10)
c = concatenate_videoclips([a, b], method="compose", padding=3)
d.write_videofile("padding3.mp4")
```
`d` should have a duration of 23 (10 + 3 + 23), but without this fix applied, it has a duration of 26, as the blank gap in the middle lasts for 6 seconds.

```
from moviepy.editor import *
a = ColorClip((500, 750), color=(0, 255, 0)).set_fps(24).set_duration(10)
b = ColorClip((750, 500), color=(255, 255, 0)).set_fps(24).set_duration(10)
c = ColorClip((600, 600), color=(0, 255, 255)).set_fps(24).set_duration(10)
d = concatenate_videoclips([a, b, c], method="compose", padding=3)
d.write_videofile("padding3.mp4")
```
With 3 clips, it is even stranger. It should have a duration of 36 (10 + 3 + 10 + 3 + 10) but it has 39 instead; there is an extra 3 seconds of black at the end.

Well done for spotting this, and as far as I can tell, your fix does work.

knezi:
@tburrows13 O, I see. Thanks for providing it.

PR #836, opened 2018-07-30
Add cython to requirements

focom:
Without this package the image wont build, it fixed my issue. So it might be worth sharing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN apt-get install -y locales && \
 ENV LC_ALL C.UTF-8
 
 # do we need all of these, maybe remove some of them?
-RUN pip install imageio numpy scipy matplotlib pandas sympy nose decorator tqdm pillow pytest requests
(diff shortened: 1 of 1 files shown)


tburrows13:
@jonnekaunisto I wonder what you think of this one? It has 2 thumbs up so I guess it is not just a one-off fix. Is it possible that a given docker image can work for some people but not others? (I've never used it).

jonnekaunisto:
I am not entirely sure why it didn't work for them without Cython. The same docker image should work exactly the same on every machine, but the Dockerfile that is in moviepy right now doesn't have a tagged version for Python or Debian such as what #1175 had in its first commit. 

Basically every time you build the image it will get the latest Python and Debian version, which will keep changing as new versions come out.

So the CPython might have been needed in 2018, with the latest release of Debian and Python then, but it doesn't seem to be needed anymore.

It might be good to tag the Dockerfile with some Python and Debian version so that it doesn't break when a new version of either of those comes out.

tburrows13:
Ok, I'll close this, and perhaps you could check that it works on the most recent version of Python and Debian and create a PR with those as tags.

PR #837, opened 2018-07-31
Adding installation command for Windows

cheezeburger:


tburrows13:
Closing because this describes the exact same method for Windows that is already in the README (which currently implicitly covers both Unix and Windows).

Issue #847, opened 2018-09-06
frozen seconds in beginning of subclip using ffmpeg_extract_subclip()


grszkthfr:
Hi everyone,

not sure if this is a bug, however I have an issue here:

The [docs of ffmpeg state](https://trac.ffmpeg.org/wiki/Seeking#Inputseeking), that `-ss` should be specified before `-i`, at least when input seeking. 

But I think `ffmpeg_extract_subclip()`, specifies it after `-i`. That is at least what causes trouble when I call `ffmpeg_extract_subclip()`

For myself:

```
~\AppData\Local\imageio\ffmpeg\ffmpeg-win32-v3.2.4.exe -y -i raw\pt01c.avi -ss 33.00 -t 40.00 -vcodec copy -acodec copy corridor\pt01-corridor.avi
```
leads to a frozen video sequence of 2-6 seconds in the beginning of the video with sound playing.

Whereas 

```
C:\Users\jog54yy\AppData\Local\imageio\ffmpeg\ffmpeg-win32-v3.2.4.exe -y  -ss 33.00 -i raw\pt01c.avi  -t 40.00 -vcodec copy -acodec copy corridor\pt01-corridor.avi
```
works as expected.

Is there any thing that I am missing? Is that how it is thought to be? Is there a way to specifiy the order of specification of `-i`and `-ss`?

Edit: Looks like Issue [#508](https://github.com/Zulko/moviepy/issues/508#issuecomment-344982840) but [this](https://github.com/Zulko/moviepy/pull/579#issuecomment-344982313) did not help.

Best Jonas

__________________________________________

Here's the rest:

### Expected Behavior

No frozen seconds before video starts playing.

### Actual Behavior

`ffmpeg_extract_subclip()` produces frozen seconds before video starts playing, sound is good form the beginning. However, couldn't test for synchrony (That is what [other](https://superuser.com/questions/138331/using-ffmpeg-to-cut-up-video) report as well, when specifing it the wrong way).

### Steps to Reproduce the Problem
 see above

### Specifications

  - Python Version: 3.7.0
  - Moviepy Version: 0.2.3.5
  - Platform Name: Windows
  - Platform Version: Windwos 10

Zulko:
You may be right. There is no way to choose the order of -i nd -ss in moviepy, but if one works and the other not, then its a bug and needs a PR. I am too busy to test it unfortunately, but I think we can trust you on this.

gpotter2:
[...]

PR #848, opened 2018-09-06
FIX changed order of specifications -ss befor -i for ffmpeg_extract_subclip()

grszkthfr:
Refers to #847 


gpotter2:
Hi ! It seems I was late on this issue :smile: you were here before me and my #882

Here is some doc I could gather, showing that this fix is the correct one:

6e78421b-086a-4db5-94b9-5389977f4040

> The fix (I tested it multiple times with various parameters) is directly inspired from  https://stackoverflow.com/a/33188399/5459467 which quotes the above doc.

PR #851, opened 2018-09-23
Audio Time Stretching

tomatitito:
I needed to slow down a couple of videos as well as stretch the audio, so that's what I added to moviepy. I added the feature as a new effect in audio/fx. The video/fx/speedx feature now has a flag to indicate wether time stretching should be done, the default is False.

As a bit of a drawback, I used the time stretching algorithm from the librosa project. That means I effectively created another dependency.


tests/test_fx.py
```diff
@@ -226,6 +227,15 @@ def test_normalize():
     clip = audio_normalize(clip)
     assert clip.max_volume() == 1
 
+def test_time_stretch():
+    clip = AudioFileClip("media/crunching.mp3")
+
+    clip1=time_stretch(clip, factor=0.5)
+    assert abs(clip1.duration - 2*clip.duration) < 0.1
```
Overdrivr: Could you replace this by [`math.isclose`](https://docs.python.org/3/library/math.html#math.isclose) ? If I'm not mistaken, it should be identical.

tests/test_fx.py
```diff
@@ -226,6 +227,15 @@ def test_normalize():
...
+
+    clip2=time_stretch(clip, factor=1.5)
+    assert abs(1.5*clip2.duration - clip.duration) < 0.1
```
Overdrivr: Same here

tests/test_fx.py
tomatitito: Thanks for your feedback! I made the changes.

tests/test_fx.py
tomatitito: Same here.

mondeja (changes requested):
Thank you for opening this pull @tomatitito. Before start reviewing the new feature, could you remove the `Pipfile` and the new added gitignore lines? Keep in mind that you can run `git add <filename.ext>` instead of `git add .` when adding files to the commit, so you doesn't need to add new files to gitignore if there aren't strictly needed.

tomatitito:
Thank you, @mondeja for your feedback. Sorry for the Pipfile, I messed up. I fixed some stuff so that the tests are green at least on my machine. I had to call my function qualified, though, as `time_stretch.time_stretch(clip, ...)`, otherwise I would get an error about the module not being callable. It seemed to me though, that all other audio fx functions can be called unqualified, even though I imported mine the same way. Any hints how I could fix this to be more in line with the rest of the moviepy library? 

Also, obviously, the ci pipeline is still not successful. I'll look into this...

mondeja:
> Any hints how I could fix this to be more in line with the rest of the moviepy library?

Just import it in `moviepy.audio.fx.__init__`: `from moviepy.audio.fx.time_stretch import time_stretch`.

mondeja:
Sorry, but we can't depend on more system-level libraries like `libsndfile`, at least for now because it adds a lot of work for us maintaining the installation (and keeping it easiest as possible) for all platforms. Perhaps you can stole the `librosa` `time_stretch` function logic from their codebase?

tomatitito:
Thanks again, @mondeja. I'll try to get rid of librosa.

OsaAjani:

PR #860, opened 2018-10-17
Fix time skip issue at end of frames iteration

superseed:
Numpy's `arange` returns evenly spaced number over a half-open range. This causes issues in the `make_frame` callable when trying to compute the current time in the clip as the ratio of the absolute current time over the total duration: the ratio will never reach 1. It will actually reach a value that depends on the number of fps specified. The only possible solution to ensure the ratio reaches 1 is then to round the ratio to two decimal places or less, as the last time value reached will otherwise be `(duration - (1 / fps))`, i.e. `(duration - 0.01666...)` for 60 fps or even `(duration - 0.03333...)` for 30 fps.

We can however avoid this situation entirely by swapping `arange` with `linspace`, as this function will return evenly values numbers over a *closed* interval (i.e. it will end with the exact end value) and thus ensure the ratio goes from exactly 0 to exactly 1.

Note: as a bonus, the step is currently a non-integer (`1.0/fps`) and using linspace is actually what the [Numpy docs](https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.arange.html) suggest:

> When using a non-integer step, such as 0.1, the results will often not be consistent. It is better to use linspace for these cases.


Overdrivr:
Thanks for your contribution. Could you please:
- add a test case for reproducing the bug and ensuring the fix works as expected
- update your branch with latest changes on master

tburrows13:
Superseded by #1195.

PR #865, opened 2018-10-29
Add pytweening dep to Dockerfile

AquiGorka:
To be able to use the Tweening library with the Docker container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN apt-get install -y locales && \
 ENV LC_ALL C.UTF-8
 
 # do we need all of these, maybe remove some of them?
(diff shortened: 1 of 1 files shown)


mbeacom:
@AquiGorka Can you shed some light as to why this is beneficial/necessary or provide a use case example?

Thanks for the contribution!

AquiGorka:
@mbeacom I project I worked on was using `moviepy` to generate movies, we included elements in the clips that moved around, appeared/disappeared, rotated, scaled, etc.
Linear animations were ok, but having the possibility to add different `tweens` made them smoother and to do so we included this library.
Since we were using `docker` for dev and building the whole thing we integrated our own custom image but I figured others might benefit from having a `tweening` library with `moviepy`.

PR #875, opened 2018-11-26
modified max duration error for better understanding

kapilkd13:
When embedding video larger than the maxduration set, the error thrown suggests to set `max_duration` param but it is rather `maxduration` without underscore. Also added a line to suggest in  which function to make changes. It is minor, but it had me scratching my head for a minute.  


kapilkd13:
@Zulko Can you look at this. Its very trivial PR, doesn't affect the core codebase in any way.

Zulko:
Sure. Sorry for being slow.

PR #878, opened 2018-12-04
Add progress callback feature

MoZhonglin:
When use MoviePy as a part of service, it's necessary to know current process state. In this PR, a new parameter called ```progress_callback``` is added, which is a callback function used to pass messages.
For example, if we are building an online video converter service, we need to show percents of processed frames. Suppose we already have a function called ```update_state(percent)``` which can push state messages to frontend updating progress bar on a web page, simply add a callback function and pass it to MoviePy:
```python
def func_to_callback(current, n_frames):
  update_state(current / n_frames)

final_clip.write_videofile('result.mp4', fps=25, progress_callback=func_to_callback)
```

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -141,7 +141,8 @@ class VideoClip(Clip):
                         rewrite_audio=True, remove_temp=True,
                         write_logfile=False, verbose=True,
                         threads=None, ffmpeg_params=None,
-                        progress_bar=True):
+                        progress_bar=True,
+                        progress_callback=None):
         """Write the clip to a videofile.
 
         Parameters
@@ -244,6 +245,10 @@ class VideoClip(Clip):
         progress_bar
           Boolean indicating whether to show the progress bar.
 
+        progress_callback
+          After writing each frame, the callback function will be invoked.
+          The function should have a signature ```func(serial_of_current_frame, total_number_of_frames)'''
+
         Examples
         ========
 
@@ -324,7 +329,8 @@ class VideoClip(Clip):
                            audiofile=audiofile,
                            verbose=verbose, threads=threads,
                            ffmpeg_params=ffmpeg_params,
-                           progress_bar=progress_bar)
+                           progress_bar=progress_bar,
+                           progress_callback=progress_callback)
 
         if remove_temp and make_audio:
             os.remove(audiofile)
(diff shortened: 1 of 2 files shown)


Zulko:
Thanks for the PR, but there has been some discussion around this in previous PR, and while I am slow at it my intention was to use [proglog](https://github.com/Edinburgh-Genome-Foundry/Proglog) (which I have also written) as a progress log manager. Proglog makes it easy to display progress bars in different contexts (console, notebook, website) and allows custom data collection and arbitrary callbacks.

MaZZly:
Progress callback would be awesome for a web service I'm currently building. any ETA when something like this would be done? @Zulko

Zulko:
I have pushed 0.2.4.0 with a whole new system for logs and progress bars based on Proglog. See the corresponding section of the README.

I don't know what you backend job manager is, but for my projects I use Redis Queue and it works like a charm with Proglog to get infos to the frontend.

See [these lines](https://github.com/Edinburgh-Genome-Foundry/Proglog/blob/f9faa8e7af7d872a4471f90934dfe347b5d87555/proglog/proglog.py#L352-L371) for the Proglog Logger class linked to Redis Queue (sorry it's not much commented), hopefully it can give you inspiration on how to make a Logger class for your own needs.

MaZZly:
@Zulko I've been able to get the progress bar logging working fine by saving the prgoress state to redis cache :+1: 

One problem I noticed with 0.2.4.0 (that doesn't happen on 0.2.3.5):
 When using `clip.audio.write_audiofile('out.mp3')` the volume is very low, and sound is garbled..

I'm not sure if this is due to me installing it incorrectly through `pip git+repourl` or if some bug was introduced with 0.2.4.0?

Could you try to extract the audio from e.g. https://sample-videos.com/ and see if the resulting mp3 plays correct?

Zulko:
Thanks for testing, and "how dear" for the sound problem, sounds like an introduced bug - that's the kind of things that the test suite doesnt catch. Can you please open an issue to remind me to investigate (should do it shortly).

MaZZly:
https://github.com/Zulko/moviepy/issues/891
There you go =)

Overdrivr:
@Zulko I'm triaging PR and stumbling on this one. I'm not clear, is it possible with proglog to pass in a callback function (or similar) in order to be notified when operation makes progress ? Is this just a matter of documenting this behavior ?

tburrows13:
@Overdrivr yes, this feature has definitely now been implemented with the Proglog. The discussion on the lack of documentation can be found at #1020.

belthaZornv:
Thanks for this, I've forked and put this in, prefer the easy callback solution than using ProgLog - unless there's a good example of using it, which doesn't seem there is.

PR #881, opened 2018-12-08
Bugfix for when srt file ends in white space

ZainRizvi:
If the srt file ends in whitespace the SubtitlesClip class's constructor crashes on line 54 since the None type was inserted where the iterator on line 54 expects the none to be an iterable.


ZainRizvi:
Edit to the bug description: By "white space" I meant "empty new-lines"

The default .srt file downloaded from YouTube-generated captions cause this failure to occur.  It can also be reproduced by taking any .srt file and adding two new-lines at the end of it

PR #579, opened 2017-06-03
Correcting errors  and minor modifications

gpantelis:
With this parameters "-vcodec", "copy", "-acodec", "copy" the output video was wrong(starting time was wrong and the video freezed after a few seconds)
But when i remove those the output video was exaclty like I want to.

Issue #508, opened 2017-03-24
ffmpeg_extract_subclip returns black frames


ksharsha:
I am trying to extract particular sub clips from videos and I see that when the duration of the sub clip is shorter (around 2- 3 seconds), only black frames are seen in the subclip. I guess this might be because ffmpeg is unable to locate the key frames. How do I deal with this?

Zulko:
Hey - That's not enough informations and I've never seen this before, can you provide some code to reproduce the issue ? (and possibly a videofile for which the issue exists too)

ksharsha:
Hi, I recreated the issue using this gist https://gist.github.com/ksharsha/f28ca2f019070389c38741fb90a077d1 . I am attaching the video file as well. You can see that the audio is still preserved however, the video is completely blank. This is the video that I am trying to extract the subclip from (tstart = 12.334277199999999, tend = 15.97166675)

earney:
It seems like this might help..
[https://trac.ffmpeg.org/wiki/Seeking](https://trac.ffmpeg.org/wiki/Seeking)

when I try what is recommended by this link, I do get video, but the video seems to be a bit messed up (like a bad key frame).  I'm not very good with ffmpeg, so maybe someone else sees a solution?

ksharsha:
@earney true, the problem is with this cmd = [get_setting("FFMPEG_BINARY"),"-y",
      "-i", filename,
      "-ss", "%0.2f"%t1,
      "-t", "%0.2f"%(t2-t1),
      "-vcodec", "copy", "-acodec", "copy", targetname]
I am not good with ffmpeg either but looks like the above cmd has to be changed.

D-Chief:
Is this issue resolved? Is there any other way instead of using ffmpeg_extract_subclip to clip videos?

landogardner:
Looks like PR #579 could fix this.

gpotter2:
Hi !

I made https://github.com/Zulko/moviepy/issues/882 which is quite clear and well-documented. It seems if fell unoticed, but it does fix this issue, in a better way than #579

Regards

ebenz99:
Updating to say I believe this may still be a problem. Using ffmpeg_extract_subclip for 2-3 second clips and continue to run into this issue.

keikoro:
@ebenz99 Should we reopen this issue?

ebenz99:
Up to you. I ended up just using the workaround I mentioned (removing some options and taking a quality hit), but it might be nice to not have to do that.

[...]

PR #882, opened 2018-12-09
Update ffmpeg_tools.py

gpotter2:
(Other bullets do not seem relèvent to this PR)

This PR follows what was tried in 
https://github.com/Zulko/moviepy/pull/579. If prevents black frames from appearing when cutting an extract out of the key frames.

6e78421b-086a-4db5-94b9-5389977f4040

The fix (I tested it multiple times with various parameters) is directly inspired from  https://stackoverflow.com/a/33188399/5459467 which quotes the above doc.

I am not sure unit tests will easily allow to detect such an issue :/

Thanks a lot for the project !

fixes https://github.com/Zulko/moviepy/pull/579
fixes https://github.com/Zulko/moviepy/issues/508


gpotter2:
Duplicated of https://github.com/Zulko/moviepy/pull/848

PR #895, opened 2018-12-28
add __getitem__ method to Clip

PaperBag42:

Added a `__getitem__` method to the base `Clip` class as an alias to existing methods.
`clip[t]` is an alias to `clip.get_frame(t)`.
`clip[t_start:t_end]` is an alias to `clip.subclip(t_start, t_end)`.

This won't work with the `SubtitlesClip` subclass, since it already has a `__getitem__` method. Should work fine with the rest.


Overdrivr:
Thanks for your contribution. Could you add a test case demonstrating this works as expected ?

mgaitan:
Oh my god, I've just done the same #1076 !

PaperBag42:
LOL, I actually forgot I ever made this PR. I'll close it, yours looks better.

Issue #207, opened 2015-09-14
ffmpeg_parse_infos failed get duration with Quicktime XMP metadata


maty974:
Hi

I was trying to create a VideoFileClip() with a custom Quicktime .mov format, and
it seems that MoviePy don't like some metadata embedded in the file :P

`
:       xmpDM:duration="0:32.303"
`

```
                    :       <xmpDM:duration
                    :        xmpDM:value="130667"
                    :        xmpDM:scale="1/1000"/>
                    :       <xmpMM:History>
```

`
xmpDM:duration="10584000000"
`

Maybe its causing a bug here ?

https://github.com/Zulko/moviepy/blob/master/moviepy/video/io/ffmpeg_reader.py#L261-L270

maty974:
Here is a full output of ffmpeg infos of a file that contain xmp metadata

```
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '/home/110_PREV_FINAL.mov':
  Metadata:
    major_brand     : qt  
    minor_version   : 537199360
    compatible_brands: qt  
    creation_time   : 2015-09-14 14:57:32
    xmp             : <?xpacket begin="﻿" id="W5M0MpCehiHzreSzNTczkc9d"?>
                    : <x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk="Adobe XMP Core 5.0-c060 61.136967, 2010/06/15-10:43:27        ">
                    :  <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
                    :   <rdf:Description rdf:about=""
                    :     xmlns:xmpMM="http://ns.adobe.com/xap/1.0/mm/"
                    :     xmlns:xmpDM="http://ns.adobe.com/xmp/1.0/DynamicMedia/"
                    :     xmlns:stDim="http://ns.adobe.com/xap/1.0/sType/Dimensions#"
                    :     xmlns:dc="http://purl.org/dc/elements/1.1/"
                    :    xmpMM:DocumentID="xmp.did:39FA818BE85AE511B9009F953BF804AA"
                    :    xmpMM:InstanceID="xmp.iid:39FA818BE85AE511B9009F953BF804AA"
                    :    xmpDM:videoFrameRate="24.000000"
                    :    xmpDM:videoFieldOrder="Progressive"
                    :    xmpDM:videoPixelAspectRatio="1/1"
                    :    xmpDM:audioSampleRate="44100"
[...]

Issue #140, opened 2015-03-06
utf8_decode() error on some mov files with quotes in metadatas


Fixazo:
Hello,
I had a problem using moviepy with some .mov files which included bad metadatas, but I had to treat those files. It told me it couldn't decode 0x92 byte or something
I tracked down the error to ffmpeg_reader.py at line : 
infos = proc.stderr.read().decode('utf8')

I replaced it with that, and it solved my problem : 
infos = proc.stderr.read()

Maybe it can help !
Bye

Zulko:
thanks, I think the "decode" is here because of python version compatibility. What's your OS and python version ?

Fixazo:
Thanks for your answer,
I'm on windows 7 x64 and running on python 2.7.2 (I should update it maybe !)
By the way, it is a really useful module!!

Fixazo:
I'm on 2.7.8 sorry!

SandeR2012:
@Zulko 
Have same issue. Python 3.4.1, moviepy==0.2.2.11
Changed line 241 in moviepy\video\io\ffmpeg_reader.py to
`infos = proc.stderr.read().decode('utf8', errors='replace')`
Video for test: https://www.dropbox.com/s/hfuz7wlhx0ziu33/dark.m4v?dl=0

TheShadow29:
I am facing the same issue on python3.7 moviepy==0.2.3.5

This is happening for some specific videos and the fix mentioned by @SandeR2012 worked for me. I can send in a pull request if needed. 

Here is a video for test: https://drive.google.com/open?id=1Ft0zEpG227j_wMiqueL_l5zEW42WRU41

keikoro:
I'm wondering if this is even still an issue considering the linked PR got closed because it's no longer relevant (due to changes in the code).

keikoro:
Closing this for now. Feel free to reopen or, ideally, open a new issue using the issue template (and link this one), in case the problem persists.

PR #901, opened 2019-01-15
Small fix to resolve unicode error

TheShadow29:
Its a small fix for unicode errors. Solves #140 and maybe #207. To the best of my understanding (https://docs.python.org/3/howto/unicode.html) this should not cause any issues. However, I am not sure what a good test would be. 

### DIFF
--- a/moviepy/video/io/ffmpeg_reader.py
+++ b/moviepy/video/io/ffmpeg_reader.py
@@ -259,7 +259,7 @@ def ffmpeg_parse_infos(filename, print_infos=False, check_duration=True,
 
     proc.stdout.readline()
     proc.terminate()
-    infos = proc.stderr.read().decode('utf8')
+    infos = proc.stderr.read().decode('utf8', errors='replace')
     del proc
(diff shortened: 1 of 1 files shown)


Overdrivr:
Hi @TheShadow29, thanks for your contribution. Could you please:

- rebase your branch against master as there as been some changes on this piece of code
- write a new test case for ffmpeg_parse_infos, by simply parsing the faulty video you've mentioned [here](https://github.com/Zulko/moviepy/issues/140)?

Thanks

TheShadow29:
Sure. I will try to do it sometime on the weekend

daniestrella1:
Is this fix added on release 1.0.0?

tburrows13:
Hi, it looks like this section of code was changed in https://github.com/Zulko/moviepy/commit/350571acfe620bf1479fba555d55efb28d51e568 so this PR can be closed as it is no longer relevant. The new fix is in #959 so feel free to try that change out and report back on it.

Issue #908, opened 2019-02-13
Imageio's new use of imageio-ffmpeg


almarklein:
*edit: hit the wrong key and posted this before it was done.* 

Hi there,

I meant to post an issue for a heads-up about the new situation, but I guess I forgot. So here it is. I realize that the change means some "upstream pain", but it was needed to keep imageio managable.

### In short

Imageio previously had you download the `ffmpeg` executables using a function or a CLI command. This behavior has changed. Instead, using imageio's ffmpeg plugin now requires the new `imagio-ffmpeg` library. This library contains a low-level ffmpeg wrapper, and ships the ffmpeg binaries via platform specific wheels.

### Background
(You can skip over this)
The reason for the change is that the strategy with the explicit download was clumsy, and more importantly a security hazard. Combined with the fact that about 50% of imageio's incoming issues were related to ffmpeg, I decided to spin the ffmpeg bit out. We also considered other ffmpeg wrappers, but distribution remains hard, so I decided on a solution that ships the exes in the platform specific wheels. The (low level) wrapper should also be more stable (e.g. wrt terminating ffmpeg subprocesses), and can be improved upon by people who care about ffmpeg more easily.

#### Dependencies

The eventual solution should be to add a dependency on `imageio-ffmpeg`, but I could not put myself to write py27 compatible code in 2019, so as noted above, `imageio-ffmpeg` is py3k only. I guess that Moviepy will drop 2.7 support somewhere this year? Ah, I just saw that you pinned imageio to `<2.5.0` for older Python versions. That should do it.

> ... deprecate their old versions, meaning that they may point to deprecated URLs to download ffmpeg

The binaries will stay up for the old versions of imageio. At some point I removed some older ones (matching a 1.5 year old imageio version), but these are restored.

#### The platform specific wheels do not work on all platforms

The platform-specific wheels are, well, platform-specific, so some platforms will just install the tarfile (some Linuxes). Also, some platforms may install the wheel, but the ffmpeg exe may not work (e.g. Windows XP).
Also see [the corresponding issue](https://github.com/imageio/imageio-ffmpeg/issues/12). In short, I suspect that the platform does not match `manulinux2010_x86_64`, which is a consequence of the platform that the (new) ffmpeg exes were build on. The solution is to document that Linux users (that don't use a "modern desktop Linux") may want to use:
```
$ pip install imageio-ffmpeg --no-binary
$ apt-get install ffmpeg  # or equivalent
```

#### Other needed changes?

[...]

PR #916, opened 2019-02-28
[docker] drop the not needed download and symlink of ffmpeg

das7pad:
`libav-tools` has `ffmpeg` as a direct dependency.

https://github.com/Zulko/moviepy/blob/2ad7285bec9dee1a477250891ddc9c418a319dc1/Dockerfile#L4

Ref: https://github.com/Zulko/moviepy/issues/908#issuecomment-468182584


tburrows13:
Agreed, I think that the current docker file is broken and this would fix it.

Issue #876, opened 2018-11-27
CompositeVideoClip has no audio


marceloandriolli:
### Expected Behavior
The video file generated at end of the process with orignal audio.

### Actual Behavior
The video file generated at end of the process has no audio.

### Steps to Reproduce the Problem
```
def add_watermark_logo(video_filename):
    """
    Adds the logo as watermark to the video, given its filename.
    :param video_url: String containing the video filename.
    :return:
    """

    video = mp.VideoFileClip(video_filename)
    logo = (mp.ImageClip("logo.png")
              .set_duration(video.duration)
              .resize(height=50)
              .margin(right=10, bottom=10, opacity=0)
              .set_pos(("right", "bottom")))

    final = mp.CompositeVideoClip([video, logo])
    final.write_videofile("{}".format(video_filename), progress_bar=False)

```

### Specifications

  - Python Version: 3.6 
  - Moviepy Version: 0.2.3.5
  - Platform Name: Docker on MacOs 10.13.6
  - Platform Version: 18.06.1-ce-mac73 (26764)

Zulko:
- It seems that your script is writing over the video it is reading. You shouldn't do that, I'm not even sure how it can work. use ``final.write_videofile("MODIFIED_" + video_filename)``. It could be a source of the problem, not sure.

- Have you tried reading the video with different readers, like VLC ? One possibility is that there is sound, but the audio codec used by default is incompatible with your reader.

- Regarding ``-an``, I won't have time to look into it soon, but that would be worrying. @Julian-O you added this line, do you know if ``-an`` could cause no audio ?

Julian-O:
@Zulko: I was the last to edit, but I didn't add this line. (I did review my changes to double-check this.)

The last-edit of that part of the code was [four years ago](https://github.com/Zulko/moviepy/commit/70892d956f6e72e37d1c0e0a0f02f5b133c8734a), but the "-an" option has existed since at least 2013, when `ffmpeg_writer.py` was spun-off from `writers.py`.

Julian-O:
I haven't tested, but, it sounds completely plausible that passing `-an` would cause no audio. 
I wouldn't have expected a codec change.
Testing with a different filename sounds like the first step though.

marceloandriolli:
Thank you @Zulko! 

I did a test reading the output video into VLC and audio works.  Do you know what is an audio codec works to QuickTime?

Sv3n:
[...]

PR #919, opened 2019-03-01
Fix no audio in output videos

Baronsindo:
i removed '-an'
because of it the outputted videos has no sound


Overdrivr:
Thanks for your contribution, does this PR refers to #876 ?
Could you write a test case demonstration the lack of audio in the exported file ?

ThePhonon:
Note: the -an option is arguably needed.  The bug is that it is provided before the '-i' option.
See https://github.com/Zulko/moviepy/pull/968

Baronsindo:
Thanks for clearing that out.

tburrows13:
The general census seems to be that #968 should be merged, not this one.

PR #927, opened 2019-03-19
Fix for #926

### LINKED ISSUE #926: ffmpeg_parse_infos silently hangs on Windows when MP4 file contains enough metadata
Loading any mp4 file with enough chapters in metadata causes a silent hang without error.

After investigating a little bit, I have found that the silent hanging is caused by the process opened to call FFMPEG when retrieving file metadata. If the metadata is too large (caused by having many chapters in the video for instance), FFMEPG will print all this metadata to STDOUT/STDERR and my hypothesis is that somehow the pipe's buffer gets filled up entirely causing the hanging. _(Not sure about this explanation though, if someone has more experience with Processes/Pipes feedback would be welcomed.)_ 

### Steps to Reproduce the Problem

```
from moviepy.editor import *
audio = AudioFileClip('new_sintel.mp4')
```

To reproduce, please find attached a zip containing a trailer of the [open Sintel movie](https://download.blender.org/durian/trailer/), where I have [manually added chapters using FFMPEG](https://medium.com/@dathanbennett/adding-chapters-to-an-mp4-file-using-ffmpeg-5e43df269687). With ~10 chapters, moviepy loads metadata successfully, but with ~15 chapters it hangs. The ZIP contains the version where it hangs:

[new_sintel.zip](https://github.com/Zulko/moviepy/files/2982581/new_sintel.zip)

Just in case, this is the dummy metadata I have placed in the file:

[chapters.txt](https://github.com/Zulko/moviepy/files/2982602/chapters.txt)

I can provide a fix, I hope it can get released pretty quickly because I need to fix this for work.

### Specifications

  - Python Version: 3.6
  - Moviepy Version: 1.0.0 (ffmpeg: 4.1)
  - Platform: Win10

Overdrivr:
Closes #926 

### DIFF
--- a/moviepy/video/io/ffmpeg_reader.py
+++ b/moviepy/video/io/ffmpeg_reader.py
@@ -256,10 +256,9 @@ def ffmpeg_parse_infos(filename, print_infos=False, check_duration=True,
         popen_params["creationflags"] = 0x08000000
 
     proc = sp.Popen(cmd, **popen_params)
-
-    proc.stdout.readline()
-    proc.terminate()
-    infos = proc.stderr.read().decode('utf8')
+    (output, error) = proc.communicate()
+    infos = error.decode('utf8')
+    
     del proc
 
     if print_infos:
--- a/tests/test_ffmpeg_reader.py
+++ b/tests/test_ffmpeg_reader.py
@@ -31,6 +31,8 @@ def test_ffmpeg_parse_infos():
     assert d['audio_found']
     assert d['audio_fps'] == 48000
 
+    d=ffmpeg_parse_infos("tests/resource/sintel_with_15_chapters.mp4")
+    assert d['audio_found']
 
 if __name__ == '__main__':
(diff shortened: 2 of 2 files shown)


bertwagner:
Hi @Overdrivr ,
I also have this same problem and I was able to find a fix for my test case:  https://github.com/bertwagner/moviepy/commit/d5fc73a80174c0b2caefd6ce227b65674b1ff2ee#diff-a49583100a493b6fe1c05b48824e4a3c

I ran it on your demo file and it seemed to work correctly.  Do you want to test it out for yourself?

If it works, feel free to add it to your fix and merge it in since you already have started this thread.  Otherwise let me know and I can submit a PR, thanks!

Overdrivr:
@bertwagner I came up to the same exact solution, nice :) I wanted to see the CI fail on Windows platform without the fix, but since you confirmed the issue that's perfect I'll push the fix. Thanks !

PR #958, opened 2019-05-14
Test that the temporary audio file exists

ybenitezf:
I'm using Moviepy in one of my projects and this problem about a temporary file that does not exist when calling `VideoClip.write_videofile` is popping up in our CI jobs.

This fix simply does the same thing as in `TextClip` testing that the tmp file exist before calling `os.remove`

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -240,7 +240,7 @@ class VideoClip(Clip):
 
         logger
           Either "bar" for progress bar or None or any Proglog logger.
-        
+
         verbose (deprecated, kept for compatibility)
           Formerly used for toggling messages on/off. Use logger=None now.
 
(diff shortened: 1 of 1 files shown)


moviepy/video/VideoClip.py
```diff
@@ -326,7 +326,8 @@ def write_videofile(self, filename, fps=None, codec=None,
                            logger=logger)
 
         if remove_temp and make_audio:
-            os.remove(audiofile)
+            if os.path.exists(audiofile):
+                os.remove(audiofile)
```
mbeacom: First off, thanks for the contribution!

Generally, this looks good to me.  Should we move the `os.path.exists()` to [`L328`](https://github.com/Zulko/moviepy/pull/958/files#diff-4b7e3f0e11cb6f470780d69172d8fbe1L328) though?  I suppose it's trivial, regardless.

ybenitezf: I think is not the same, in L328 u are asking if `remove_temp` parameter is given and is not the same as to ask if the temp file exists

tburrows13: Either way, it makes no difference. Thanks for contributing it though!

Issue #934, opened 2019-04-01
Issue with ffmpeg version


AWAS666:
Problem:
After installing ffmpegversion 20190331-27c94c5 moviepy couldn't export any audio

Solution:
Installed ffmpeg 4.1.1 Release Build and that got it working again

[ffmpeg source](https://ffmpeg.zeranoe.com/builds/)

just a heads up for you guys even though it is more ffmpeg related
have a nice day :)

tburrows13:
Should be fixed in #968.

PR #968, opened 2019-06-04
Reorder FFMPEG_VideoWriter command arguments

ThePhonon:
stream options must be provided before the '-i' option to affect that stream.
The existing code caused '-an' (no audio) to be applied to the next stream.  This causes no audio to be generated if the next stream is the audio stream for the clip.

e.g.  On Windows:

```
import os
os.environ["IMAGEIO_FFMPEG_EXE"] = "path\to\newest\ffmpeg.exe"
video = VideoFileClip("my_file_with_audio.mp4").subclip(0, 10)
video.write_videofile("created_clip.mp4")
```

will generate a video file with no audio.


tburrows13:
Closes #876 #934 and possible #1032.

Issue #113, opened 2015-01-16
ImageMagick write gif success but no file found 


lisongx:
`imageio`  throw segmentation fault

seansay:
also, xxxx.ficache seems to be in the same folder I write gif to

seansay:
wrote to video file works fine

Zulko:
Thanks for the report. Can you provide the following ?

- Your OS and python version
- The code that you are using (or at least a bit of it),
- The error that you get (or do you really have no message at all apart from the seg fault ?)
- Your moviepy version (I imagine it's the last one).

I am calling @almarklein, the imageio author, into the discussion, it may concern him.

Cheers !

almarklein:
The `xxxx.ficache` is an artifact of freeimage; it may use a temporary file to preserve RAM. I suppose you see it because it segfaults before FreeImage had a chance to clean it up. Other than that, there's not much to tell without more information.

almarklein:
Oh wait. It might actually be fixed in current master. Someone fixed a similar bug recently: https://github.com/imageio/imageio/pull/53

Zulko:
@seansay Any improvement ? If not, can you give a few more infos ?

Ziewvater:
I'm having this same issue as well. Trying to use `write_gif`with default of `program='imageio'` leads to a segmentation fault with an output that looks like this each time:
```
MoviePy] Building file 54.gif with imageio
|----------| 0/77   0% [elapsed: 00:00 left: ?, ? iters/sec][1]    95953 segmentation fault  python scriptname.py
```

Using ImageMagick avoids segmentation fault, but results in a success message, but no gif file to be found on my machine. Code and printout looks like this:
```
from moviepy.editor import *
video = VideoFileClip("Movie File Name.mp4").resize(0.3).subclip(3,5)
video.write_gif("test.gif", program="ImageMagick")

[MoviePy] >>>> Building file test.gif
[MoviePy] Generating GIF frames...
[MoviePy] Optimizing the GIF with ImageMagick...
[MoviePy] >>>> File test.gif is ready !>>>
```

Looks like I'm using moviepy 0.2.2.5. Also, if it helps, I was playing around to see what I was doing wrong and revisited an older project using an older version of moviepy (0.2.1.8.02), and the segmentation fault issues did not occur when using `to_gif`. In the more recent version, however, the segmentation fault occurred when using both `to_gif` and `write_gif`. I'm using Python 2.7.9, and on OSX 10.10

[...]

PR #972, opened 2019-06-15
Several ImageMagick related bug fixes

KiLLAAA:
- Fix for Zulko/moviepy#113

- Fix for another bug when setting fuzz had no influence on output file due to wrong order of params passsed to ImageMagick


tburrows13:
This looks good to me. Let us know if there are any issues.

Issue #689, opened 2017-12-27
Refer to magick on https://zulko.github.io/moviepy/install.html


mgrazebrook:
A cousin of #378 

This documentation is out of date and should refer to magick not convert:

https://zulko.github.io/moviepy/install.html

The current text
> Once you have installed it, ImageMagick will be automatically detected by MoviePy, except on Windows !. Windows user, before installing MoviePy by hand, go into the moviepy/config_defaults.py file and provide the path to the ImageMagick binary called convert. It should look like this
> 
> IMAGEMAGICK_BINARY = "C:\\Program Files\\ImageMagick_VERSION\\convert.exe"
> 

The github version of the docs has the same bug:

> Windows users, before installing MoviePy by hand, need to edit moviepy/config_defaults.py to provide the path to the ImageMagick binary, which is called convert. It should look like this

tburrows13:
Fixed in #980.

Issue #378, opened 2017-01-03
Windows: specifying path to ImageMagick in config_defaults.py


WDBell:
```
IMAGEMAGICK_BINARY
    For linux users, 'convert' should be fine.
    For Windows users, you must specify the path to the ImageMagick
    'convert' binary. For instance:

    IMAGEMAGICK_BINARY = r"C:\Program Files\ImageMagick-6.8.8-Q16\convert.exe"

"""

FFMPEG_BINARY = 'ffmpeg-imageio'
#~ IMAGEMAGICK_BINARY = 'auto-detect'
IMAGEMAGICK_BINARY = "C:\\Program Files\\ImageMagick-7.0.4-Q16\\magick.exe"
```

The final few lines of config_defaults.py indicate how the location of ImageMagick should be specified for use by moviepy. However, often when Image ImageMagick is installed these days (apparently with versions numbered 7x) convert.exe is **not** included. Fortunately, ImageMagick can be specified as shown here.

Suggest change to config_defaults.py.

sephib:
Found the same solution - Please update the documentation - would have saved me 1 Hour+

saifrahmed:
I get the error ".This error can be due to the fact that ImageMagick is not installed on your computer, or (for Windows users) that you didn't specify the path to the ImageMagick binary in file conf.py, or.that the path you specified is incorrect"

-- do we edit config_defaults.py (which comes with the distro) or conf.py -- or do we rename config_defaults.py to conf.py.  Does the file need to be in the same directory as the rest of the scripts?

WDBell:
@saifrahmed: Are you on Windows?

WDBell:
I'm on Windows 10. These are the last few lines of the file config_defaults.py on my computer. You may notice that ImageMagick, including its binaries, is stored in a location away from other scripts. config_defaults.py is stored in the folder C:\Python34\Lib\site-packages\moviepy.

```
IMAGEMAGICK_BINARY
    For linux users, 'convert' should be fine.
    For Windows users, you must specify the path to the ImageMagick
    'convert' binary. For instance:

    IMAGEMAGICK_BINARY = r"C:\Program Files\ImageMagick-6.8.8-Q16\convert.exe"

"""

FFMPEG_BINARY = 'ffmpeg-imageio'
#~ IMAGEMAGICK_BINARY = 'auto-detect'
[...]

PR #980, opened 2019-07-07
Updating Docs for ImageMagick Installing Guide

ABODFTW:
Based on #689 and #378 I have created this Pull request where I updated the installing docs and the comments on the code base on config_defaults.py, So it's less confusing for windows users when using ImageMagick with MoviePy.


tburrows13:
I've added some more information from the readme into the docs, so this can can be merged. Thanks for your contribution!

mgrazebrook:
Thanks. Looks good.

PR #987, opened 2019-07-16
Specify Coverage version explicitly.

Julian-O:
The Travis builds have been broken for about 4 weeks.

I didn't fully explore the reasons, but Coverage 5.0a5 (an alpha version) was being installed, and then clashing with the Coverall which required version <5.0.

As a result, the Travis tests were not being run.

Adding a specific requirement for coverage solved the problem.

----

Also removed a deprecated parameter from pytest.raises(), which was leading to lots of deprecation warnings in the tests. [Sorry - should have been two branches, but they got mised together. The changes are trivial.]

- [N/A] If this is a bugfix, I have provided code that clearly demonstrates the problem and that works when used with this PR
- [N/A ] I have added a test to the test suite, if necessary
- [N/A] I have properly documented new or changed features in the documention, or the docstrings
- [N/A] I have properly documented unusual changes to the code in the comments around it
- [N/A] I have made note of any breaking/backwards incompatible changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,6 +89,7 @@ doc_reqs = [
     ]
 
 test_reqs = [
+        'coverage<5.0',
         'coveralls>=1.1,<2.0',
         'pytest-cov>=2.5.1,<3.0',
         'pytest>=3.0.0,<4.0',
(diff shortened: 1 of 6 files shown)


tburrows13:
Thank you very much for tracking this down. It seems to do the job!

PR #1008, opened 2019-09-09
Get ffmpeg to use all audio streams

vmaliaev:
For some formats ffmpeg picks just one audio stream without "-map 0" instead of all.


tburrows13:
It would be nice to have an example of a file that this can be demonstrated on, but I'll merge it anyway as it looks to be correct.

PR #1016, opened 2019-09-30
Fixed problems with proc package

Kaszanas:
Unit testing yielded couple of errors during the processing of the video.
All of them were connected to the use of proc package.


tburrows13:
Sorry, I didn't mean to get rid of the 'draft' status.

Julian-O:
I am confused by the purpose of the check before terminating:

    if self.proc.poll() is not None:
        self.proc.terminate()

Firstly, is there any harm in skipping the check and just calling terminate every time?

Secondly, the return code definition says:

     A None value indicates that the process hasn’t terminated yet.

So, this seems to call terminate only if it *has* already been terminated.

I am not an expert at `subprocess`, so this could just be my confusion rather than a bug.

tburrows13:
Adding `.wait()` was added as part of another PR some time ago so is no longer shown here in the 'Files changed'. Checking `.poll()` has come up again in #1296.

mondeja:
Since @Kaszanas has not posted the errors, #1296 has been merged and the source code has changed a lot since this pull, I'm closing this. Feel free to reopen it if your errors persist.

PR #1018, opened 2019-10-01
Add Travis support for Python 3.7 and 3.8

tburrows13:


mbeacom:
@tburrows13 `3.7` on Travis requires `dist: xenial`

mbeacom (dismissed):
This will require updating the travis dist from `trusty` to `xenial`

tburrows13:
Cheers, that does agree with my research. I don't know if that will break the other checks though. There's definitely ways around it anyway.

tburrows13:
It seems that it can't find an ffmpeg version for Xenial

tburrows13:
I think that this is ready to merge now (after squashing first!). There's a fairly large special case for Python 2 (which is basically the previous contents of .travis.yml) that can be removed when we stop testing against Python 2.

I've been trying a few different things with pytest, tox and Travis (#1083, #1085 and https://github.com/tburrows13/moviepy/tree/travis-matrix), none of which have really worked. In summary, getting a unified test process between local and CI testing is complicated and not worth it. Travis does now have support for Windows, but it is not yet possible to run ffmpeg on it yet because the server version is missing multimedia dlls.

Some further things to explore would be testing macOS and pypy on Travis. From my recent endeavours, it seems like they will both be possible. The hardest parts are usually getting ffmpeg and ImageMagick to work.

tburrows13:
I'd appreciate a quick review of this before merging.

PR #1038, opened 2019-11-17
Add video option to AudioClip.write_audiofile()

thenewguy:
This exposes FFMPEG_AudioWriter's input_video kwarg which now allows AudioClip.write_audiofile() to specify the path to a video file that will be copied to the ffmpeg file output.

It is analogous to VideoClip.write_videofile() with the audio kwarg set as string.  Clip integration is not yet written but should be easily extendable if needed.  I currently do not need it so I did not implement.

PR #1037, opened 2019-11-17
test possible audio input to write_videofile

thenewguy:
Provides tests for #876 and #1032

Intends to supersede #919 and #968

PR #1036, opened 2019-11-16
Log ffmpeg command

thenewguy:
Fixes #270

PR #1039, opened 2019-11-18
Combine PR for #1038, #1037, #1036

thenewguy:
Ref #1038, #1037, #1036 

I think this is ready for commit.  Works for me and tests are provided


tburrows13:
Thank you very much for creating this, along with the documentation and tests.

I'm afraid that I don't quite understand the point. Why do you need to pass videos to AudioFileClip? Is there any difference between that and creating a VideoFileClip with audio set to your AudioFileClip? Is it just a convenience function then?

thenewguy:
It is a convenience function but also different.

I was working with audio files and had a video finalized file with no audio track.  I wanted to merge them without processing the video in any way.  I noticed that the AudioFileWriter already had this exposed and it just needed to be passed to the AudioClip.

This path felt cleaner.  If you don't care for it I understand

tburrows13:
@thenewguy sorry, but it is not very helpful to have 2/3 features in one PR. I'd happily merge #1036, but the other changes would require more consideration/testing. Are you happy with #1036 as it is now, or did you make further changes when including it in this one?

moviepy/audio/AudioClip.py
```diff
@@ -185,6 +185,19 @@ def write_audiofile(self, filename, fps=None, nbytes=2, buffersize=2000,
...
+        
+        video (default None)
+          A video file path passed as input to ffmpeg to be copied into
+          the final output.
+          
+          ** NOTE: video options are intentionally limited. AudioClip is meant for
+                   working with audio. If you need advanced video functionality you
+                   should be using VideoClip or one of it's subclasses.
+        
+        video_codec (default 'copy')
```
mgaitan: default "copy" but the actual default in the signature is `None`. It's a bit confusing. Could you clarify in the docstring?

moviepy/audio/io/ffmpeg_audiowriter.py
```diff
@@ -51,7 +55,7 @@ def __init__(self, filename, fps_input, nbytes=2,
                 '-ar', "%d" % fps_input,
                 '-ac', "%d" % nchannels,
                 '-i', '-']
-               + (['-vn'] if input_video is None else ["-i", input_video, '-vcodec', 'copy'])
+               + (['-vn'] if input_video is None else ["-i", input_video, '-vcodec', video_codec or 'copy'])
```
mgaitan: btw, why `video_codec`'s default is not just `"copy"` ?

thenewguy:
@tburrows13 I didn't see your last comment until today.  I will try to revisit this and see if this diverged from the linked issues in your reply

moviepy/audio/AudioClip.py
thenewguy: It looks like because None means copy per https://github.com/Zulko/moviepy/blob/d20acaaffb91bf46959ee3fc17be4cedb61eda29/moviepy/audio/io/ffmpeg_audiowriter.py#L58 but it could just as easily be 'copy'

keikoro:
@thenewguy Pinging you as the last comment is from June. (:

mondeja:
Since this pull request has a lot of features mixed, I think that it's better to do in separate pulls. Please, reopen your other pulls if you still want this added.

PR #1041, opened 2019-11-27
autodetect ImageMagick on windows

Nex4rius:
Autodetect ImageMagick on windows


moviepy/config.py
```diff
@@ -56,7 +56,11 @@ def try_cmd(cmd):
             IMAGEMAGICK_BINARY = wr.QueryValueEx(key, 'BinPath')[0] + r"\convert.exe"
             key.Close()
         except:
-            IMAGEMAGICK_BINARY = 'unset'
+            try:
+                IMAGEMAGICK_PATH   = sp.check_output('dir /B /O-N "C:\Program Files\ImageMagick-*"'                           , shell=True).decode("utf-8").split("\r\n")[0]
+                IMAGEMAGICK_BINARY = sp.check_output('dir /B /S   "C:\Program Files\{}\*convert.exe"'.format(IMAGEMAGICK_PATH), shell=True).decode("utf-8").split("\r\n")[0]
```
tburrows13: This looks like it may be pretty useful. The only thing I'd comment is that you've put `IMAGEMAGICK_PATH` in all-caps which by convention means that it is a constant used by the program in a variety of places. Here, though, it is just a variable to increase readability so I'd suggest renaming it to `imagemagick_path` or something like that.

Nex4rius: sure it's done

tburrows13:
Would someone else like to approve this PR?

moviepy/config.py
```diff
@@ -56,7 +56,11 @@ def try_cmd(cmd):
             IMAGEMAGICK_BINARY = wr.QueryValueEx(key, 'BinPath')[0] + r"\convert.exe"
             key.Close()
         except:
-            IMAGEMAGICK_BINARY = 'unset'
+            try:
+                imagemagick_path   = sp.check_output('dir /B /O-N "C:\Program Files\ImageMagick-*"'                           , shell=True).decode("utf-8").split("\r\n")[0]
```
mgaitan: 2 concerns. why that long blank after the first argument? 
and what happen on those windows that are not on "C:" and not in english. ? isn't that the "BinPath" requested above, in the line 56, via `wr.QueryValueEx(key, 'BinPath')[0]` ?

moviepy/config.py
```diff
@@ -56,7 +56,11 @@ def try_cmd(cmd):
...
+                IMAGEMAGICK_BINARY = sp.check_output('dir /B /S   "C:\Program Files\{}\*convert.exe"'.format(imagemagick_path), shell=True).decode("utf-8").split("\r\n")[0]
+            except:
```
mgaitan: Also, would be better to catch an explicit exception. A catch-all (whitout even log it) is unpythonic.

moviepy/config.py
Nex4rius: That long blank is so the line below lines up perfectly to make it more readable.
If you are one of the few who decided to install windows not on C then it wont work.
I have window in german and the folder is called "Programme" but it still works with the english folder name "Program Files"  in the path.
Line 56 yes maybe but it doesn't work (even says so in the installation guide).

moviepy/config.py
Nex4rius: I don't see the benefit of an explicit exception here.
If it doesn't work then you end up with the error message that imagemagick couldn't be found and you have to add the path manually (that output is generated somewhere else in moviepy).

Nex4rius:
I have to disagree.

I looked at https://github.com/Zulko/moviepy/pull/1109 and with that it checks for the ImageMagick enviroment variable and then uses that.
I just looked and after installing imagemagick it didn't set the enviroment variable which means I still have to **manually** add it.

I still think it's better if it can autodetect it without configuring / setting enviromental variables / doing anything.

tburrows13:
Sorry, I think you're right, I had possibly misinterpreted #1109. I agree that this does additional processing that #1109 does not do.

You now need to run black over it with `black -t py36 .`. I'd also suggest using `encoding="utf-8"` in `check_output` instead of decoding the output.

Nex4rius:
I added the encoding but what is that `black -t py36` command?

tburrows13:
In cases like this, yes, it is arguably less useful, but overall it ensures that the style is consistent which is a great benefit.

You can install it with `pip install black`, and then run `black -t py36 .` from the root directory of MoviePy. Thanks!

tburrows13:
Hey, thanks for getting it sorted with black! Its a big help to have the code all formatted consistently (I know that yours was fine (just different in matters of taste) before, but a lot of contributions aren't and it is so much easier to run a single tool on a submission than it is to nitpick every little change.) For future reference it should be a one-line command and commit, it looks like you tried to do it manually? Either way, it's fine now.

I made a couple more smaller changes that I just noticed. @mgaitan was right about not using a bare except: https://stackoverflow.com/questions/54948548/what-is-wrong-with-using-a-bare-except. We don't want to catch things like `KeyboardInterrupt` or `SystemExit`.
And your code was giving warnings about invalid escape characters because it used backslashes in the string without marking it as raw with the `r` before it.

If you're happy with my changes, then I'm happy to merge!

Nex4rius:
Sure looks fine now :)

Nex4rius:
What exactly are we waiting for until merge?

tburrows13:
Nothing, just all the maintainers are quite busy :) Commenting reminders every so often like you've done is a good idea.

tburrows13:
Are you using image magick 6 or 7? I notice that this uses `convert.exe` which I think is IM6 only? Is there a reason why it shouldn't be changed to `magick.exe` for IM7?

Nex4rius:
I have imagemagick 7 and it has both convert.exe and magick.exe

PR #1043, opened 2019-11-30
Added encoding optional parameter to subtitle file loading feature

aalkz:

I had many issues with file encodings because I deal with special characters in my SRT files and this was a necessary step for me to solve them for good. It's an easy and straightforward change, but really handy if you need to specify your encoding when loading the SRT file.

Before you could only do this:
```sub = SubtitlesClip("subtitles.srt", generator)```

Now you can also specify your encoding if you want or need to:
```sub = SubtitlesClip("subtitles.srt", generator, encoding='utf-8')```

The `encoding` parameter is optional.

Special thanks to @darkade for helping me figure this out.


tburrows13:
This is a neat feature. It took us a while to get to it, but I'm happy to merge now.

PR #1046, opened 2019-12-08
Improve afx.audio_normalize documentation

dspinellis:
- [X] I have added a test to the test suite, if necessary
- [X] I have properly documented new or changed features in the documention, or the docstrings
- [X] I have properly documented unusual changes to the code in the comments around it
- [X] I have made note of any breaking/backwards incompatible changes

### DIFF
--- a/moviepy/audio/fx/audio_normalize.py
+++ b/moviepy/audio/fx/audio_normalize.py
@@ -6,7 +6,7 @@ def audio_normalize(clip):
     """ Return a clip whose volume is normalized to 0db.
 
     Return an audio (or video) clip whose audio volume is normalized
-    so that the maximum volume is at 0db, the maximum achivable volume.
+    so that the maximum volume is at 0db, the maximum achievable volume.
 
     Examples
     ========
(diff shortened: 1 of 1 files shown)


dspinellis:
Please note that for some reason this effect is not auto-documented [here](https://zulko.github.io/moviepy/ref/audiofx.html).

moviepy/audio/fx/audio_normalize.py
```diff
@@ -3,8 +3,19 @@
...
+    """ Return a clip whose volume is normalized to 0db.
+
+    Return an audio (or video) clip whose audio volume is normalized
+    so that the maximum volume is at 0db, the maximum achivable volume.
+
+    Examples
+    ========
+
+    >>> from moviepy.editor import *
+    >>> videoclip = VideoFileClip('myvideo.mp4').fx(afx.audio_normalize)
```
tburrows13: Because of some 'moviepy.editor' `exec()` hackery, I believe that this can be replaced with
```
    >>> videoclip = VideoFileClip('myvideo.mp4').afx.audio_normalize()
```

dspinellis: It doesn't seem to work that way.  I'm getting the error: `AttributeError: 'function' object has no attribute 'audio_normalize'`. I propose to leave it as is.

tburrows13: Ok, you're correct on this one. I think that the api could be improved to work more like how I thought it did, but that would be in a new PR, perhaps in v2.0.

PR #1057, opened 2020-01-05
Correct the error message to new filename.

jwg4:
The file which needs to be manually edited now seems to be `config_defaults.py`._

- [.] I have added a test to the test suite, if necessary
- [.] I have properly documented new or changed features in the documention, or the docstrings
- [.] I have properly documented unusual changes to the code in the comments around it


tburrows13:
Thanks

Issue #938, opened 2019-04-05
AttributeError: 'NoneType' object has no attribute 'stdout'


Technerder:
### Notice
**If you have arrived here from google, you will probably find that updating your moviepy version to v1.0.3 or higher will fix the issue (`pip install moviepy --upgrade`). See #1185.**

---

### Expected Behavior
Expected to get an edited version of the inputted video.

### Actual Behavior
Produces this error:
```error
Moviepy - Building video test.webm.
MoviePy - Writing audio in testTEMP_MPY_wvf_snd.ogg
chunk:   0%|          | 0/552 [00:00<?, ?it/s, now=None]Traceback (most recent call last):
  File "/home/tech/Developement/CrabRaveGenerator/Test.py", line 8, in <module>
    ]).write_videofile("test.webm", fps = 30)
  File "</usr/local/lib/python3.5/dist-packages/decorator.py:decorator-gen-51>", line 2, in write_videofile
  File "/usr/local/lib/python3.5/dist-packages/moviepy-1.0.0-py3.5.egg/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, *a, **k)
  File "</usr/local/lib/python3.5/dist-packages/decorator.py:decorator-gen-50>", line 2, in write_videofile
  File "/usr/local/lib/python3.5/dist-packages/moviepy-1.0.0-py3.5.egg/moviepy/decorators.py", line 137, in use_clip_fps_by_default
    return f(clip, *new_a, **new_kw)
  File "</usr/local/lib/python3.5/dist-packages/decorator.py:decorator-gen-49>", line 2, in write_videofile
  File "/usr/local/lib/python3.5/dist-packages/moviepy-1.0.0-py3.5.egg/moviepy/decorators.py", line 22, in convert_masks_to_RGB
    return f(clip, *a, **k)
  File "/usr/local/lib/python3.5/dist-packages/moviepy-1.0.0-py3.5.egg/moviepy/video/VideoClip.py", line 317, in write_videofile
    logger=logger)
  File "</usr/local/lib/python3.5/dist-packages/decorator.py:decorator-gen-73>", line 2, in write_audiofile
  File "/usr/local/lib/python3.5/dist-packages/moviepy-1.0.0-py3.5.egg/moviepy/decorators.py", line 54, in requires_duration
    return f(clip, *a, **k)
[...]

PR #1059, opened 2020-01-16
Don't close the audio reader which might be linked from a shallow copy.

### LINKED ISSUE #1025: AttributeError: 'NoneType' object has no attribute 'get_frame'
MoviePy while effective, sometimes it's very unpredictable and unreliable. Once it runs, next time it fails. I was just running a concat on subvideos, now the same script is failing. I don't recall changing anything. Am I doing anything wrong?!
```
    for i in range(parts):
        cl = VideoFileClip(bestFilename[i])
        start = 0
        end = float(length / parts)
        s = cl.subclip(start, end);
        sub.append(s)
        cl.reader.close()
    final = concatenate([sub[0], sub[1].crossfadein(.5), sub[2].crossfadein(.5), sub[3].crossfadein(.5)], padding=-1, method="compose")
    final.write_videofile(output+filename+".mp4",audio=False)
```

```
    self.make_frame = lambda t: self.reader.get_frame(t)
AttributeError: 'NoneType' object has no attribute 'get_frame'

```

jwg4:
This fixes #1025 

A recent change closed the audio reader object when the video file reader object was closed. However the former could be shared between the clip being closed and another clip which was still needed, because of shallow copying, I think it's fine to let the garbage collector `del` the reader object when it's definitively no longer needed.

(Strikethrough - not applicable or not needed):

- [X] I have added a test to the test suite, if necessary
- [X] ~~I have properly documented new or changed features in the documention, or the docstrings~~
- [X] ~~I have properly documented unusual changes to the code in the comments around it~~
- [X] ~~I have made note of any breaking/backwards incompatible changes~~

### DIFF
--- a/moviepy/audio/io/AudioFileClip.py
+++ b/moviepy/audio/io/AudioFileClip.py
@@ -87,8 +87,8 @@ class AudioFileClip(AudioClip):
     def close(self):
         """ Close the internal reader. """
         if self.reader:
+            self.reader.close_proc()
             self.reader = None
 
     def __del__(self):
-        self.reader.close_proc()
         self.close()
--- a/moviepy/audio/io/AudioFileClip.py
+++ b/moviepy/audio/io/AudioFileClip.py
@@ -90,4 +90,5 @@ class AudioFileClip(AudioClip):
             self.reader = None
 
     def __del__(self):
+        self.reader.close_proc()
         self.close()
--- a/moviepy/audio/io/AudioFileClip.py
+++ b/moviepy/audio/io/AudioFileClip.py
@@ -87,7 +87,6 @@ class AudioFileClip(AudioClip):
     def close(self):
         """ Close the internal reader. """
         if self.reader:
-            self.reader.close_proc()
             self.reader = None
 
     def __del__(self):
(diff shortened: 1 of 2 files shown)


jwg4:
The appveyor build is broken - not just on my branch. I have fixed this in #1067

tburrows13:
I've confirmed that this fixes #938 which has been a _major_ bug for the past year on both macOS and Windows (and presumably linux). I wanted to properly investigate the handling and closing of external processes in v2.0, but that is likely still some time away so I'd like to release a v1.0.3 with just this PR in. The only catch is that _with_ this PR, Windows often fails with `OSError: [WinError 6] The handle is invalid` in Python 3.6 and below due to https://bugs.python.org/issue37380 (See also #1066), so basically we would only be able to support Py3.7+ on windows. I hope that a proper fix can be found for v2.0 that would re-enable support for Py3.6, but that will have to come later.

I don't have any time to investigate further anytime soon, so if we don't release this PR, then we'll presumably have to hold off until v2.0.
I think that this trade off is worth it, and it can be announced in the patch notes for v1.0.3.

@Zulko (or anyone else) please advise.

tburrows13:
Thank you for the contribution! The test in particular was very helpful when trying to diagnose the issue. It has been included in #1185, which has superseded this PR, so I am closing this one. I'd be happy to receive PRs about further improvements to how clips are deleted and closed, but they would have to be fully thought out and work the same with both video and audio.

PR #1062, opened 2020-01-18
headblur.py: Fix "ValueError: assignment destination is read-only"

berndf:
… is read-only" in line 45

When using moviepy/video/fx/headblur.py I get the error (numpy 1.17.4):

```
File "/usr/lib/python3.7/site-packages/moviepy/video/fx/headblur.py", line 45, in fl
    im[y1:y2, x1:x2] = mask*blurred + (1-mask)*orig
```

It seems that im is readonly and it also won't accept "im.setflags(write=1)":
  ValueError: cannot set WRITEABLE flag to True of this array

Therefore it seems we need to copy im and modify that.

### DIFF
--- a/moviepy/video/fx/headblur.py
+++ b/moviepy/video/fx/headblur.py
@@ -42,6 +42,7 @@ def headblur(clip,fx,fy,r_zone,r_blur=None):
         
         orig = im[y1:y2, x1:x2]
         blurred = cv2.blur(orig,(r_blur, r_blur))
+        im=im.copy() # Cannot modify im directly in numpy 1.17
         im[y1:y2, x1:x2] = mask*blurred + (1-mask)*orig
         return im
     


moviepy/video/fx/headblur.py
```diff
@@ -42,6 +42,7 @@ def fl(gf,t):
         
         orig = im[y1:y2, x1:x2]
         blurred = cv2.blur(orig,(r_blur, r_blur))
+        im=im.copy() # Cannot modify im directly in numpy 1.17
         im[y1:y2, x1:x2] = mask*blurred + (1-mask)*orig
```
mgaitan: what about to do this as a particular case, so we don't copy if it isn't strictly needed?
something like 

```suggestion
        try:
            im[y1:y2, x1:x2] = mask*blurred + (1-mask)*orig
        except ValueError:
            # Cannot modify im directly in numpy 1.17
            im= im.copy() 
            im[y1:y2, x1:x2] = mask*blurred + (1-mask)*orig
```

berndf: Hmm, for me it's a question of which system to optimize for. If, going forward, modifying im directly doesn't work, then that code will in the future always first try the failing code before doing the necessary thing. Your modification optimizes for outdated numpy versions while introducing a penalty for up-to-date versions. Of course, any trick to avoid the copy in current numpy is more than welcome!

jellyb0y:
Hi! I think this is so much relevant, at least for me =)
Could you merge it into next dev release ?!

tburrows13:
@jellyb0y this has been fixed in #1414 :)

PR #1067, opened 2020-01-29
Upgrade pip by calling via python (in appveyor).

jwg4:
Currently the appveyor build is broken. See for instance https://ci.appveyor.com/project/Zulko/moviepy/builds/30261132

As recommended in the appveyor error message, we can fix by  calling `python -m pip` instead of `pip` when upgrading pip.

(The usual tasks don't seem applicable here.)


tburrows13:
Thank you very much for sorting this out.

PR #1077, opened 2020-02-09
refactor Pythonic sake

mgaitan:
A general refactor to simplify common patterns a make a the codebase a bit more *pythonic*. Flatten code, reduce complexity, less lambdas, and more

I've also augmented tests (and simplify them using pytest helpers) in few cases.


Zulko:
Hi Martin, thank you very much for that. I am completely under the water but I clicked on some of these and they really make sense. One thing I am not sure about is when you replace "if x is not None" (which accepts x=0) by "if x" (rejects the case x=0), this is not equivalent but in these particular cases I am not sure it matters.

Do you want push rights on this repo? that would allow you to merge your recent PRs if you don't get other comments in the next few days.

moviepy/video/VideoClip.py
```diff
@@ -577,8 +546,9 @@ def blit_on(self, picture, t):
         # is the position relative (given in % of the clip's size) ?
         if self.relative_pos:
             for i, dim in enumerate([wf, hf]):
-                if not isinstance(pos[i], str):
-                    pos[i] = dim * pos[i]
+                if isinstance(pos[i], str):
```
tburrows13: I'm not sure that this is a worthwhile change.

Zulko: Oh yes true, thanks for catching that.

mgaitan: ok, reverted

mgaitan:
hi @Zulko. thank you very much for the offer of permission, I'm very glad, Of course I could accept it, with this little disclaimer: I don't have much time (you could check easily that my own libraries are a bit abandoned) and I prefer to stick to this "PR approach" instead to push directly and receive the review of most active contributors. 

Regarding your comment, you're right, may be it's better to be conservative and keep the `is None` as a sentinel checking and exclude the zero case.   

Of course this PR still needs some work to fix the regressions and hopefully add some tests to those functions/method that were somehow refactored and doesn't have a good coverage. 

Meanwhile, keep checking and commenting what you see.

moviepy/tools.py
```diff
@@ -65,39 +65,41 @@ def is_string(obj):
...
+    15.4 
+    >>> cvsecs((1, 21.5))   # (min,sec) 
+    81.5 
+    >>> cvsecs((1, 1, 2))   # (hr, min, sec)  
+    3662  
+    >>> cvsecs('01:01:33.045') 
+    3693.045
+    >>> cvsecs('01:01:33,5')    # coma works too
+    3693.5
+    >>> cvsecs('1:33,5')    # only minutes and secs
```
mgaitan: this is a new format supported. Is it ok, right?

mgaitan:
@Zulko @tburrows13 I guess I gonna stop change things now. Please take a view and merge it if you like.

moviepy/video/fx/crop.py
```diff
@@ -32,32 +32,26 @@ def crop(clip, x1=None, y1=None, x2=None, y2=None,
...
-        if x1 is not None:
-            x2 = x1+width
-        else:
-            x1 = x2-width
-    
-    if height and (y1 is not None or y2 is not None):
-        if y1 is not None:
-            y2 = y1+height
-        else:
-            y1 = y2 - height
```
tburrows13: Similar problems here:
If you pass in `x1=0, y1=0, width=5, height=5` then your version performs incorrectly because all 4 of the above expressions evaluate to `False`. Therefore, `x2` and `y2` get set incorrectly.

Use

```
if width and x1 is not None:
    x2 = x1 + width
```
etc

moviepy/video/tools/cuts.py
```diff
@@ -240,33 +239,27 @@ def select_scenes(self, match_thr, min_time_span, nomatch_thr=None,
...
             
-            poor_matches = set([end for (end,d_min, d_max) in ends_distances
-                            if d_min>nomatch_thr])
-            short_matches = [end for end in ends
-                             if (end-start)<=0.6]
+            poor_matches = {end for (end,d_min, d_max) in ends_distances if d_min > nomatch_thr}
+            short_matches = [end for end in ends if (end-start) <= 0.6]
             
-            if len( poor_matches.intersection(short_matches) ) == 0 :
+            if poor_matches & short_matches:
```
tburrows13: This line is wrong. Not only does it do the exact opposite, you can't use the `&` on a list and a set.

`if not poor_matches & set(short_matches):` is correct, but I think that `if not poor_matches.intersection(short_matches):` is the most readable, considering how often the `&` is used as a set intersection operator. Feel free to disagree.

moviepy/video/tools/drawing.py
```diff
@@ -162,20 +162,20 @@ def color_gradient(size,p1,p2=None,vector=None, r=None, col1=0,col2=1.0,
     elif shape == 'radial':
         if r is None:
             r = norm
-        if r==0:
+        elif r == 0:
```
tburrows13: I don't think that this change is correct. The first `if` sets `r`. The subsequent `if` and `else` set `arr`. You could put an extra new line in-between the 2 `if`s to help distinguish them apart instead.

tests/test_tools.py
```diff
@@ -7,56 +7,54 @@
...
-        right = rights[i]
-        message = "{0} did not get associated with {1}".format(left, right)
-        assert left == right, message
+    assert tools.find_extension(given) == expected
 
-def test_2():
-    """Test for raising erre if codec not in dictionaries."""
 
+def test_find_extensions_not_found():
+    """Test for raising erre if codec not in dictionaries."""
```
tburrows13: erre -> error (not your fault, but might as well fix it :) )

moviepy/video/tools/drawing.py
```diff
@@ -219,27 +219,26 @@ def color_split(size,x=None,y=None,p1=None,p2=None,vector=None,
...
     if grad_width or ( (x is None) and (y is None)):
-        if p2 is not None:
+        if p2:
             vector = (np.array(p2) - np.array(p1))
-        elif x is not None:
+        elif x:
             vector = np.array([0,-1.0])
             p1 = np.array([x, 0])
-        elif y is not None:
+        elif y:
```
tburrows13: Not sure about this one. Under your change, if only one of x and y is passed in (which is the intended way of doing it), _but_ it is passed in as 0, then it will crash on line 235 with a `TypeError: cannot unpack non-iterable NoneType object` because `vector` will not be set.

Under the original version, vector and p1 get set and it will _probably_ work fine, setting it all to one colour, which, though useless, I think is preferable.

mgaitan:
@tburrows13 thanks for the careful review. I'm in the middle of a trip now, but gonna update this following your comment ASAP, probably during the next wekeend.

tburrows13:
Sure, no hurry

mgaitan:
this is ready for a new review/merge .

moviepy/tools.py
tburrows13: Looks fine to me

mgaitan:
So, could we merge this or an extra review is needed?

mgaitan:
@Zulko could you check and merge this one? It's blocking the other refactors proposed. 

Alternatively, give me permission on the repo.  (I'll use it with care, promise)

tburrows13:
I've got no issues with merging, so I'll go ahead and do it. I'd happily support giving @mgaitan the Collaborator role as well. I think if you merge the current master branch into your other PRs, then they'll be sorted now.

mgaitan:
excellent, thanks!

tburrows13:
Sure! I've been away this last weekend, but now that my university term has finished, I'm keen to get v1.0.2 out within a few days. Feel free to comment on any PRs that you think would be suitable and I can add them to the milestone.

Zulko:
Thank you so much guys for taking care of the library! I am a bit out of touch but super happy to see that!

Martin I sent you a collaborator invitation, maybe add your name in the maintainers list in the Readme? Maybe also mark Tom and yourself as main maintainers, since you are currently the most up-to-date on the library?

PR #1082, opened 2020-02-20
Fixed potential crash trying to call the logger string as a function

tburrows13:
If `logger` was not passed in as a `proglog.ProgressLogger` object, then it would have crashed when it tried to call `logger` as a function.

See VideoClip.py:376 for correct equivalent.


tburrows13:
This problem would never actually manifest itself unless you call `ffmpeg_write_video()` on its own, because the it is only called from `VideoClip.write_videofile()` which sets `logger` explicitly anyway.

PR #1084, opened 2020-02-22
Clean up imports

tburrows13:
I've gone through the entire repo with [isort](https://isort.readthedocs.io/en/stable/). I've clicked through each one, and they are all just reordering the imports. The only exception is `editor.py` which is rather unique, with lots of ifs and trys in order to import _everything_, so I've left it as it is.


mondeja:
We can set up isort checking as part of CI so we don't need to make this cleanup periodically. Do you agree @tburrows13?

tburrows13:
I'm not too bothered. It is just another thing that people would have to run before committing, and I don't think it really matters that much. Someone doing it 'manually' every few months/years like I've done is probably sufficient, but I don't mind if you want to add it to the CI.

tburrows13:
Also there's a possible chance that it conflicts with `black`, so ensure that's not the case before going any further!

mondeja:
> I'm not too bothered. It is just another thing that people would have to run before committing, and I don't think it really matters that much.

You are right, three commands are too much, but this could be fixed easily with pre-commit and adding a `dev` optional requirements group. Inside that group would be all other dependencies (test, lint, optional, pre-commit, isort...). If we explain how to install (pretty simple, just `pre-commit install` after install dependencies with `pip install -e .[dev]`) wouldn't be needed execute anything before commits, the CI will be executed locally formatting files in place and raising flake8 errors blocking the commit if found.

> Also there's a possible chance that it conflicts with black, so ensure that's not the case before going any further!

Don't worry, it's just a bit of configuration, I've done it before :+1:

So, what do you say, `pre-commit` + `isort` and to forget executing the linting process before commits?

Issue #1097, opened 2020-03-06
Formatting code with Black


mgaitan:
I propose to enforce code formatting with [Black](https://github.com/psf/black)

Consider the arguments developed in https://github.com/django/deps/blob/master/accepted/0008-black.rst

- There's no debate about the usefulness of readable code formatting. 
- With automated code formatting, this workload can be transferred to computers, saving human bandwidth for higher value-added activities.
- It's about making our Python code look as much as possible like what everyone else in the Python community writes.

tburrows13:
Yes please! I suppose the best way would be to run it on the entire project locally and submit that as a PR. Then use some tool like https://github.com/marketplace/actions/black-code-formatter to keep it consistent?

mgaitan:
Implemented here https://github.com/mgaitan/moviepy/pull/1

I'll open the proper PR after #1077 be merged.

Issue #1081, opened 2020-02-19
Drop support for python < 3.6


mgaitan:
I've been working in a couple of refactors and I feel it's time to stop supporting both python 2.7 and python 3.5, and ensure support to python 3.7 (see #1018), 3.8 and any new python release in the future. 

As you may know, python2.7 reaches its end-of-life date starting 2020, and 3.5's one is the next september. Several of the most relevant libraries of the python ecosystem (e.g, django and numpy, just to mention few) are already python3 only since a while. 

This would simplify the code and the tests a bit (not a big deal, to be sincere), but most importantly, it'll allow us to start using new idioms/features of the language without hacks, conditional requirements, etc.

Zulko:
I am in favor. Anyone against?

tburrows13:
According to `pypinfo`, 10% of downloads in the last 30 days from PyPI are on Python 2. On https://pypistats.org/packages/moviepy it hovers at around 20% (same source data as `pypinfo`, so probably just because of those massive spikes).

I agree that it is worth doing. Python 3.5 I'm not so sure if the tiny benefits outweigh the decreased support. I don't really mind though.

For completeness, here are some PyPI stats from the last 30 days:

| version   | percent | download_count |
| --------- | ------- | -------------- |
| 1.0.1     |  73.94% |        146,036 |
| 0.2.3.2   |  13.82% |         27,296 |
| 1.0.0     |   5.31% |         10,481 |
| 0.2.2.11  |   4.83% |          9,538 |
| 0.2.3.5   |   1.70% |          3,359 |
| 0.2.2.13  |   0.20% |            393 |
| 0.2.3.4   |   0.07% |            132 |
| 0.2.3.1   |   0.06% |            125 |
| 0.2.1.6.3 |   0.04% |             75 |
| 0.2.3.3   |   0.04% |             71 |


| python_version | percent | download_count |
| -------------- | ------- | -------------- |
| 3.6            |  63.88% |        127,917 |
| 3.7            |  11.74% |         23,507 |
| 2.7            |  10.84% |         21,699 |
| None           |   6.01% |         12,041 |
| 3.5            |   3.95% |          7,906 |
| 3.8            |   3.50% |          7,002 |
| 3.4            |   0.07% |            148 |
| 3.9            |   0.01% |             16 |
| 2.6            |   0.00% |              2 |
| 3.2            |   0.00% |              1 |


| system_name          | percent | download_count |
| -------------------- | ------- | -------------- |
[...]

PR #1101, opened 2020-03-17
Apply and ensure Black formatting

mgaitan:
Fix for #1097


tburrows13:
So am I correct in saying that this doesn't actually auto-format the code, it just marks PRs as unsuccessful (and therefore unmergable) if they fail?

mgaitan:
@tburrows13 yes, you are right.

mgaitan:
well, I don't know why the git action didn't run. It was automatic when I tried this on my fork. 
It seems related to this bug https://github.community/t5/GitHub-Actions/Github-Workflow-not-running-from-pull-request-from-forked/td-p/32979. 

Anyway, should be ok after merge.

mgaitan:
And I'm wondering if wouldn't be better to implement #1081 first, removing old python version from the CI matrixes, and apply this targeting directly py36. So, we don't need another PR to apply a new black rewrite.

tburrows13:
I agree, I intend to start a new branch today for v2.0 that remove all py27 testing. Then we can merge PRs like this one into it.

tburrows13:
A bit later than expected, I've created #1103. When #1103 is merged into the v2 branch, you can update this PR with a fresh run of Black (without worrying about Py2.7, and then we can merge this in before anything else is merged).

I think adding something into the 'Standard workflow' in `Contributing.md` would be helpful as well.

Also I think the name 'sanity checks' is not very descriptive. How about 'Code Format Check' or something like that?

mgaitan:
@tburrows13 update done.

tburrows13:
Thanks for this. I'm happy to merge now.

PR #1103, opened 2020-03-20
Start v2.0 branch. Remove support for Python 2.7 & 3.4.

tburrows13:
Changes in this PR:

- Remove most compatibility try/except blocks that allowed Py2.7 support.
- Remove compat.py and replace all places where it was used with hardcoded Py3.7 code.
- Similar with tools.is_string
- Stop Travis testing on Py2.7
- Clean up version restrictions in setup.py

I've not removed Py3.5 support because there doesn't appear to be any convincing reason to unless we decide to explicitly use a Py3.6 feature or we start hitting compatibility problems with it. There is only one difference it makes at the moment; there seems to be issues with testing it with matplotlib, but tests are easily skipped with `@pytest.mark.skipif(PYTHON_VERSION == '3.5' and TRAVIS, reason="travis py35")` like in `test_examples.py`.


tburrows13:
I'm happy for this to be merged, if someone else would like to press the button. Closes #1081.

Issue #1096, opened 2020-03-06
Make effects be callable classes


mgaitan:
This is just a raw proposal 

currently, effects are functions that receive the clip as the first argument, and the rest are arguments for the fx itself. If we refactor them as classes where the init receive the arguments of the fx and  `__call__` receives the clip instances, the fx function 

```
        >>> clip.fx( volumex, 0.5).fx( resize, 0.3).fx( mirrorx )
```

could be something like 

```
 >>> clip.fx(volumex(0.5), resize(0.3), mirrorx())
```

wich, in my opinion is more explicit and readable, and allow better code reuse. 

In addition, making them instances allow us to implement further niceties like pipe effects (see #1076 )

```
clip | volumex(0.5) | resize(0.3) | mirrorx()
```

Lastly, we could do it just defining a decorator that convert functions to classes dynamically, and adding some type checking to `fx` to make it backward compatible.

tburrows13:
This would definitely be an improvement than the current syntax that you demonstrated, but I think that the method that editor.py uses is even better, which is just `clip = clip.volumex(0.5).resize(0.3).mirrorx()`.

However, not all effects and transitions are currently available for use this way because they are not added as methods to VideoClip in editor.py. I propose just updating this list in preference to making them classes like you suggest. (I'll start working on this now).

Feel free to disagree!

tburrows13:
My comment is now implemented in #1104, so if you're happy with that, then we can close this.

PR #1104, opened 2020-03-20
Support improved syntax for all effects in editor.py

tburrows13:

Previously, certain effects were being added as methods to VideoClips and AudioClips. This makes the syntax for using them much easier, however, not all effects were included in this. Seeing as the list of effects are already generated in `moviepy/video/fx/all/__init__.py` and `moviepy/audio/fx/all/__init__.py` (imported into `editor.py` as `vfx` and `afx` respectively), it made sense to instead use these lists instead.

Note that there is no such list for `transfx` (which are just the contents of `moviepy/video/compositing/transitions.py` so this is still just manually implemented).


mgaitan:
@tburrows13 I totally agree this is a much simpler API than my "class based" idea. What I really don't like is the implementation. 

IMHO we should try to get rid off the `editor.py` module, and do "the magic" directly in the classes, using proper metaprogramming tools that python offers, instead the nasty and dangerous `exec` monkeypatching. 

In this way, in addition to be more "pythonic", we could do just `from moviepy import VideoClip` and that would be enough to have the full featured class. 

If you like, I could do a PoC. Meanwhile, you could close #1096 and any other related ticket.

mgaitan:
Please check #1105 for a simple refactor proposal based on this one

tburrows13:
Closing in favour of #1105

Issue #591, opened 2017-06-18
AttributeError: module 'moviepy.audio.fx.all' has no attribute 'audio_fadein'


JSkorpio05:
I have used cx_freeze to build a python project into a single folder with an .exe and it's dependencies, but when I run the .exe I get the error: `AttributeError: module 'moviepy.audio.fx.all' has no attribute 'audio_fadein'`

I have tried both `from moviepy.editor import *` and also `from moviepy.video.io.VideoFileClip import VideoFileClip` and here is the python code:

```
    pygame.display.set_mode((854, 480), pygame.NOFRAME)
    pygame.display.set_caption('©2017 CherryByte™ Software')
    pygame.mouse.set_visible(False)
    logo = VideoFileClip('CherryByte Logo.mp4')
    logo.preview()
    pygame.mouse.set_visible(True)
```
It seems to run fine from the IDE (PyCharm) but once built, it seems to fail. Here is a shot of the Traceback:
![2017-06-18 3](https://user-images.githubusercontent.com/22495680/27264490-b54d0dd6-5477-11e7-9653-eaa51843bb5a.png)

Python version
3.6.1 (v3.6.1:69c0db5, Mar 21 2017, 18:41:36) [MSC v.1900 64 bit (AMD64)]
Version info.
sys.version_info(major=3, minor=6, micro=1, releaselevel='final', serial=0)

ZiddyEng:
For everyone that has the same issue i solved it by modifying the selected  __init__ file shown in the picture below:

![File Location][1]


  [1]: https://i.stack.imgur.com/zoxBq.jpg

Inside it there is a piece of code that import every function inside the fx folder:

    __all__ = [name for _, name, _ in pkgutil.iter_modules(
        fx.__path__) if name != "all"]
    for name in __all__:
        exec("from ..%s import %s" % (name, name))
Comment this block and import manually every function needed, like so:

    from moviepy.video.fx.accel_decel import accel_decel
    from moviepy.video.fx.blackwhite import blackwhite
    from moviepy.video.fx.blink import blink
    from moviepy.video.fx.crop import crop
    from moviepy.video.fx.even_size import even_size
    from moviepy.video.fx.fadein import fadein
    from moviepy.video.fx.fadeout import fadeout
    from moviepy.video.fx.mirror_x import mirror_x
    from moviepy.video.fx.mirror_y import mirror_y
    from moviepy.video.fx.resize import resize
    #etc.

[...]

PR #1105, opened 2020-03-24
Fluent API for effects (monkeypatch as methods)

mgaitan:
This PR is a (little bit) cleaner implementation to patch "functions effects" as methods. Supersedes #1104

Basically, avoid to use `exec` and deprecates `fx.all` modules, importing every function in each package's init. 

In addition, it defines `__all__` to reduce the namespace pollution when import `from moviepy.editor import *`


moviepy/audio/fx/audio_loop.py
```diff
@@ -17,12 +17,17 @@ def audio_loop(audioclip, nloops=None, duration=None):
     >>> videoclip.set_audio(audio)
 
     """
+    try:
+        clip = clip.audio
+    except AttributeError:
+        # assume it's a an audioclip
```
mgaitan:
```suggestion
        # assume it's already an audioclip
```

moviepy/audio/fx/audio_loop.py
```diff
@@ -17,12 +17,17 @@ def audio_loop(audioclip, nloops=None, duration=None):
     >>> videoclip.set_audio(audio)
 
     """
+    try:
```
mgaitan: I did this to avoid exclude it during the `VideoClip` patching

mgaitan:
what happened here? you closed this in favor of something else?

tburrows13:
Not intentional, sorry!

tburrows13:
Oh, it closed itself when I deleted the v2 branch. Apparently there's no way to change the base now (???). You might have to make a new one, sorry.

mgaitan:
ack, no problem.

tburrows13:
I think it's all sorted now, you can ignore this little fiasco :)

moviepy/editor.py
```diff
@@ -6,19 +6,44 @@
 
 In particular it will load many effects from the video.fx and audio.fx
 folders and turn them into VideoClip methods, so that instead of
->>> clip.fx( vfx.resize, 2 ) # or equivalently vfx.resize(clip, 2)
+>>> clip.fx( vfx.resize, 2 ) or equivalently vfx.resize(clip, 2)
```
mgaitan, line 9:
```suggestion
>>> clip.fx( vfx.resize, 2 )     # equivalent to vfx.resize(clip, 2)
```

mgaitan:
@tburrows13 do you have an opinion about this? 

It's much humble that [what I had in mind](https://github.com/Zulko/moviepy/pull/1104#issuecomment-602127064), that basically was do this patching via a metaclass to avoid the need to import the "patched classes" from a different place (i.e avoid `editor.py` at all). I explored that idea a bit, but I then I realized it's not as easy I thought because of circular dependencies (after all, we are patching the "base" classes), 

However I guess it's cleaner enough for now, and much more pythonic than the execs and `all.py` modules. 

The unique downside (very minor IMHO) is that fx namespaces changed if for some reason you import effects directly from the packages. 

In master if you import something from the `fx` package, you get the module
 
```python
In [1]: from moviepy.video.fx import mirror_x                                                                                                                                                 

In [2]: mirror_x?                                                                                                                                                                             
Type:        module
String form: <module 'moviepy.video.fx.mirror_x' from '/home/tin/lab/moviepy/moviepy/video/fx/mirror_x.py'>
File:        ~/lab/moviepy/moviepy/video/fx/mirror_x.py
Docstring:   <no docstring>
```

and here you get the actual fx function for the same import

```python
In [3]: from moviepy.video.fx import mirror_x                                                                                                                                                 

In [4]: mirror_x?                                                                                                                                                                             
Signature: mirror_x(clip, apply_to='mask')
Docstring: flips the clip horizontally (and its mask too, by default) 
File:      ~/lab/moviepy/moviepy/video/fx/mirror_x.py
Type:      function
```

However, this way still works in the new implementation

```python
In [1]: from moviepy.video.fx.mirror_x import mirror_x                                                                                                                                        

In [2]: mirror_x?                                                                                                                                                                             
Signature: mirror_x(clip, apply_to='mask')
Docstring: flips the clip horizontally (and its mask too, by default) 
File:      ~/lab/moviepy/moviepy/video/fx/mirror_x.py
Type:      function
```

so, we are mostly ok.

moviepy/audio/fx/audio_loop.py
tburrows13: Isn't this the same as using the `apply_to_audio()` decorator?

tburrows13: Oh there's actually an `audio_video_fx()` decorator that the other audio fx use. Is there a reason that it cannot be applied here?

moviepy/editor.py
```diff
@@ -53,44 +77,25 @@
...
-    "afx.volumex",
-]:
-
-    exec("AudioClip.%s = %s" % (method.split(".")[1], method))
+audio_fxs = inspect.getmembers(afx, inspect.isfunction)
+video_fxs = itertools.chain(
+    inspect.getmembers(vfx, inspect.isfunction),
+    inspect.getmembers(transfx, inspect.isfunction),
+    audio_fxs,
+)
```
tburrows13, line 93: Wouldn't it be simpler to just do 

    video_fxs = (
        audio_fxs
        + inspect.getmembers(vfx, inspect.isfunction)
        + inspect.getmembers(transfx, inspect.isfunction)
    )

`itertools.chain()` seems unnecessary.

tburrows13:
This is definitely better than what we have currently, and better than #1104. Thanks for detailing how this change affects importing from the `fx` package, but I don't think that that should be a major concern, just something to point out in the changelog (which I can write before merging).

I wouldn't be opposed to getting rid of `editor.py` altogether as it is rather unpythonic, but as you point out, it might be hard to still keep the fx as methods of the classes. I'm not that familiar with package-level things, but one idea that I had was to move all of `editor.py` into `__init__.py`, so users just `import moviepy` or `from moviepy import *`. Having just tested it, it allows doing `from moviepy.video.io.VideoFileClip import VideoFileClip`, and then still being able to use `VideoFileClip.rotate()` (which is only added as a method in `moviepy/__init__.py()`). I think that we could go even further and disallow importing directly from `moviepy` (by editing `__all__`?) to enforce standard pythonic importing directly from their relevant submodules, but still do the fx as methods stuff in `moviepy/__init__.py()`.

Perhaps that last paragraph should be a different issue/PR, and that last sentence would be quite a bit more of a drastic change and might require reworking the hierarchy of moviepy if everyone is accessing the classes from their proper locations. Furthermore, I still don't know if doing things in `moviepy/__init__.py()` is any more pythonic than this PR.

Your thoughts would be appreciated :)

tburrows13:
Closes #591

tburrows13:
Ok, here's my current thinking.

Move most of the standard imports and the fx 'trickery' into the main `__init__.py`, and recommend that standard moviepy programs just import what they need from `moviepy`, not `moviepy.editor`.
However, leave the stuff in `moviepy.editor` that only makes sense when doing 'live' editing, such as `sliders` (which depends on matplotlib), `preview` and `show` (which depend on Pygame) and the ipython stuff. This means that `import moviepy.editor` should only be used for its original intended purpose, which is live editing of movies, and importing moviepy (without editor) should take less than half the time and it won't initialise a Pygame process.

We can also look at including extra imports into either `__init__.py` or `editor.py` that are not currently in `editor.py` where appropriate.

@Zulko @mgaitan I'd appreciate your feedback on this.

Issue #1107, opened 2020-03-25
Image Magick can't be found


maikeeb:
I am trying to follow a tutorial I found on youtube and all this did for him was add a 3 second clip before the mp4 of a white background with red text of 'my movie' in the middle.

```
from moviepy.editor import *

movie = VideoFileClip("my_video.mp4")

title = TextClip("My Movie", color='red', size=(1920, 1080), bg_color='white', fontsize=30)
titleclip = title.set_duration(3)
render = concatenate_videoclips([titleclip, movie])
render.write_videofile('render.mp4', fps=30)

```
When I run this code it gets an error and this is the error message that I get in pycharm.

```
Traceback (most recent call last):
  File "\venv\lib\site-packages\moviepy\video\VideoClip.py", line 1168, in __init__
    subprocess_call(cmd, logger=None)
  File "\venv\lib\site-packages\moviepy\tools.py", line 46, in subprocess_call
    proc = sp.Popen(cmd, **popen_params)
  File "\Python\Python37\lib\subprocess.py", line 800, in __init__
    restore_signals, start_new_session)
  File "\Python\Python37\lib\subprocess.py", line 1207, in _execute_child
    startupinfo)
FileNotFoundError: [WinError 2] The system cannot find the file specified

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/transition.py", line 5, in <module>
    title = TextClip("My Movie", color='red', size=(1920, 1080), bg_color='white', fontsize=30)
  File "\venv\lib\site-packages\moviepy\video\VideoClip.py", line 1177, in __init__
    raise IOError(error)
OSError: MoviePy Error: creation of None failed because of the following error:

[WinError 2] The system cannot find the file specified.

.This error can be due to the fact that ImageMagick is not installed on your computer, or (for Windows users) that you didn't specify the path to the ImageMagick binary in file conf.py, or that the path you specified is incorrect

Process finished with exit code 1


```

### Specifications

  - Python Version: 3.7
  - Moviepy Version:1.0.1
  - Platform Name: Windows

mgaitan:
Have you installed ImageMagick and followed the instructions in the README?
Check this please,  https://github.com/Zulko/moviepy#optional-but-useful-dependencies

maikeeb:
I followed the instructions until the point of finding the convert.exe file, I can't find it the image magickk folder in my program files.

maikeeb:
I fixed the problem thanks for sending me the documentation.

[...]

Issue #883, opened 2018-12-10
ImageMagick not found.


element465:
### Expected Behavior
To create the video in the sample that creates the Knights video.

### Actual Behavior
Get the dreaded ImageMagick is not installed error, even though it is installed and the    has been changed. I have had this issue multiple times after multiple reinstalls of all different versions of Ubuntu, I am running this on a Vultr VPS. Everything is up to date, and ImageMagick 6 is installed.

```
Analyzed the audio, found a period of 0.48 seconds
Analyzed the video, found a period of 1.23 seconds

[MoviePy] This command returned an error !Traceback (most recent call last):
  File "/home/movie/knight.py", line 78, in <module>
    font="Century-Schoolbook-Roman", color="white")
  File "/usr/local/lib/python2.7/dist-packages/moviepy/video/VideoClip.py", line 1173, in __init__
    raise IOError(error)
IOError: MoviePy Error: creation of None failed because of the following error:

convert: unable to read font `Century-Schoolbook-Roman' @ warning/annotate.c/RenderType/872.
convert: unable to read font `Century-Schoolbook-Roman' @ error/annotate.c/RenderFreetype/1127.
convert: not authorized `/tmp/magick-26237HHBVp6dUVpNW' @ error/constitute.c/ReadImage/412.
convert: unable to read font `Century-Schoolbook-Roman' @ warning/annotate.c/RenderType/872.
convert: unable to read font `Century-Schoolbook-Roman' @ error/annotate.c/RenderFreetype/1127.
convert: not authorized `/tmp/magick-26237XtZzKZXcqetx' @ error/constitute.c/ReadImage/412.
convert: unable to read font `Century-Schoolbook-Roman' @ warning/annotate.c/RenderType/872.
convert: unable to read font `Century-Schoolbook-Roman' @ error/annotate.c/RenderFreetype/1127.
convert: not authorized `/tmp/magick-26237bqOuXBGTR687' @ error/constitute.c/ReadImage/412.
convert: unable to read font `Century-Schoolbook-Roman' @ warning/annotate.c/RenderType/872.
convert: unable to read font `Century-Schoolbook-Roman' @ error/annotate.c/RenderFreetype/1127.
convert: not authorized `/tmp/magick-26237HgCa56Rpk2OI' @ error/constitute.c/ReadImage/412.
convert: no images defined `PNG32:/tmp/tmp9JOwXM.png' @ error/convert.c/ConvertImageCommand/3210.
.

[...]

Issue #793, opened 2018-05-07
config_defaults.py is a poor UX


mgaitan:
Currently, the README recommends to manually edit `config_defaults.py` to set a custom FFMPEG path as imageio's alternative (to be used under an offline enviroment, experimental version, etc.) 
Altought 
This manual intervention is very weird, not only because it's no so easy to know where is this module (for instance, check #789 or #320)  but because a python package not necessarily keeps in its source format after be installed. But moreover, it ties the configuration to the code. Why?

Altought it's possible to use enviroment variables as an alternative, it's only mentioned in `config_defaults.py`'s docstring,. 

On https://github.com/Zulko/moviepy/issues/148#issuecomment-87778674 I proposed to use a user-level configuration file to override default (built-in) configuration. It could be a ini-style or json file

Zulko:
Im just seing that now. I'm not sure why in #148 it is question of changing a file before installing, the original idea was to change the config from inside the code with something like:

```
from moviepy.config import change_settings
change_settings(FFMPEG_BINARY="path/to/ffmpeg/binary")
```

Looking back at the code it has been changed in a way that seems to also support a configuration file, [see here](https://github.com/Zulko/moviepy/blob/956c46256cac34fd8fa12930eed8bebc824e322e/moviepy/config.py#L81-L91). It looks like either my doing or @cclauss, I dont get the git blame, and I dont remember exactly. Should have commented it better at the time.

keikoro:
Not sure what the status or next steps on this are, but I tagged it with `documentation` for now...

PR #1109, opened 2020-03-25
Simpler custom configuration and better related docs

mgaitan:
- [x] I have properly documented new or changed features in the documention or in the docstrings

This is a proposal fix for #793  (and probably #1107, #883 and more issues related to custom config). It seems windows users have a lot of troubles particularly with the setup of ImageMagick and the instructions in the `readme`  were totally outdated and complicated (even with a duplicated paragraph). 

The point is that, currently, the way to setup a "custom" path to the external tools is through environment variables and **not** editing a python module as the README said. 

In addition to update the doc (please, feel free to correct/improve it, english is not my best skill)
, I deleted the `config_default.py` module because is fairly useless.  Environment variables are now documented, including how to set them via python or, optionally, via a dotenv file. 

As the config is via envars, functions to read/change config variables has no sense and were also removed.


mgaitan:
I would like to receive some feedback on this, guys. And please feel free to reword or correct  the prose, particularly in the docs.

moviepy/video/VideoClip.py
```diff
@@ -1245,8 +1245,8 @@ def __init__(
             subprocess_call(cmd, logger=None)
         except (IOError, OSError) as err:
             error = (
-                f"MoviePy Error: creation of {filename} failed because of the"
-                "following error:\n\n{err}.\n\n."
+                "MoviePy Error: creation of {filename} failed because of the"
+                f"following error:\n\n{err}.\n\n."
```
tburrows13, line 1249: I only discovered this accidentally by looking at my IDE word colouring, but definitely something for both of us to be aware of in the future:
The `f` before the string must be on the same line as the `{}` that you have the variable in because the way it works is through implicit concatenation of strings, so only the first line was an f-string. I've fixed it here and in the other place (line 77 of config.py) it occurred.

docs/install.rst
```diff
@@ -46,14 +54,63 @@ For advanced image processing you will need one or several of these packages. Fo
...
+    Normally you can leave this one to its default ('ffmpeg-imageio') in which
+    case imageio will download the right ffmpeg binary (on first use) and then always use that binary.
+
+    The second option is ``auto-detect``. In this case ffmpeg will be whatever
+    binary is found on the computer: generally ``ffmpeg`` (on Linux/macOS) or ``ffmpeg.exe`` (on Windows).
+
+    Lastly, if you want to use a binary at a special location on your disk, enter it like this::
+
+        FFMPEG_BINARY = r"path/to/ffmpeg" # on linux
+        FFMPEG_BINARY = r"path\to\ffmpeg.exe" # on windows
```
tburrows13, line 91: It seems a little strange leaving the syntax like this when the user is no longer writing them as Python variables. Presumably it should be like

    os.environ["FFMPEG_BINARY"] = "path/to/ffmpeg"

or like the actual contents of the `.env` file:

    FFMPEG_BINARY="path/to/ffmpeg"

Either of them might need the 2 versions (linux and windows), I'm not sure.

The same applies to ImageMagick on line 102.

tburrows13:
I've gone through your changes and made some tweaks to the wording in a few places. You've made a great improvement, and I think that it will now be much easier for users (especially windows users) to start using MoviePy.

Once you've had a look at my comments and we've sorted them, then I'm happy to merge.

README.rst
```diff
@@ -101,20 +97,6 @@ For advanced image processing, you will need one or several of the following pac
...
-
-For Windows users, before installing MoviePy by hand, go into the ``moviepy/config_defaults.py`` file and provide the path to the ImageMagick binary called ``magick``. It should look like this:
-
-.. code:: python
-
-    IMAGEMAGICK_BINARY = "C:\\Program Files\\ImageMagick_VERSION\\magick.exe"
-
-If you are using an older version of ImageMagick, keep in mind the name of the executable is not ``magick.exe`` but ``convert.exe``. In that case, the IMAGEMAGICK_BINARY property should be ``C:\\Program Files\\ImageMagick_VERSION\\convert.exe``
-
-For Ubuntu 16.04LTS users, after installing MoviePy on the terminal, IMAGEMAGICK will not be detected by moviepy. This bug can be fixed. Modify the file in this directory: /etc/ImageMagick-6/policy.xml, comment out the statement <!-- <policy domain="path" rights="none" pattern="@*" /> -->.
```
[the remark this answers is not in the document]
mgaitan: I've restored this paragraph in the doc's `install.rst`. Imho it's too specific to be in the README.

mgaitan:
@tburrows13 I've tried to clarify the example for setting config via dotenv and added the `check()` tip. 

please, check and merge.

PR #1113, opened 2020-03-27
Fixed name of optional dependency on README

jonnekaunisto:
The name of the extra dependency in the setup.py is doc, which is incorrectly documented in the README and will throw error " moviepy 1.0.2 does not provide the extra 'docs'"


mgaitan:
thanks!

Issue #1089, opened 2020-02-22
Moviepy Roadmap - version 2.0


tburrows13:
I’d like to propose an upcoming v2.0 which would allow us the opportunity to make a few small breaking changes, and remove some of the previously deprecated features.

Here’s some of the things that I’d like to include:

- [x] [#1103, #1106] Remove support for old Python versions (#1081)
    - Definitely 2.7 and 3.4, perhaps 3.5 as well?
    - Py3.5 is not EOL yet, accounts for about 2% of our installs (https://pypistats.org/packages/moviepy), and the number of [new features in Py3.6](https://docs.python.org/3.6/whatsnew/3.6.html) is very small (but it might be worth it just for the f-strings)
    - This would allow us to get rid of a lot of complicated and hacky code

- [ ] I’d like to go through the backlog of complaints about resource usage and proper closing of clips and create a consistent cross-platform solution that can work for everyone, using some to-be-decided combination of `.close()`, `.__del__()` and context managers (`with ...:`). This would most likely not be backwards-compatible (#1016, #1059).

- [ ] We could enforce use of “keyword-only” arguments with the * (https://www.python.org/dev/peps/pep-3102/)
    - For instance, the definition of VideoFileClip would become
        ```
        def __init__(self, filename, *, has_mask=False,
             audio=True, audio_buffersize=200000,
             target_resolution=None, resize_algorithm='bicubic',
             audio_fps=44100, audio_nbytes=2, verbose=False,
             fps_source='tbr'):
        ```
    - So `VideoFileClip("file.mp4", True, True, 100000)` would now be invalid

- [ ] If it were possible to get rid of ImageMagick as a dependency, that would be amazing, its possible that everything that we use it for can now be done with imageio or one of our other, better behaved, dependancies. (Discussion in #908, #1472)

- [ ] There are a lot of functions/methods with lots of optional parameters like https://github.com/Zulko/moviepy/blob/master/moviepy/video/fx/crop.py that require certain combinations of arguments to work. It would be nice to have a neat way of expressing this, maybe with decorators?

- [x] [#1115] We can remove old deprecations
  - `to_videofile` -> `write_videofile`
  - `to_audiofile` -> `write_audiofile`
  - [`ColorClip(col=...)`](https://github.com/Zulko/moviepy/blob/1c368a195dbbba68a398b93152bef1e638d141f2/moviepy/video/VideoClip.py#L1034)

- [x] [#1170] We could go even further and improve the API even more
[...]

PR #1116, opened 2020-03-29
Validate argument's combinations

mgaitan:
This is another PR for the #1089. Not all of the combinations are valid.  This is a simpler approach to validate, doing assertions before "override" (calculate) values. 

An alternative could be do some checking in advance using https://github.com/mgaitan/one (see examples), but I prefer to do it explicitly in a function and not in a decorator (_explicit is better than implicit_)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,98 @@
-# Change Log
+# Changelog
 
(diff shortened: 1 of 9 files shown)


mgaitan:
@tburrows13 which other functions needs this kind of validations?

OsaAjani:

PR #1118, opened 2020-03-30
Fixed error message for build docs

jonnekaunisto:
Changed "docs" to doc since that is the correct optional dependency name


tburrows13:
Well spotted, thanks for another fix!

PR #1121, opened 2020-03-31
Add new 'Unreleased' section to CHANGELOG.md.

tburrows13:
With this PR, I'd like to suggest that PR owners add explanations of their PRs into the 'Unreleased' section of CHANGELOG.md. This should reduce the burden of the releaser and ensure that the changelog of a feature is created while it is still under discussion and in people's minds. Hopefully it will also increase the usefulness of the changelog which is especially important with the breaking changes that are coming up that might catch people off guard.

I've gone ahead and populated it with the [changes that have been made so far](https://github.com/zulko/moviepy/compare/v1.0.2...HEAD).

The standard that I'm using is https://keepachangelog.com/en/1.0.0/. We could potentially also apply the categories listed there to PR labels as well?

I don't think that 'insignificant' PRs that only really affect the developers like #1118 and #1101 need to feature in the changelog.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,9 +2,9 @@
 Please tick when you have done these. They don't need to all be completed before the PR is created.
 Delete them if they are not appropriate for this pull request.
 -->
-- [ ] <!--Only relevant if this is a bug fix-->I have provided code that clearly demonstrates the bug and only works correctly when used with this PR
-- [ ] I have added suitable tests to the test suite in `tests/`
-- [ ] I have properly documented new or changed features in the documention or in the docstrings
-- [ ] I have properly documented unusual changes to the code in the comments around it
-- [ ] I have made note of any breaking/backwards incompatible changes
-- [ ] I formatted my code using `black -t py36` 
+- [ ] I have provided code that clearly demonstrates the bug and that only works correctly when applying this fix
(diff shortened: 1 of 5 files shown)


mgaitan:
I understand the goal of this PR but I'm afraid that add more requirements for a PR may be a burden  and could make casual contributors desist. And specially for many developers which are more confident writing python than english, myself included ;-) )

I would explore the idea to auto-generate the changelog, and more generally, the whole release process. 

An interesting tool for this seems to be https://github.com/michaeljoseph/changes

(but are dozens of similar ones, we need to research a bit) .

tburrows13:
All very valid points; thank you for your feedback.

I have already been using an automatic changelog generator (https://github.com/github-changelog-generator/github-changelog-generator, sort-of-documented here: https://github.com/Zulko/moviepy/wiki/How-to-create-a-release). This has been fine for the past few releases that have only had bug fixes and tiny feature additions. However, now that we are doing a bit more, I think that a personalised guide to updating to v2.0 is necessary.

I totally agree with your point about not wanting to make it too much of a burden for casual contributors. Either way, I'd expect to have to edit the final contents anyway, either to correct English, formatting or layout.
How about we remove the requirement for contributors to create their changelog entry, and if I decide that it could be useful, I could add it into their PR before merging? Before the final release I'll go back though the merged PRs/commits to check that nothing important has been missed.

mgaitan:
that makes sense to me.

tburrows13:
Thanks, I'll merge this. If you've got other concerns about the template, I'm very happy to discuss them!

PR #1122, opened 2020-03-31
videoclip adds a mechanism to save the video alpha channel

biandh:
1. update video_clip.writefile function, add save video alpha channel mechanism
2. update video_clip.set_audio function, covers some frequently occurring cases

- [ ] I have properly documented new or changed features in the documention or in the docstrings
- [ ] I formatted my code using `black -t py36`

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -268,7 +268,7 @@ class VideoClip(Clip):
         with_mask
           If video has alpha channel, and you want to keep it after processing, this parameter must be True
           You can use any FFMPEG encoding format that supports alpha channels.
-          Examples are 'png、qtrle' for '.mov'
+          Examples are 'png/qtrle' for '.mov'
           Note: 'png' encoding maybe takes long time, but the resulting video is small,
                 'qtrle' is the opposite
 
@@ -288,7 +288,6 @@ class VideoClip(Clip):
         >>> clip.write_videofile("my_new_video.mp4", codec='qtrle', with_mask=True) # is fail, FFMPEG will throw an exception: "codec not currently supported in container", because 'qtrle' not support '.mp4'
         >>> clip.write_videofile("my_new_video.mp4", codec='libx264', with_mask=True) # is bad, because 'libx264' not support alpha channel, video will lose the alpha channel
         >>> clip.close()
-
 
         """
         name, ext = os.path.splitext(os.path.basename(filename))
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -739,7 +739,7 @@ class VideoClip(Clip):
         self.size = self.get_frame(0).shape[:2][::-1]
 
     @outplace
-    def set_audio(self, audioclip, start_time=0, keep_original=True):
+    def set_audio(self, audioclip, start_time=0, keep_original=False):
         """
         Attach an AudioClip to the VideoClip.
         :param audioclip: new audioclip
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -34,10 +34,6 @@ from .io.ffmpeg_writer import ffmpeg_write_video
 from .io.gif_writers import write_gif, write_gif_with_image_io, write_gif_with_tempfiles
 from .tools.drawing import blit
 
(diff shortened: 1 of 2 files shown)


moviepy/video/VideoClip.py
```diff
@@ -721,13 +735,25 @@ def set_make_frame(self, mf):
...
         attribute set to ``audio``, which must be an AudioClip instance.
         """
-        self.audio = audioclip
+        if start_time > 0:
+            from moviepy.editor import concatenate_audioclips
+            dummy = audioclip.audio_loop(duration=start_time).volumex(0).subclip(0, start_time)
+            audioclip = concatenate_audioclips([dummy, audioclip])
+
+        if self.audio is not None and keep_original:
+            from moviepy.editor import CompositeAudioClip
```
tburrows13: You can import this at the top of the file, and don't import from moviepy.editor, use the exact file (moviepy.audio.AudioClip). Same goes for the import on line 748, and any others.

biandh: ok, I have fix it.

moviepy/video/VideoClip.py
```diff
@@ -264,6 +269,13 @@ def write_videofile(
         verbose (deprecated, kept for compatibility)
           Formerly used for toggling messages on/off. Use logger=None now.
 
+        with_mask
+          If video has alpha channel, and you want to keep it after processing, this parameter must be True
+          You can use any FFMPEG encoding format that supports alpha channels.
+          Examples are 'png、qtrle' for '.mov'
```
mgaitan, line 275: wouldn't be cool to warn if `with_mask` is true but the format doesn't support it?

moviepy/video/VideoClip.py
```diff
@@ -721,13 +739,24 @@ def set_make_frame(self, mf):
         self.size = self.get_frame(0).shape[:2][::-1]
 
     @outplace
-    def set_audio(self, audioclip):
-        """Attach an AudioClip to the VideoClip.
-
-        Returns a copy of the VideoClip instance, with the `audio`
+    def set_audio(self, audioclip, start_time=0, keep_original=False):
```
mgaitan, line 742: few notes

1. In general, would be much better to open independent PRs, one for each proposed new feature. Thiat would be cleaner for the history / release notes and also for the discussion.  
2. the idea of this seems handy but I'm not sure if it's not much "overspecialized" . is it a great advantage comparing with composite the audio clip in advance and then apply to the videoclip? what if instead a "silence" I would like a fadein? or what about if instead to start at a time I would like to end at a time? 

3. I'm sure there is a better way to create a "silence" clip than set `volume 0`. check examples of `make_frame` in the test.  

4. In any case, this feature would needs some tests

mgaitan: @Zulko your opinion would be valuable here, specially regarding my comment's point 2 and 3.

moviepy/video/VideoClip.py
biandh: if  format doesn't support, for example:

clip = VideoFileClip("myvideo.mov", has_mask=True).subclip(100,120)
clip.write_videofile("my_new_video.mp4", codec='qtrle', with_mask=True)  #The result will fail to execute, FFMPEG will throw an exception: "codec not currently supported in container", because 'qtrle' not support '.mp4'. 

but this:
clip.write_videofile("my_new_video.mp4", codec='libx264', with_mask=True) #The result will be executed, but will not meet expectations, because 'libx264' not support alpha channel, the video will lose the alpha channel.

This is almost an inevitable problem, and for this reason I have detailed this in the code comments.

tburrows13: Is this line supposed to read `Examples are 'png', 'qtrle' for '.mov'`? Your version has some strange comma character.

biandh: Yes, here are some encoding formats that support the alpha channel:
'png' -->mov/mp4/mkv/...
'qtrle' -->mov
'vp6a' -->flv/f4v
'vp8/vp9' -->webm
etc.

biandh: > Your version has some strange comma character.
I have fixed it.

mondeja:
Added in #1237.

PR #1141, opened 2020-04-09
Raise exception if ImageMagick binary path is unset

jonnekaunisto:
When making a TextClip such as in the example moving_letters.py without installing ImageMagick this error is received.

```Traceback (most recent call last):
  File "/Users/jkaunist/Desktop/Projects/moviepy/.venv/lib/python3.7/site-packages/moviepy/video/VideoClip.py", line 1200, in __init__
    subprocess_call(cmd, logger=None)
  File "/Users/jkaunist/Desktop/Projects/moviepy/.venv/lib/python3.7/site-packages/moviepy/tools.py", line 35, in subprocess_call
    proc = sp.Popen(cmd, **popen_params)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 775, in __init__
    restore_signals, start_new_session)
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 1522, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'unset': 'unset'

During handling of the above exception, another exception occurred:

  File "moving_letters.py", line 10, in <module>
    "Cool effect", color="white", font="Amiri-Bold", kerning=5, fontsize=100, print_cmd=True
  File "/Users/jkaunist/Desktop/Projects/moviepy/.venv/lib/python3.7/site-packages/moviepy/video/VideoClip.py", line 1212, in __init__
    raise IOError(error)
OSError: MoviePy Error: creation of None failed because of the following error:

[Errno 2] No such file or directory: 'unset': 'unset'.
.This error can be due to the fact that ImageMagick is not installed on your computer, or (for Windows users) that you didn't specify the path to the ImageMagick binary in file conf.py, or that the path you specified is incorrect 
```

With this change it will print.

```Exception: ImageMagick is not installed on your computer, or (for Windows users) that you didn't specify the path to the ImageMagick binary in file conf.py.```


tburrows13:
Hey, thanks for taking the time to make this change. I agree that it is an improvement over what is already present in MoviePy. However, the integration with ImageMagick is being redone in #1109 , which I believe makes these changes obsolete. You might be able to rework your suggestions here to work with #1109, but I'm not sure if it is necessary. In #1109, ImageMagick can always be auto-detected, so I don't think it will ever be 'unset'.

jonnekaunisto:
> Hey, thanks for taking the time to make this change. I agree that it is an improvement over what is already present in MoviePy. However, the integration with ImageMagick is being redone in #1109 , which I believe makes these changes obsolete. You might be able to rework your suggestions here to work with #1109, but I'm not sure if it is necessary. In #1109, ImageMagick can always be auto-detected, so I don't think it will ever be 'unset'.

I had a look at #1109, and the IMAGEMAGICK_BINARY  is still sometimes set to "unset". For example, if you do not have Image Magick installed on your computer at all I believe this would happen. 

To make this pull request work with #1109 I believe I just have to use IMAGEMAGICK_BINARY instead of get_setting("IMAGEMAGICK_BINARY"). I can change that after #1109 merges. I think my change would work with that change, but let me know if I missed something.

tburrows13:
Yes, I think you are right, good point. Let's revisit this then #1109 is merged.

jonnekaunisto:
I believe #1109 has merged and I've updated this PR to work with that. Please revisit this.

tburrows13:
Would you mind updating the error message? You originally copied it off the one further down the page, but that was changed in #1109: https://github.com/Zulko/moviepy/pull/1109/files#diff-4b7e3f0e11cb6f470780d69172d8fbe1L1216.

keikoro:
Pinging @jonnekaunisto regarding the requested update to the error message.

tburrows13:
I'm going to close this because ImageMagick is being removed before v2.0 releases.

PR #1144, opened 2020-04-14
Allow setting tempfile path

### LINKED ISSUE #892: RFE: Ability to chose path for temp file on write_videofile
I'm trying to use moviepy on a Google Cloud Function (similar do AWS Lambda) and can't get around this problem. Even though I can download the original movie and save the resized movie, in an specified directory, the operation fails because write_videofile creates a temp file wherever the code is running. On GCF's case it runs on a read-only fs, which makes the operation impossible.
The ability to chose where to store the temp file would be huge.

tburrows13:
Closes #892. Depends upon #1137.

A simple feature to allow the user to specify which directory the temporary audiofile should be created in.


tburrows13:
I'm ready for this to be merged.

Issue #1131, opened 2020-04-05
ColorClip's color doesn't default to black


mgaitan:
The docstring says 

> If argument ``ismask`` is False, ``color`` indicates  the color in RGB of the clip (default is black)

however

```
In [1]: from moviepy.editor import ColorClip                                                                                                                                                  

In [2]: c = ColorClip((10,10), duration=1)                                                                                                                                                    
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-5c98bac84586> in <module>
----> 1 c = ColorClip((10,10), duration=1)

~/lab/moviepy/moviepy/video/VideoClip.py in __init__(self, size, color, ismask, duration, col)
   1094                 color = col
   1095         w, h = size
-> 1096         shape = (h, w) if np.isscalar(color) else (h, w, len(color))
   1097         ImageClip.__init__(
   1098             self, np.tile(color, w * h).reshape(shape), ismask=ismask, duration=duration

TypeError: object of type 'NoneType' has no len()
```

PR #1147, opened 2020-04-15
fixed default value bug on ColorClip

jonnekaunisto:
Fixed bug in issue #1131

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -1050,7 +1050,20 @@ class ColorClip(ImageClip):
 
     def __init__(self, size, color=None, ismask=False, duration=None):
         w, h = size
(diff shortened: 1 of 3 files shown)


moviepy/video/VideoClip.py
```diff
@@ -1050,7 +1050,20 @@ class ColorClip(ImageClip):
...
+        if ismask:
+            shape = (h, w)
+            if color is None:
+                color = 0
+            elif not np.isscalar(color):
+                raise Exception("Color has to be a scalar when mask is true")
+        else:
+            if color is None:
+                color = (0, 0, 0)
+            elif hasattr(color, "__getitem__"):
```
tburrows13, line 1063: Is this supposed to be `not hasattr`?

tburrows13:
Thanks for the ping. It looks great.

Issue #879, opened 2018-12-05
Hesitant installation instructions for OpenCV in the documentation are no longer needed


chrisconlan:
I thought this was pretty funny: <http://zulko.github.io/moviepy/opencv_instructions.html> ... because I have also struggled with OpenCV installation in the past.

Thankfully, there has been a new development that solves this problem. This package: <https://pypi.org/project/opencv-python/> is very popular and installs OpenCV 3 with a `pip` one-liner.

`pip install opencv-python`

I have used it a few places. It seems to respect virtual environments smoothly.

Just FYI so you can stop telling users to install OpenCV from source.

PR #1148, opened 2020-04-15
deleted outdate opencv install instructions and fixed english syntax

jonnekaunisto:
deleted outdated opencv instructions as referenced by issue #879


tburrows13:
Thanks for this. Do you want to edit install.rst and readme.md to change the info about opencv to point to this Python package instead?

PR #1151, opened 2020-04-17
Flake8 formatting cleanup

tburrows13:
Nothing of any substance is supposed to have changed in this PR. I've simply gone through all the flake8 warnings and corrected the code where necessary, or specifically ignored that type of error where it conflicts with black.

Most notable changes:
All raw `except:` have been replaced with `except Exception:`
All assigned lambdas have been replaced with a `def`
The Travis build will now fail when flake8 detects undefined variable names. (Removal of `--exit-zero` on the important check)
All regex strings with backslashes have been made 'raw' by putting an `r` at the start of the string. This prevents a `DeprecationWarning`.


moviepy/audio/io/readers.py
```diff
@@ -162,8 +162,6 @@ def close_proc(self):
             self.proc = None
 
     def get_frame(self, tt):
-
-        buffersize = self.buffersize
```
tburrows13, line 166: Unused variable

tburrows13:
I'm ready for this to be merged as well.

PR #1152, opened 2020-04-18
Add graphics blend modes (replacement for #785)

tburrows13:
This PR is basically #785 with a few tweaks to make it suitable for the current codebase. I tried both pushing it directly to https://github.com/spirable/moviepy/tree/feature/add_blend_modes and opening a PR there, but neither worked, so I've created a new PR here instead.


moviepy/video/compositing/BlendedCompositeVideoClip.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def playing_clips(self, t=0):
+        playing = []
+        for index, clip in enumerate(self.clips):
+            if clip.is_playing(t):
+                playing.append((clip, self.clips_blending[index]))
+        return playing
+
+    def _check_clips_blending(self, clips_blending):
+        if clips_blending is None:
```
amirothman, line 118: why not set the default value of `clips_blending` at the `__init__` params instead? seems that might be a more a natural place to set this value

moviepy/video/compositing/BlendedCompositeVideoClip.py
```diff
@@ -0,0 +1,126 @@
...
+            clips_blending = []
+        n_clips = len(self.clips)
+        if len(clips_blending) != n_clips:
+            clips_blending = [
+                {"blend_mode": "normal", "blend_opacity": 1.0, "blend_weight": 1.0,}
+                for i in range(n_clips)
+            ]
```
amirothman, line 125: why are replacing all of the items in list of parameters instead of just adding/subtracting according to the difference in length? this seems to be a bit confusing, maybe a little bit of explanation in its comments?

tests/test_compositing.py
```diff
@@ -39,3 +39,28 @@ def test_clips_array_duration():
...
+    comp = BlendedCompositeVideoClip(
+        [bg, t1, t2],
+        clips_blending=[
+            {"blend_mode": "normal"},
+            {"blend_mode": "soft_light", "blend_opacity": 0.8},
+            {"blend_mode": "hard_light", "blend_weight": 0.8},
+        ],
+    )
+    comp.write_videofile(join(TMP_DIR, "test_blended_composition.mp4"))
+    return
```
amirothman, line 66: how come there are no assertion(s) on this test?

tburrows13: Well it is really just to check that it does not crash. I don't know how you'd actually be able to check the results given how many pixels there are. I suppose that I could generate a clip that I check now, and then compare that static clip with the generated clip each time. Would that seem like a better idea worth doing? I suppose that would also require that `VideoFileClip` for loading the 'correct' clip is deterministic enough, which I'm not sure that it is.

tburrows13:
Thanks for the review @amirothman. It looks helpful, and I'll be sure to consider it carefully. However I'm will be pretty busy for the next few weeks, so I shall have to resist spending lots of time on it in the meantime, as much as I would love to!

tests/test_compositing.py
amirothman: cool. that seems like a legit reason. maybe can indicate in the comments or something

amirothman:
> Thanks for the review @amirothman. It looks helpful, and I'll be sure to consider it carefully. However I'm will be pretty busy for the next few weeks, so I shall have to resist spending lots of time on it in the meantime, as much as I would love to!

no worries. thanks for the contribution :D

tburrows13:
Note to self: implement support for layers (#1176) here as well.

RenatoLopes771:
Hello @tburrows13. I think you ended up forgetting about this PR. Be sure to give it a check again, as to not make your code go to waste.

ShakedDovrat:
+1 asking for this feature

OsaAjani:

PR #1153, opened 2020-04-20
Fixed a typo

spollard:
A simply typo


tburrows13:
Thank you for spotting that! (Ignore the failed check, that's my fault - fixed now).

PR #1155, opened 2020-04-20
added missing media file chaplin

jonnekaunisto:
Found here: https://github.com/earney/moviepy_media


tburrows13:
Linking #749 for future reference.

tburrows13:
Thanks for the contribution. I'm going to leave unmerged for now if that's ok. When we are closer to the v2.0 release I'd like to do a push to thoroughly update the docs and examples, which have been neglected for a while. If this is needed during the overhaul, then I'll merge it. Thanks.

PR #1165, opened 2020-04-28
updated dockerfile dependencies to come from setup.py

jonnekaunisto:


tburrows13:
Hey thanks for this change. I was wondering if you'd actually tested moviepy on docker recently? I'd noticed that the dockerfile seemed out of date, but because I'd never used docker, I didn't feel qualified to fix it. If this change is all that is needed to make it work, that would be great.

jonnekaunisto:
Yeah this commit mostly just keeps the dependencies updated, but it would probably make more sense to use the alpine image instead of the debian image for the dockerfile, since alpine is basically the standard image to use.

Updating to alpine would require mostly just finding the packages from the alpine packages list, which sometimes have slightly different names than debian ones.

I'm happy to take a crack at that if it makes sense.

tburrows13:
I've got no opinion on alpine vs debian/ubuntu. If you think that it would be an improvement, then go for it, but from my (limited) research, it seems like debian is a perfectly valid choice.

jonnekaunisto:
Yes debian is perfectly fine. The dockerfile with these configurations fails for one test, which is test_matplotlib.

Error is: _tkinter.TclError: no display name and no $DISPLAY environment variable

Which doesn't seem too bad since the docker container doesn't have any display

tburrows13:
Is this PR finished? Are you happy for it to be merged?

jonnekaunisto:
Yes, please merge this pull request.

PR #1157, opened 2020-04-23
Change ways to composite video clips by using PIL

ODtian:
This is mainly due to the performance issues on CompositeVideoClip, by changing dependency to PIL, it becomes 25 times faster.

Issue #1125, opened 2020-04-01
What is the meaning of 'fl'?


tburrows13:
`fl` appears all over the source code, however I've never known what it actually means. Is it 'filter', or something else? Would it be possible to change the function names to something more explicit?

Zulko:
``fl`` means filter and I am sorry (moviepy was my first ambitious API). I could be replaced by "filter", but if you go down the route of changing API names for v2, I'd suggest a complete rethink. One of the most confusing things with moviepy is that is uses ``fl``, ``set_duration``, ``set_start``, while these methods are not in-place. A clearer interface would be ``with_duration``, etc. so that a typical code would read:

```
new_clip = clip.with_duration(5).with_filter(black_and_white)
```

tburrows13:
I'd support renaming all such methods in v2.0. @mgaitan thoughts?
And we'd have to decide whether to deprecate or fully remove the old ones.

tburrows13:
Closing as `fl` has been replaced with `transform` in v2.0 now.

Issue #886, opened 2018-12-15
TextClip positioning does not work


TowelDude:
I have written the following function to add a timestamp to a video:
```

def stamp_clip(clip, start_t=0):
    clips = []
    for x in range(int(clip.duration)):
        txtclip = TextClip(str(timedelta(seconds=x + start_t)),
                           font='Arial-Black', fontsize=50, stroke_color='black', stroke_width=2, color='white')

        txtclip.duration = 1
        txtclip.set_position((100, 200))
        sub = clip.subclip(x, x+1)
        # clip.add_mask()
        comp = CompositeVideoClip([sub, txtclip])
        comp.duration = 1
        clips.append(comp)

    return concatenate_videoclips(clips)

```

### Expected Behavior
A time stamp will appear at `(100,200)`.

### Actual Behavior
The timestamp does not move.

I have tried many overloads of this function `"center","right"`, `0.2,0.5` etc.  and nothing worked

### Steps to Reproduce the Problem
The function above.

### Specifications

  - Python Version: 2.7.15
  - Moviepy Version: 0.2.3.5
  - Platform Name: Windows 
  - Platform Version: 10

TowelDude:
Update:

Changing the line:
```
  txtclip = TextClip(str(timedelta(seconds=x + start_t)),
                           font='Arial-Black', fontsize=50, stroke_color='black', stroke_width=2, color='white')
```

To the following:


```
  txtclip = TextClip(str(timedelta(seconds=x + start_t)),
                           font=font, fontsize=font_size, color='white', stroke_color='black', stroke_width=1) \
                           .set_duration(1) \
                           .set_pos(('left', 'bottom'))
```

and, of course, removing redundant rows, solves the issue. 

I really have no idea what caused the issue, but the solution might give some hints for where the bug is.

Cheers

Sv3n:
All functions that have an `@outplace` decorator return a new modified clip, so you need to store their return value. See: https://zulko.github.io/moviepy/getting_started/effects.html for some documentation on this behavior.

Zulko:
Sorry for the confusion, I understand that "with_duration", "with_position" would have been better names than "set_duration", "set_pos", to indicate that the transformations are not in-place, but hey I was young at the time.

c4llous:
the correct syntax will be txt.clip((100,200))

PR #1170, opened 2020-04-30
Renaming methods, parameters and variables for increased usability, consistency and readability

tburrows13:
> Explicit is better than implicit.
\- [The Zen of Python](https://www.python.org/dev/peps/pep-0020/)

There have been discussions in the past about how some of MoviePy's names are misleading and unintuitive (#886, #1125). I think that v2.0 is a good time to attempt to increase the usability and consistency of the package. This PR also changes a lot of the internal variable names to hopefully increase readability.

Because this is such a massive PR (just about every file has been changed in some way), I've kept a note of every change I made, which is presented here in a table to save you from having to examine every change in the diff. Before merging I will double check that this contains every API change and we can make the list prominent in a "Porting your code to v2.0" guide. I've also been considering a standalone script that can be run on a codebase to automatically convert all the function and parameter names to be compatible with v2.0, much like the Python 2to3 tool. That will be a separate PR or a completely separate thing though.

I'd appreciate it if anyone who uses or contributes to MoviePy could read through some of these suggestions and offer feedback about any and all changes. Lots of these are subjective so I'd like feedback to make sure that I'm not just doing what makes sense to me.

The example scripts should have been kept updated as I went (I used Pycharm's refactoring feature to do it), but if they no longer work, I am not particularly concerned because I'm expecting to update them all before v2.0 is released anyway. Many of them were already out-of-date and broken before this PR.

## Renamed class methods
Class|Old Name|New Name|Notes
|-----|-----|-----|-----|
Clip|fl|with\_filter
Clip |fl\_time|with\_time\_filter
Clip |set\_start|with\_start
Clip |set\_end|with\_end
Clip |set\_duration|with\_duration
Clip |set\_make\_frame|with\_make\_frame
Clip |set\_fps|with\_fps
Clip |set\_ismask|with\_is\_mask
Clip |set\_memoize|with\_memoize
VideoClip |fl\_image|with\_image\_filter
VideoClip|set\_audio|with\_audio
VideoClip|set\_mask|with\_mask
VideoClip|set\_opacity|with\_opacity
VideoClip|set\_position|with\_position
tools.py|cvsecs|convert_to_seconds
decorators.py|convert\_to\_seconds|convert\_parameters\_to\_seconds
credits.py|credits1|CreditsClip|from function to class
segmenting.py|findObjects|find\_object

## Not-yet-renamed class methods that I am undecided on - help needed!
Whilst the following renames are, I think, more consistent with the other renaming, they don't seem to work as well individually. Not sure.
| Class              | Old Name                | New Name?      |
|--------------------|-------------------------|----------------|
| Clip               | fl, fl\_time, fl\_image | These have been changed as above, but I think that there might be some better names possible. Perhaps avoiding the word filter? Unless filter is standard terminology for what these functions do?             |
| Clip               | fx                      | with\_fx       |
| Clip               | subclip                 | with\_subclip  |
| Clip               | cutout                  | with\_cutout   |
| VideoClip               | subfx                 | with\_subfx  |
| VideoClip               | afx                  | with\_afx   |
| moviepy\.video\.fx | speedx                  | change\_speed? |

## Renamed class attributes
| Class             | Old Name  | New Name    |
|-------------------|-----------|-------------|
| Clip              | ismask    | is\_mask    |
| TextClip          | txt       | text        |
| ImageSequenceClip | lastimage | last\_image |
| ImageSequenceClip | lastindex | last\_index |

## Renamed method parameters
| Method                                                                 | Old Name                             | New Name                                                     |                                                        |
|------------------------------------------------------------------------|--------------------------------------|--------------------------------------------------------------|--------------------------------------------------------|
| Clip\.with\_filter                                                     | fun                                  | func                                                         |                                                        |
| Clip\.fl\_time                                                         | ?t\_func                             | time\_func                                                   |                                                        |
| Clip\.set\_ismask \(now Clip\.set\_is\_mask\), VideoClip\.\_\_init\_\_ | ismask                               | is\_mask                                                     |                                                        |
| in VideoClip\.save\_frame, VideoClip\.write\_images\_sequence          | withmask                             | with\_mask                                                   |                                                        |
| VideoClip\.write\_images\_sequence                                     | nameformat                           | name\_format                                                 |                                                        |
| Clip\.subclip, SubtitlesClip\.in\_subclip                              | t\_start, t\_end                     | start\_time, end\_time                                       | Not sure about this one                                |
| in Clip\.cutout, VideoClip\.subfx                                      | ta, tb                               | start\_time, end\_time                                       | Could go to t\_start, t\_end if the above isn't kept   |
| VideoClip\.set\_opacity                                                | op                                   | opacity                                                      |                                                        |
| TextClip\.\_\_init\_\_                                                 | txt                                  | text                                                         |                                                        |
| FFMPEG\_VideoReader\.\_\_init\_\_                                      | pix\_fmt                             | pixel\_format                                                |                                                        |
| audio\.io\.preview                                                     | videoFlag, audioFlag                 | video\_flag, audio\_flag                                     |                                                        |
| audio\.tools\.find\_audio\_period                                      | aclip                                | clip                                                         |                                                        |
| audio\.tools\.find\_audio\_period                                      | t\_min, t\_max, t\_res               | min\_time, max\_time, time\_resolution                       | Not sure about this one                                |
| video\.fx\.blink                                                       | d\_on, d\_off                        | duration\_on, duration\_off                                  |                                                        |
| video\.fx\.headblur                                                    | r\_zone, r\_blur                     | radius, intensity                                            |                                                        |
| video\.fx\.lum\_contrast                                               | contrast\_thr                        | contrast\_threshold                                          |                                                        |
| video\.fx\.make\_loopable                                              | cross                                | overlap\_duration                                                | |
| video\.fx\.resize                                                      | newsize                              | new\_size                                                    |                                                        |
| video\.fx\.supersample                                                 | nframes                              | n\_frames                                                    |                                                        |
| audio\.fx\.audio\_loop                                                 | nloops                               | n\_loops                                                     |                                                        |
| video\.tools\.tracking\.manual\_tracking                               | nobjects                             | n\_objects                                                   |                                                        |
| CreditsClip, TextClip                                                  | fontsize                             | font\_size                                                   |                                                        |
| video\.tools\.cuts\.find\_video\_period                                | tmin                                 | start\_time                                                  |                                                        |
| tools\.FramesMatch, video\.io\.ffmpeg\_tools\.ffmpeg\_extract\_subclip | t1, t2                               | start\_time, end\_time                                       | Could also become t\_start, t\_end                     |
| tools\.FramesMatch                                                     | d\_min, d\_max                       | min\_distance, max\_distance                                 |                                                        |
| tools\.FramesMatches\.from\_clip                                       | dist\_thr                            | distance\_threshold                                          |                                                        |
| tools\.FramesMatches\.from\_clip                                       | max\_d                               | max\_duration                                                |                                                        |
| tools\.cuts\.FramesMatches\.select\_scenes                             | match\_thr, no\_match\_thr           | match\_threshold, no\_match\_threshold                       |                                                        |
| tools\.cuts\.detect\_scenes                                            | thr                                  | luminosity\_threshold                                        |                                                        |
| tools\.drawing\.color\_gradient                                        | r                                    | radius                                                       |                                                        |
| tools\.drawing\.color\_gradient, \.color\_split                        | col1, col2                           | color\_1, color\_2                                           |                                                        |
| tools\.drawing\.color\_split                                           | grad\_width                          | gradient\_width                                              |                                                        |
| tools\.drawing\.circle                                                 | col1, col2                           | color, bg\_color                                             |                                                        |
| tools\.segmenting\.find\_objects                                       | rem\_thr                             | size\_threshold                                              |                                                        |
| video\.io\.ffmpeg\_tools\.ffmpeg\_extract\_subclip                     | filename, targetname                 | inputfile, outputfile                                        |                                                        |
| video\.io\.ffmpeg\_tools\.ffmpeg\_merge\_video\_audio                  | video, audio, output, vcodec, acodec | videofile, audiofile, outputfile, video\_codec, audio\_codec |                                                        |
| video\.io\.ffmpeg\_tools\.ffmpeg\_extract\_audio                       | output                               | outputfile                                                   |                                                        |
| video\.io\.ffmpeg\_tools\.ffmpeg\_resize                               | video, output                        | inputfile, outputfile                                        |                                                        |

## Not-yet-renamed method parameters
Either because I couldn't decide work out whether it was an improvement, or because I couldn't work out what they actually represented (in which case they definitely need changing!)
| Method                                           | Old Name              | New Name                  |
|--------------------------------------------------|-----------------------|---------------------------|
| multiple                                         | nbytes, audio\_nbytes | n\_bytes, audio\_n\_bytes |
| multiple                                         | nchannels             | n\_channels               |
| video\.fx\.supersample                           | d                     | ???                       |
| AudioClip\.to\_soundarray, AudioClip\.get\_frame | tt                    | ???                       |
| video\.tools\.drawing                            | p1, p2                | pos1, pos2                |

## Internal variable changes
These are just a selection, not an exhaustive list, of the internal changes that I've made. These don't affect the external API.
| Old Name      | New Name              |
|---------------|-----------------------|
| f             | func                  |
| fun           | func                  |
| gf            | get\_frame            |
| tt            | timings               |
| c             | clip                  |
| l             | line                  |
| gf            | get\_frame            |
| xx, yy        | xs, ys                |
| lastread      | last\_read            |
| fileName      | file\_name            |
| videoFlag     | video\_flag           |
| audioFlag     | audio\_flag           |
| nVars         | n\_vars               |
| D             | duration              |
| newsize2      | exec\_new\_size\_func |
| transpo       | transpose             |
| a             | angle                 |
| angle         | get\_angle            |
| fdepr         | deprecated\_f         |
| \*a           | \*args                |
| \*\*k, \*\*kw | \*\*kwargs            |
| attr          | attribute\_string     |
| a             | attribute\_value      |
| new\_a        | new\_attribute\_value |

In general, I've left the following short variables because I think that they have sufficiently common uses so as not to cause confusion:
i (for iteration)
t (represents time)
w, h (represents width, height)
x, y (represents coordinates)
dx, dy (represents change in x, y)
im, img, pic (represents image)
ext (represents extension)
bg (represents background)
pos (represents position)

I've also not gone through and renamed variables inside the following because I didn't understand enough about how they worked to feel confident in guessing the correct full names of the abbreviations. It would be nice for someone who understands them to do that though.
video.tools.cuts.FramesMatches.from_clip()
video.tools.drawing
video.tools.segmenting
VideoClip.blit_on (also because it is being reworked in #1157)

## Removed functions
error_print (in tools.subprocess_call): never used, functionality covered by logger
sys_write_flush (in tools): never used, functionality now covered by logger
ffmpeg_tools.ffmpeg_movie_from_frames: previously deprecated, seems pretty much useless


jonnekaunisto:
This is a very good update for moviepy, but I am wondering why change set_* to with_*. In my opinion set is better, since it implies that you're setting that variable to something.

Other projects such as pygame use this kind of naming too as seen here: https://www.pygame.org/docs/ref/display.html

tburrows13:
Yes, 'getters' and 'setters' are common method names, with well-known meanings. The reason for the change is that MoviePy doesn't adhere to the current conventions, because they all use the `@outplace` decorator, which returns a _copy_ of the clip, not modifying the clip in place. This (rightfully imo) causes a lot of confusion with new users, so the change to `with_` is made to try and help that. (see #886 for an example).

I suppose the other option is to remove the `@outplace` decorators, and let all these methods modify the clip directly. However, the issue with that is that then you can't chain them together like `clip = VideoFileClip("example.mp4", etc, etc).set_duration(3).resize((300, 400)).set_layer(3)` and instead you would need to call each of these on a separate line. Also doing that would make it a lot harder to upgrade code to v2.0.

Hopefully this assessment makes sense?

This does raise the question, should we then apply the `with_` prefix to _all_ methods that don't act inplace such as `.fx`, `.subclip`, `cutout` and all the `fx` names like `resize`, `rotate`, `audio_normalize` etc? In particular renaming all the effect methods doesn't seem like a good idea.

jonnekaunisto:
Thank you for the clarification. Adding "with" does make sense to me now. 

Also about your last point about resize and rotate and others, I believe there is somewhat of a  convention in python where you make the function name into the past form when the function returns a copy.
An example in python is `sorted(list)` return a sorted copy of a list and `list.sort()` sorts the list itself. 

Maybe resize could be changed to `resized` and rotate to `rotated`?

Here is a discussion about that convention on stackoverflow: https://stackoverflow.com/questions/8211890/python-naming-conventions-for-functions-that-do-modify-the-object-or-return-a-mo

Zulko:
A bit late but just wanted to say awesome work! My two cents towards more explicit naming:


In the not-yet-renamed functions:

- `fl`, `fl_time`, `fl_image`: `transformation`, `time_transformation`, `image_transformation`.
- `fx`: `effect`.

In the not-renamed category:

- I would still replace `bg` by `background` and `pos` by `position` just to be consistent with the "no abbreviations in parameter names" policy. Same for w and h although these are more often used as "mathematical" variables so I don't feel very strongly about making these longer. 
- for the `d` in `supersample(d)` you could replace it with `time_delta` or just leave it to `d`. The doc string explains what it is about.

In the "not-yet-renamed" method parameters:

- `tt`: either `ts` or `times_array`

mxbi:
What about using abbreviations? Eg. `time_transformation` -> `time_transform` or even `time_tfm`. `transformation` is quite long.

As for `with_`, it makes sense if _some_ of the transforms are in-place and some are not, but don't they all return copies? It seems weird to rename all of them as I don't think this will cause that much confusion. (As long as people see an example where they can clearly see it returning new clips)

After all, I think this is "default" pythonic behaviour (see eg. PIL, numpy, pandas, ...). Perhaps we should be labelling operations that happen inplace instead.

Zulko:
@mxbi it is a matter of preference in the end, but abbreviations like `tfm` in functions and parameter names can be a problem for two reasons: they may not be transparent to some readers (they will be scratching their head as to what it means and what your code does), and they are more difficult to remember for developers ("was it `transform` or `trf` or `tfm` or `tf` ?"). My experience is that APIs which uses plain-english names generally make people happier (expect people who hate long lines :stuck_out_tongue: ). Of course ideally it would be plain english *and* as short as possible.

Regarding the motivation for using `with_start`, `with_fps` vs `set_start`, `set_fps`: even though the docs are clear on this point, developers in a hurry will just assume that the `set_` operations are in-place (as `set_` traditionally indicates). There must be 50 issues between Github and Stack Overflow saying *"I don't understand, I have a line saying `clip.set_duration(50)` but it has no effect on my clip"*. So while I'd be open to other alternatives to `with_`, I don't think v2.0 should continue with `set_`.

keikoro:
As I'm only becoming more active again now, I'm only reading through this now ^^ so idk if it still makes sense to make suggestions, however, I'd use `infile` and `outfile` instead of `inputfile` and `outputfile`. I think they are popular or well-known aliases across programming languages and would make the attributes shorter. But if we stick to the longer versions, I feel the individual words should be separated with an underscore (i.e. `input_file`).

tburrows13:
Thank you all for your input. The central question of inplace vs outplace: given that both numpy and pillow use outplace for things like crop and rotate, and the fact that that's already how moviepy works, there's no reason that we shouldn't continue doing the same.

Whilst I can't find any other examples of projects using `with_`, I'm still convinced that outplace `set_` is wrong, so I'm happy to go ahead with changing them to `set_`.



> * `fl`, `fl_time`, `fl_image`: `transformation`, `time_transformation`, `image_transformation`

Nice, although I think I prefer @mxbi's suggested `transform`, `time_transform`, `image_transform`.

I'm wondering if `fx` should become `apply_effect`?

I won't do the `subclip` -> `with_subclip` and `cutout` -> `with_cutout` that I suggested in the original post.

tburrows13:
New changes

- `fl`, `fl_time`, `fl_image`: `transform`, `time_transform`, `image_transform`
- `make_loopable`: `cross_duration` -> `overlap_duration`
- Lots more `pix_fmt` -> `pixel_format` (from #1237 - v2.0.0.dev2)

I'm going to merge this now, but I'll likely come back and make a few more minor changes later (and I can still respond to any feedback about these changes in another PR).

tburrows13:
Noting #1424 changes `volumex` to `audio_volume`

mondeja:
I should remove [this entry](https://github.com/Zulko/moviepy/pull/1424/commits/1ffb3366100d50e2b372ea5b7358e9af9ddf044d#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR16) from the CHANGELOG and let this issue track all changes in their CHANGELOG entry @tburrows13?

tburrows13:
Its fine for now. Before the next release I'll write up a proper explanation of how to migrate to v2.0, including which things have been renamed, and how to fix them. I can mess around with the changelog then if need be.

Zulko:
Sorry for chiming in late on this but does audio_volume multiply the volume
the same way volumex used to do? The name sounds like an attribute, and
audio_volume(2) sounds like you are setting the volume to a level of 2,
while actually you are multiplying it by 2. What about an API like
multiply_volume() or change_volume(factor=1.2) ?

mondeja:
> What about an API like multiply_volume() or change_volume(factor=1.2)?

I agree and my vote would be for something like `change_volume(factor=1.2)`. But to maintain consistency with other audio FXs should be `audio_change_volume` and `audio_change_stereo_volume`. What do you think @tburrows13?

tburrows13:
> Sorry for chiming in late on this but does audio_volume multiply the volume the same way volumex used to do? The name sounds like an attribute, and audio_volume(2) sounds like you are setting the volume to a level of 2, while actually you are multiplying it by 2. What about an API like multiply_volume() or change_volume(factor=1.2) ?

That's a valid point. I'd support `audio_multiply_volume` and `audio_multiply_stereo_volume`. If that is too unwieldy and long I think that we could drop the `audio` prefix here as well. It is only needed on the other audio effects because fadein, fadeout, loop and normalise are not explicitly audio features.

mondeja:
Okay, I will go with `multiply_volume` and `multiply_stereo_volume` :+1:

PR #1172, opened 2020-05-01
MacOS CI test

jonnekaunisto:
There were some issues with Python and Matplotlib, so the test uses Python Framework build from conda. 

See here: https://matplotlib.org/3.1.0/faq/osx_framework.html

Conda doesn't support versioning for the python framework build so the test only runs on 3.7.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -1,0 +1,75 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  # Uses Python Framework build because on macos matplotlib requires it
+  macos:
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        python-version: [3.7]
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,75 @@
...
+        # W293: blank line contains whitespace
+        # W503: line break before binary operator
+        flake8 moviepy --ignore=E203,E302,E402,W291,W293,W503 --max-complexity=10 --max-line-length=127 --statistics --count --exit-zero
+        # Check examples and tests with slightly relaxed rules to allow 'star' imports
+        # F403 'from moviepy.editor import *' used; unable to detect undefined names
+        # F405 'name' may be undefined, or defined from star imports: moviepy.editor
+        flake8 . --ignore=E203,E302,E402,W291,W293,W503,F403,F405 --exclude moviepy/ --max-complexity=10 --max-line-length=127 --statistics --count  --exit-zero
+    - name: Test with pytest
+      run: |
+        pythonw -m pytest tests/ --doctest-modules -v --cov moviepy --cov-report term-missing
```
tburrows13, line 75: Maybe we could get rid of everything after the -v? There doesn't seem much point checking for code coverage and just printing it to the console. Presumably the linux test will still call coveralls, but this one doesn't need to check coverage.

tburrows13:
Thank you very much for making this, and thanks for your varied contributions to the project over the last few weeks. It has all been very helpful!

Issue #1145, opened 2020-04-14
Some advice for improve CompositeVideoClip TextClip and the blit


ODtian:
Since I was been stumped with the perfermance of my edit, I try to figure out the reason that slow down the program, and luckly I found couple of them.
The first one is CompositeVideoClip, its blit all clip for every frame,so the speed is decreasing as the number of clips increased, speed comes 1s/it when there are 20 clips needs to be composited, one solution is using PIL.Image.paste, it much faster, create a bg image object and paste other clip on it, it's really fast,came up to 30it/s
second one is blit func, I think PIL Image.paste is useful too.
And last one is Textclip, I use PIL imagefont to gen text image, Its much faster than imagemagisk,maybe it is possible to escape imagemagisk.

tburrows13:
Thank you for that investigation. Certainly I'd be up for depending more on pillow (and getting rid of/reducing need for imagemagick. There's also https://gist.github.com/Zulko/e072d78dd5dbd2458f34d2166265e081#file-text_clip_with_gizeh-py that I found, that may be even faster for text.

ODtian:
thanks for your share!actully I make a simple grid layout manager to design text animation, with single char,and that's why I need fast text gen speed.I love this and I will keep working on it.

Zulko:
For context, off the top of my head:

- ImageMagick:
  - Advantage: nice text formatting, has access to the system fonts. Just historically my first choice.
  - inconvenients: difficult to install, slow.
- Pillow:
  - Advantages: easy to install, fast
  - inconvenients: doesnt have access to system fonts, you need to provide a path to a ttf (?)
- Gizeh/Cairo
  - Advantages: fast, access to system fonts
  - Cairo not streightforward to install on some systems.

ODtian:
I use a day to rewrite the CompositiveVideoClip and it's fruitful.

In `moviepy.video.compositing.CompositeVideoClip.CompositeVideoClip` change the `make_frame` func:
```
        def make_frame(t):
            full_w, full_h = self.bg.size
            f = self.bg.get_frame(t).astype('uint8')
            bg_im = Image.fromarray(f)
            for c in self.playing_clips(t):
                img, pos, mask, ismask = c.new_blit_on(t, f)

                x, y = pos
                w, h = c.size

                out_x = x < -w or x == full_w
                out_y = y < -h or y == full_h

                if out_x and out_y:
                    continue

                pos = (int(round(min(max(-w, x), full_w))),
[...]

Issue #855, opened 2018-10-08
Even after commenting out that one line in /etc/ImageMagick-6/policy.xml, I still get "that" error.


neatville:
It's more or less the same issue as #693. The line of code that triggers it:  
`txt_clip = TextClip("example", fontsize=50,color='white')`  
Error summary:
```
OSError: convert-im6.q16: not authorized `@/tmp/tmpxnbqj8cq.txt' @ error/property.c/InterpretImageProperties/3516.
convert-im6.q16: no images defined `PNG32:/tmp/tmpblh_ia5s.png' @ error/convert.c/ConvertImageCommand/3258.
```
My system is a ubuntu derivative. I installed ImageMagick from the Ubuntu repositories with apt-get. It made me wonder if any cache would need to get cleared somehow, but I still got this error after a system reboot.  
If this makes any difference, `print(get_setting("FFMPEG_BINARY")` gives me this: `/home/me/.imageio/ffmpeg/ffmpeg-linux64-v3.3.1`, and `which convert` returned a positive result.

Rub21:
I am using moviepy and  ImageMagick  in docker, I get the same error after commenting the line

keikoro:
Closing this issue as it's rather old and missing info we require now via the issue template.

The docker issue was apparently addresssed in #1165 (in case it persists, please open a new issue, thanks).

PR #1175, opened 2020-05-03
Fix Dockerfile

rikhuijzer:
The Dockerfile was broken. The error stated that the package `libav-tools` was unavailable. 
To fix this I have fixed the Python version to `python:3.7-stretch` since `python:3.7-buster` does not contain `libav-tools`. See also the [Debian package index](https://packages.debian.org/search?keywords=libav-tools).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3
+FROM python:3.7-stretch
 
 # Install numpy using system package manager
 RUN apt-get -y update && apt-get -y install libav-tools imagemagick libopencv-dev python-opencv
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM python:3.7-stretch
+FROM python:3
 
 # Install numpy using system package manager
-RUN apt-get -y update && apt-get -y install libav-tools imagemagick libopencv-dev python-opencv
+RUN apt-get -y update && apt-get -y install ffmpeg imagemagick libopencv-dev python-opencv
 
 # Install some special fonts we use in testing, etc..
 RUN apt-get -y install fonts-liberation


tburrows13:
I wonder if it works if you remove the installation of `libav-tools`? It seems to be a compatibility layer for ffmpeg, so perhaps we could instead get rid of it entirely? I think that it would be better to remove it than to pin it down to an older version of Debian?

@jonnekaunisto you had some experience with the Docker build, did you have any suggestions?

jonnekaunisto:
The error that is thrown by the current Dockerfile is:

`Package libav-tools is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  ffmpeg`

So I believe you can just replace it with ffmpeg instead.

rikhuijzer:
Thanks @jonnekaunisto changing to `ffmpeg` is indeed a much better idea. I just did that and rebuild the Docker image. That works. However, I continued by trying to run some code which depends on `ImageMagick` and got `unable to read font [...]`. This is starting to become quite frustrating since I just spend my entire afternoon on a fruitless global MoviePy installation with ImageMagick on NixOS. So, I think this Dockerfile fix will be better than no fix, but it does not seem to deliver a version with a functioning ImageMagick attached.

rikhuijzer:
This ImageMagick problem might be related to the fairly new comment at https://github.com/Zulko/moviepy/issues/855.

tburrows13:
In which case, I'm closing this in favour of #1165, which includes this change. I'm not sure about your ImageMagick problem... that program can be quite a pain at times. It is currently only used in TextClips and writing gifs. Gif writing already has alternatives that use different libraries (like pillow and ffmpeg directly), and it is likely that there will be a pillow alternative to TextClip in v2.0 (see #1145), which will heavily reduce the need for ImageMagick.

rikhuijzer:
Sounds good. I now solved my ImageMagick problem on NixOS by switching my Python code to a font listed at `convert -list font`. I do still suspect that the default font for TextClip is `Courier` which causes problems in all my Docker builds (tried many Debian distributions).

tburrows13:
fyi `TextClip.list("font")` also lists all fonts available to ImageMagick (all it is is a wrapper around `convert -list font`.

PR #1188, opened 2020-05-10
Update CI ImageMagick version

tburrows13:
People will be using ImageMagick v7 with moviepy, so we should test against it.


mondeja:
This change is doing a lot of different things unrelated to ImageMagick version update. Could you consider separate them in another pull request?

tburrows13:
Oops it seems to have got mixed up with #1176. Probably a faulty rebase or something. I'll sort it out (but this PR probably won't be needed once I've finished the new text clip and imagemagick is gone).

mondeja:
> but this PR probably won't be needed once I've finished the new text clip and imagemagick is gone

If so, could you check it and close if not needed?

tburrows13:
Yeah, this PR is completely messed up. I'll close on the assumption that it will be not needed in the future.

PR #1201, opened 2020-05-24
Prevent exception when filename is given.

obiwankennedy:
just a short fix, I have issue about textclip, so I fixed it for me, and make the fix available here. Perhaps, it is a logic error. I did not check that. but at least you know that there is something strange around this code when passing a filename argument.

My textCilp construction is: 
`txtclip = TextClip(filename="/home/user/path/to/where/i/want/{}".format(text),font="Inria-Sans",fontsize=150, color="white")`


tburrows13:
Thanks for the fix. I'll get round to checking this properly soon.

tburrows13:
Agreed, this is a logic error: If `txt == None` and `temptxt` is not provided, then it will crash (as you found). I can't work out why `tempfilename` would end up being `None`, but I think that it is a good idea to clean up regardless of logic earlier in the function.

tburrows13:
I'll merge once we've sorted out the appveyor issue in #1207.

PR #1281, opened 2020-08-01
Correct CompositeVideoClip docstring (swap width, height ordering)

btRooke:
Simple fix to correct the description of `CompositeVideoClip`'s size parameter.


tburrows13:
Thanks for spotting that! I think that I've now fixed the test error, so I'll merge this now.

PR #1296, opened 2020-08-12
fix to terminate the process that has ended

luyanger1799:
check if the process exists before using teminate() to stop the process.

- [ ] I have formatted my code using `black -t py36`

### DIFF
--- a/moviepy/audio/io/readers.py
+++ b/moviepy/audio/io/readers.py
@@ -249,10 +249,11 @@ class FFMPEG_AudioReader:
 
     def close(self):
         if self.proc:
-            self.proc.terminate()
-            self.proc.stdout.close()
-            self.proc.stderr.close()
-            self.proc.wait()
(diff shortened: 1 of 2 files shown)


tburrows13:
Hello, thanks for the PR. I'm not quite sure why this is necessary? Does calling terminate crash if the process has already terminated?

mondeja (approved):
I'm seeing this legit :+1: Thank you for the improvement @luyanger1799 :pray:

If the process has been stopped by another program or the user want to interact directly with the project, this becomes useful.

> Does calling terminate crash if the process has already terminated?

Nope

tburrows13:
I don't really understand, so I'll go with your judgement on this. Presumably the same thing should be added to https://github.com/Zulko/moviepy/blob/master/moviepy/video/io/ffmpeg_reader.py#L228 then?

You can commit directly to this PR if you need to, so that it is done as one PR.

PR #1304, opened 2020-08-24
Update dockerfile to modify policy.xml in-place

TrippleCCC:

The last line in the Dockerfile currently removes all content from the policy.xml file since the file cannot be read from and simultaneously written too. This fix edits the file in place using sed -i.

This bug can be reproduced by building the docker image and then running docker run (tagname) cat /etc/ImageMagick-6/policy.xml


tburrows13:
Thanks for spotting this. I'll merge it now, although I'm still hoping that we can get rid of ImageMagick entirely soon!

PR #1328, opened 2020-10-04
Add support for ReadTheDocs

tburrows13:
I've been planning documentation improvements for a while, and I'd like to switch to using readthedocs to host them. This means that any changes made to master will be instantly reflected in the documentation with no extra effort required from maintainers, and it comes with easy options for users to switch to an older snapshot of the docs. I'll also merge this change into the v1.0.x branch so that will always be available in its current form.

The URL will be moviepy.readthedocs.io, but I need @Zulko to enable the githook to allow auto-creation of the docs. Would you mind linking your account to https://readthedocs.org? Then I can add you as a maintainer to the moviepy readthedocs project and then you can link it to this GitHub repo.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,23 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
+# Build documentation with MkDocs
+#mkdocs:
+#  configuration: mkdocs.yml
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.7
+  install:
+      - method: pip
+        path: .
+        extra_requirements:
+           - doc
--- a/setup.py
+++ b/setup.py
@@ -88,9 +88,7 @@ optional_reqs = [
 
 doc_reqs = [
     "pygame>=1.9.3,<2.0; python_version<'3.8'",
-    "numpydoc",
-    "sphinx_rtd_theme>=0.1.10b0,<1.0",
-    "Sphinx>=1.5.2,<2.0",
+    "numpydoc<2.0",
 ]
 
 test_reqs = [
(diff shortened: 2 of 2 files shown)


Zulko:
@tburrows13  registered to ReadTheDocs (under username Zulko), I can see the moviepy project there, with an August build failing because of a numpydoc requirement (which this PR solves if I understand correctly). Let me know what the next steps are.

tburrows13:
Great. I've added you as a 'maintainer' on readthedocs (though perhaps this wasn't needed because it knows that you own the associated github repo. Can you please go to "Admin" > "Integrations" > "Add Integration" > Type="Github incoming webhook". That should be all that is required, and I can adjust the settings from there. (It should still fail the first build because this PR hasn't yet been merged, but it will then rebuild on every commit)

Zulko:
Done :+1:

tburrows13:
That looks like it worked, thanks!

tburrows13:
I definitely plan to play around with the theming/style a bit more (on top of the many text improvements needed), but it is now viewable at [moviepy.readthedocs.io](moviepy.readthedocs.io), and you can switch to the 1.0.x branch at the bottom. I'm not in a rush to transition from the current docs.

Zulko:
That's great but I noticed that the Moviepy docs now have inserted ads, which is sad. Apparently I could pay 60$/year to remove them, but that's a bummer. Is readthedocs worth it? Would there be a way to get similar auto-deployment features using github actions?

tburrows13:
Yeah, the ads are unfortunate. The readthedocs autobuilding will be very useful for quick feedback whilst I rework the docs, but yes, there’s actually no reason why we shouldn’t then try autobuilding into gh-pages later. I don’t think we’d get the multiple doc instance selector which is pretty neat, but perhaps on release of v2.0 we could then keep the gh pages up-to-date with the latest master automatically and link to a readthedocs instance for people who still need legacy?

keikoro:
Adding the correct link here (the one above is wonky): https://moviepy.readthedocs.io/en/latest/

keikoro:
Btw. IIRC, GitHub's now made it possible to use other branches than `gh-pages` for pages (for hosting docs).

keikoro:
Yep. Source: https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

Zulko:
I like the long-term solution of having rtd only for legacy versions. In the meantime, whathever makes your life easier during the v2 transition :+1:

Issue #1278, opened 2020-07-29
Transpose axes not matching error when rotating by 90 or -90 degrees


OneWorld-github:
```
import moviepy.editor as mped
import numpy as np
colour_clip = mped.ColorClip(size=[300, 200], color=np.array([250, 90, 0]).astype(np.uint8), duration=3).set_position((0, 0))
rot_trans_col_clip = colour_clip.add_mask().rotate(90, unit='deg', expand=True, resample='bilinear')
stacked_clips = mped.CompositeVideoClip([rot_trans_col_clip]).set_duration(3)
stacked_clips.write_videofile('90_degrees_in_moviepy.mp4', fps=5)
```

#### Expected Behavior

a 90 degree rotated asset

#### Actual Behavior

```
Traceback (most recent call last):
  File "c:/..../pillow-rotation.py", line 16, in <module>
    rot_trans_col_clip = colour_clip.add_mask().rotate(90, unit='deg', expand=True, resample='bilinear')
  File "C:\....\Python38\lib\site-packages\moviepy\video\fx\rotate.py", line 71, in rotate
    return clip.fl(fl, apply_to=["mask"])
  File "C:\....\Python38\lib\site-packages\moviepy\video\VideoClip.py", line 943, in fl
    newclip = VideoClip.fl(self, fl, apply_to=apply_to,
  File "C:\....\Python38\lib\site-packages\moviepy\Clip.py", line 151, in fl
    new_a = a.fl(fun, keep_duration=keep_duration)
  File "C:\....\Python38\lib\site-packages\moviepy\video\VideoClip.py", line 943, in fl
    newclip = VideoClip.fl(self, fl, apply_to=apply_to,
  File "C:\....\Python38\lib\site-packages\moviepy\Clip.py", line 138, in fl
    newclip = self.set_make_frame(lambda t: fun(self.get_frame, t))
  File "<decorator-gen-57>", line 2, in set_make_frame
  File "C:\....\Python38\lib\site-packages\moviepy\decorators.py", line 14, in outplace
    f(newclip, *a, **k)
  File "C:\....\Python38\lib\site-packages\moviepy\video\VideoClip.py", line 666, in set_make_frame
    self.size = self.get_frame(0).shape[:2][::-1]
  File "<decorator-gen-10>", line 2, in get_frame
  File "C:\....\Python38\lib\site-packages\moviepy\decorators.py", line 89, in wrapper
    return f(*new_a, **new_kw)
[...]

Issue #1126, opened 2020-04-01
Clip rotation does not work for RGBA images in 90 degrees steps


GrigoreAlexandru:
### Expected Behavior
clip rotates

### Actual Behavior
ValueError: axes don't match array

### Steps to Reproduce the Problem

```
img = PIL.Image.new("RGBA",(10,10))
array = np.array(img)
clip = ImageClip(array).add_mask().rotate(90)
```

### Specifications

  - Python Version: 3.7
  - Moviepy Version: 1.0.2
  - Platform Name: Windows
  - Platform Version: 10

### More info

from rotate.py

```
transpo = [1,0] if clip.ismask else [1,0,2]

....        
im = gf(t)
....
        if (a==90) and expand:
            return np.transpose(im, axes=transpo)[::-1]
```
This happens because gf() returns the frame 2 times, once with the RGB matrix and the other one with the alpha channel matrix.

So to fix this you would need a way to find out what matrix you're getting and pass the proper axes to transpose. This is done with the
```
transpo = [1,0] if clip.ismask else [1,0,2]
```
but it's weird since the clip is constant and so is .ismask, furthermore, from my findings, ismask returns false even when using clip.add_mask(), not sure what gives.

One other way is to use rot90 from numpy instead of transpose which eliminates the need to differentiate between the matrices.

ODtian:
hi, I use fl() method to make a opacity animation, but I got an error occured, I use gf func to get numpy array and use pillow to set a opacity to it then convert to nparray to return, is there any thing need to be noticed for the rgba array?

GrigoreAlexandru:
> hi, I use fl() method to make a opacity animation, but I got an error occured, I use gf func to get numpy array and use pillow to set a opacity to it then convert to nparray to return, is there any thing need to be noticed for the rgba array?

What error are you getting ?
Debug step by step whatever gf() returns, check the matrix size on each step. Most likely you're getting first a RGB matrix then on the second return you get the alpha matrix. 2 solutions on top of my head are to either merge the 2 matrices and use pillow to modify the alpha or just work with the alpha matrix alone.

You can also animate clips by returning a closure to set_opacity().

tburrows13:
Fixed in #1335

Issue #1042, opened 2019-11-28
CompositeVideoClip([xxx]).rotate(90)  ValueError: axes don't match array


clotyxf:
### Specifications

- Python Version: 3.6.3
- Moviepy Version: 1.0.1
- Platform Name: Centos
- Platform Version: 7.2

### Actual Behavior

```
size = (640, 320)
fps = 24
bg_clip = mpy.ImageClip(img=image, fromalpha=True)
bg_clip = bg_clip.resize(size).set_duration(d).set_fps(fps )
bg_clip = mpy.CompositeVideoClip([bg_clip])
bg_clip.rotate(90)
```

ValueError: axes don't match array
```
change the angle to 89.99, is worked.
```

riohay:
bg_clip.rotate(90, expand=False) would give what you want.

misland:
'''
        if (a==90) and expand:
            return np.transpose(im, axes=transpo)[::-1]
        elif (a==-90) and expand:
            return np.transpose(im, axes=transpo)[:,::-1]
        elif (a in [180, -180]) and expand:
            return im[::-1,::-1]
'''
code in moviepy->video->fx->rotate.py,line 58_63,that's why 89.99 worked.
@clotyxf could you reopen this issue?89.99 shouldn't be the workaround
@tburrows13 hi,could you explain why,many thanks :)

misland:
> bg_clip.rotate(90, expand=False) would give what you want.

image will lose part of it when expand is set to true,so this should be the workaround

misland:
I just copy the code in line 69 to line 59
```
if (a==90) and expand:
            return pil_rotater(im, a, resample=resample, expand=expand)
            # return np.transpose(im, axes=transpo)[::-1]
```
then it works fine,I don't know why we do this specialized processing when the angle is set to `90`

tburrows13:
Hi, it definitely seems like it is a bug to do with RGBA images. See #781 and #990 for the same error message.

The reason why angles 90, -90, 180 and -180 are special is because they can be easily done by directly transposing the image array using numpy. More complicated rotations require pillow, which is an optional library, and may perform the rotations slower. However, if this bug cannot properly be fixed, then it might be worth using the pillow method as a fall back.

tburrows13:
Fixed in #1335

Issue #990, opened 2019-07-27
Excepcion happens during TextClip transposing


dvorotnev:
### Actual Behavior
<details>
<summary>Exception traceback </summary>

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-5db7c433b1f3> in <module>
----> 1 text_video = text_video.rotate(90)

~/.local/lib/python3.6/site-packages/moviepy/video/fx/rotate.py in rotate(clip, angle, unit, resample, expand)
     69             return pil_rotater(im, a, resample=resample, expand=expand)
     70 
---> 71     return clip.fl(fl, apply_to=["mask"])

~/.local/lib/python3.6/site-packages/moviepy/video/VideoClip.py in fl(self, fl, apply_to, keep_duration)
    942         # Therefore the result is not an ImageClip, just a VideoClip.
    943         newclip = VideoClip.fl(self, fl, apply_to=apply_to,
--> 944                                keep_duration=keep_duration)
    945         newclip.__class__ = VideoClip
    946         return newclip

~/.local/lib/python3.6/site-packages/moviepy/Clip.py in fl(self, fun, apply_to, keep_duration)
    149                 a = getattr(newclip, attr)
    150                 if a is not None:
--> 151                     new_a = a.fl(fun, keep_duration=keep_duration)
    152                     setattr(newclip, attr, new_a)
    153 

~/.local/lib/python3.6/site-packages/moviepy/video/VideoClip.py in fl(self, fl, apply_to, keep_duration)
    942         # Therefore the result is not an ImageClip, just a VideoClip.
    943         newclip = VideoClip.fl(self, fl, apply_to=apply_to,
--> 944                                keep_duration=keep_duration)
    945         newclip.__class__ = VideoClip
    946         return newclip

~/.local/lib/python3.6/site-packages/moviepy/Clip.py in fl(self, fun, apply_to, keep_duration)
    136 
    137         #mf = copy(self.make_frame)
--> 138         newclip = self.set_make_frame(lambda t: fun(self.get_frame, t))
    139 
    140         if not keep_duration:

</home/pacman/.local/lib/python3.6/site-packages/decorator.py:decorator-gen-172> in set_make_frame(self, mf)

~/.local/lib/python3.6/site-packages/moviepy/decorators.py in outplace(f, clip, *a, **k)
     12     """ Applies f(clip.copy(), *a, **k) and returns clip.copy()"""
     13     newclip = clip.copy()
[...]

PR #1335, opened 2020-10-07
Fix 90 degrees rotation

### LINKED ISSUE #781: rotating ImageClip with mask sets wrong transpo
The transpo is set for the whole clip as [1, 0, 2]. Then when it comes to mask it going straight to rotate fl with the same transpo and falling.

Also, unlike the mirror functions, rotate() doesn't accept apply_to argument, but force applying to mask. If it was, I could disable the apply_to mask, then rotate the mask manually and reapply the mask to image.

nicoeps:
Resolves #781 #990 #1042 #1126 #1278 

Rotating clips by either 90 or 270 degrees throws an exception because axes don't match when transposing. The clips seem to contain a frame (particularly the second frame of the clip) that has two dimensions instead of three. These changes should fix this issue by checking the shape for each frame and setting the appropriate value for transposing.

```python
# Code that demonstrates the bug
import moviepy.editor as mp
text_video = mp.TextClip("test", color="blue", fontsize=72)
text_video = text_video.rotate(90).set_duration(2)
text_video.write_videofile("test.mp4", fps=24)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed <!-- for any bug fixes -->
 - Fixed BitmapClip with fps != 1 not returning the correct frames or crashing [#1333]
+- Fixed `rotate` sometimes failing with `ValueError: axes don't match array` [#1335]
 
 
 ## [v2.0.0.dev2](https://github.com/zulko/moviepy/tree/v2.0.0.dev2)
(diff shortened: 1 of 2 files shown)


tburrows13:
This looks great, thank you so much! You can ignore the failed code format check, that's not your fault (unless you want to work out why it has stopped working for everybody!).

Presumably you've checked that this fix works on videos that failed before?

tburrows13:
There was a merge conflict because I had just merged #1170 which I've now fixed.

PR #1340, opened 2020-10-09
Move most imports from editor.py to __init__.py

tburrows13:
This PR moves most of the importing to `__init__.py`. This means that most users can import everything that they need directly from `moviepy` instead of from `moviepy.editor` or `moviepy.video.io.VideoFileClip`... you get the picture :)

I've kept `editor.py` to handle the stuff that the user runs interactively: Pygame, matplotlib and IPython.

This change is also fully backwards compatible because in `editor.py`, I call `from . import *` so anything that is now in `__init__.py` is also available from `editor.py`.

I've included an example change to `test_fx.py` that demonstrates the change. This could be applied to most of the test suite imports. I've not yet updated the documentation but in general it will advise things like `from moviepy import VideoFileClip` or `from moviepy import *`.

Let me know your thoughts and feedback.

Todo:
- [x] Add changelog entry
- [x] Update documentation


moviepy/__init__.py
```diff
@@ -1 +1,57 @@
...
+
+# FX
+from .video import fx as vfx
+import moviepy.audio.fx as afx
+import moviepy.video.compositing.transitions as transfx
+
+# Tools
+import moviepy.video.tools as videotools
+import moviepy.video.io.ffmpeg_tools as ffmpeg_tools
+from .tools import convert_to_seconds
```
mondeja, line 36: Some imports are relative (`from .foo.bar`) and others absolute (`import moviepy.foo.bar`). Why not use absolute imports for all?

tburrows13: That's just the way it was in editor.py. Thanks for noticing, I'll change it

tburrows13:
I think that this is ready to merge

moviepy/video/io/html_tools.py
```diff
@@ -73,7 +73,7 @@ def html_embed(
 
     Examples
     =========
-
+    TODO Update example
```
mondeja, line 76: Why not to place here "TODO Create example" removing the example and update the example of `ipython_display` function of the same module with:

```python
>>> from moviepy.editor as *
    >>> # later ...
    >>> clip.ipython_display(width=360)
    >>> clip.audio.ipython_display()

    >>> clip.write_gif("test.gif")
    >>> ipython_display('test.gif')

    >>> clip.save_frame("first_frame.jpeg")
    >>> ipython_display("first_frame.jpeg")
```

moviepy/__init__.py
```diff
@@ -1 +1,82 @@
...
+
+def show(self, *args, **kwargs):
+    """NOT AVAILABLE: clip.show requires importing from moviepy.editor"""
+    raise ImportError("clip.show requires importing from moviepy.editor")
+
+
+VideoClip.preview = preview
+VideoClip.show = show
+AudioClip.preview = preview
+
```
mondeja, line 79: Could you define here public exports in `__all__` variable?

moviepy/editor.py
```diff
@@ -1,133 +1,63 @@
...
->>> clip.resize(2)
-
-It also starts a PyGame session (if PyGame is installed) and enables
-clip.preview().
+- Starts a pygame session to enable ``clip.show()`` and ``clip.preview()`` if pygame is installed
+- Enables ``clip.ipython_display()`` if in an IPython Notebook
+- Allows the use of ``sliders`` if Matplotlib is installed
 """
 
-__all__ = [
```
mondeja, line 17: Could you maintain `__all__` here, removing those imports that now are not in the file?

mondeja (changes requested):

moviepy/__init__.py
tburrows13: Wouldn't that just be a list of all imported names except that ones that are deleted below? Is there a good reason for doing that?

mondeja: If you do `import *` for a module that has `__all__` variable, only that objects described in the iterable will be exported. In fact, the addition of would deprecate #1419.

mondeja (approved):
Great! Thank you @tburrows13

tburrows13:
Honestly probably my favourite change in 2.0 👍🏼

PR #1342, opened 2020-10-09
Optimise clip compositing using PIL (replacement for #1157)

tburrows13:
I managed to break #1157 whilst fixing merge conflicts, so here is the code changes from that PR plus my little change to fix the failing test (when the clip had a mask).


moviepy/video/compositing/CompositeVideoClip.py
```diff
@@ -116,16 +116,21 @@ def __init__(
...
+        im = Image.fromarray(frame)
 
-            frame = self.bg.get_frame(t)
-            for clip in self.playing_clips(t):
-                frame = clip.blit_on(frame, t)
-            return frame
+        if self.bg.mask is not None:
+            frame_mask = self.bg.mask.get_frame(t)
+            im_mask = Image.fromarray(255 * frame_mask).convert("L")
+            im = im.putalpha(im_mask)
```
TheSpaceSheep, line 128: There is a bug here : im.putalpha(im_mask) works in-place and returns None, so it should not be assigned back to im.

PR #1349, opened 2020-10-12
Format files

keikoro:
Renames back `Pipfile.lock` (result of [this conversation](https://github.com/Zulko/moviepy/commit/c96280849b39357b9361e6012d23c3f367243bc9)) and improves wording in `setup.py`.


tburrows13 (approved):

keikoro:
Btw. I was wondering why we don't include `black` in any of the package lists.

Tbf, I wouldn't know where to put it right now and haven't looked deep enough into the extra requirements to know if they can be combined/can reference each other, in which case a "dev" mode (black + test reqs + docs reqs) might make sense.

tburrows13:
> Btw. I was wondering why we don't include `black` in any of the package lists.
> 
> Tbf, I wouldn't know where to put it right now and haven't looked deep enough into the extra requirements to know if they can be combined/can reference each other, in which case a "dev" mode (black + test reqs + docs reqs) might make sense.

Yeah, it isn't included because it is not needed for running moviepy, running the tests, or building the docs. A 'dev' or 'all' list should probably include it. No idea if that is possible/recommended.

Issue #1253, opened 2020-06-30
time_mirror output appeared OSError when access time is negative


LaoYuanPython:
```
if __name__=='__main__':
    clip = VideoFileClip(r"F:\video\WinBasedWorkHard_src.mp4", audio=True).fx(vfx.crop,0, 300, 540, 840)
    newclip = clip.fx(vfx.time_mirror)
    newclip.write_videofile(r"F:\video\WinBasedWorkHard_time_mirror.mp4", threads=8)
```

#### Expected Behavior
Output file success

#### Actual Behavior
```
Moviepy - Building video F:\video\WinBasedWorkHard_time_mirror.mp4.
MoviePy - Writing audio in WinBasedWorkHard_time_mirrorTEMP_MPY_wvf_snd.mp3
chunk:  98%|█████████▊| 1178/1208 [00:21<00:00, 49.13it/s, now=None]Traceback (most recent call last):
  File "F:/study/python/project/moviepyTest/moviepyTest.py", line 7, in <module>
    newclip.write_videofile(r"F:\video\WinBasedWorkHard_time_mirror.mp4", threads=8)
  File "<decorator-gen-61>", line 2, in write_videofile
  File "C:\Program Files\Python37\lib\site-packages\moviepy\decorators.py", line 56, in requires_duration
    return f(clip, *a, **k)
  File "<decorator-gen-60>", line 2, in write_videofile
  File "C:\Program Files\Python37\lib\site-packages\moviepy\decorators.py", line 135, in use_clip_fps_by_default
    return f(clip, *new_a, **new_kw)
  File "<decorator-gen-59>", line 2, in write_videofile
  File "C:\Program Files\Python37\lib\site-packages\moviepy\decorators.py", line 24, in convert_masks_to_RGB
    return f(clip, *a, **k)
  File "<decorator-gen-58>", line 2, in write_videofile
  File "C:\Program Files\Python37\lib\site-packages\moviepy\decorators.py", line 89, in wrapper
    return f(*new_a, **new_kw)
  File "C:\Program Files\Python37\lib\site-packages\moviepy\video\VideoClip.py", line 337, in write_videofile
    logger=logger,
  File "<decorator-gen-47>", line 2, in write_audiofile
  File "C:\Program Files\Python37\lib\site-packages\moviepy\decorators.py", line 56, in requires_duration
    return f(clip, *a, **k)
  File "<decorator-gen-46>", line 2, in write_audiofile
[...]

Issue #1227, opened 2020-06-12
Cutout removes from the audio twice


spollard:
#### Expected Behavior
 Calling ```newclip = clip.cutout(start_remove,end_remove)``` should result in a clip with a ```newclip.audio.duration``` equal to the ```newclip.duration``` and a ```newclip.audio.end``` equal to the ```newclip.audio.end```. 

#### Actual Behavior
Calling ```newclip = clip.cutout(start_remove,end_remove)``` correctly removes (end_remove-start_remove) seconds from the video, but it removes 2 * (end_remove-start_remove) from the audio. 

#### Steps to Reproduce the Problem

```
from moviepy.editor import VideoFileClip

clip = VideoFileClip(PATH_TO_VIDEO) # insert path to your video here
print(f"Clip duration {clip.duration} and end {clip.end}")
print(f"clip audio duration {clip.audio.duration} and end {clip.audio.end}")
newclip = clip.cutout(1,5)
print(f"Newclip duration {newclip.duration} and end {newclip.end}")
print(f"Newclip audio duration {newclip.audio.duration} and end {newclip.audio.end}")
```

#### Specifications

  - Python Version: 3.6.9
  - Moviepy Version: 1.0.0-1.0.3
  - Platform Name: Linux/Ubuntu
  - Platform Version: 18.04

#### Specific issue

Clip.cutout() in Clip.py is decorated by apply_to_audio from decorators.py, and so it is called later on clip.audio. Clip.cutout() calls newclip.set_duration while Clip.set_duration also is decorated by apply_to_audio, and so it is called on clip.audio later. 

The problem is that set_duration is called on newclip.audio *before* cutout() is called on newclip.audio. This results in the audio's duration being set to a shorter time (clip.duration - (end_remove-start_remove)) before the audio is cutout(), which calls set_duration again on the audio clip and reduces the audio duration to clip.duration - (end_remove-start_remove) - (end_remove-start_remove). 

#### Workarounds 

I am working around this issue this way:

```
newaudio = clip.audio.cutout(1,5)
newclip = clip.cutout(1,5)
newclip.audio = newaudio
```

tburrows13:
Thanks for the report. I've just tested it and I can confirm that this is a bug.

spollard:
A possible solution to this would be adjust the way that the apply_to_audio decorator works. At the moment, the newclip.audio is the passed in function (f) applied to newclip.audio. 
Inside apply_to_audio(): 
   ...
[...]

PR #1360, opened 2020-10-20
cutout error fixed

rishabhvarshney14:
related to #1227

I have tested these functions that used the apply_to_audio decorator
- with_start
- with_end
- with_duration
- subclip
- cutout
- loop
- time_mirror
- time_symmetrize

I have an issue with with_start. I am not sure what with_start suppose to do as compared to with_end with_start does not start with the expected starting time also time_mirror and time_symmetry gave error the same as here #1253.

~~Also in [speedx.py](https://github.com/Zulko/moviepy/blob/4356fcd596cda574e7e4ead9178b457cd4a3280a/moviepy/video/fx/speedx.py) it imports apply_to_audio and apply_to_mask but does not use it. Is this correct?~~ *Fixed*

- [X] I have provided code that clearly demonstrates the bug and that only works correctly when applying this fix
- [X] I have formatted my code using `black -t py36`

### DIFF
--- a/moviepy/decorators.py
+++ b/moviepy/decorators.py
@@ -42,7 +42,7 @@ def apply_to_audio(func, clip, *args, **kwargs):
 
     new_clip = func(clip, *args, **kwargs)
     if getattr(new_clip, "audio", None):
-        new_clip.audio = func(new_clip.audio, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


mondeja:
> Also in speedx.py it imports apply_to_audio and apply_to_mask but does not use it. Is this correct?

All module imports not used are has been removed in #1439

PR #1364, opened 2020-10-25
deprecation warnings

somstrom:
I have replaced deprecation warnings and also run `black -t py36 .` on my branch to format text.


tburrows13:
Closing this as a duplicate of #1365, and that one is from a proper source branch.

PR #1370, opened 2020-11-02
Tests for loop using BitmapClip #1368

somstrom:
I added some tests for loops using BitmapClip and they passed.

- I have added suitable tests demonstrating a fixed bug or new/changed feature to the test suite in `tests/`
- I have properly documented new or changed features in the documentation or in the docstrings
- I have formatted my code using `black -t py36`

### DIFF
--- a/tests/test_fx.py
+++ b/tests/test_fx.py
@@ -114,14 +114,20 @@ def test_even_size():
 
 
 def test_fadein():
-    color_dict = {"I": (0, 0, 0), "R": (255, 0, 0), "G": (0, 255, 0), "B": (0, 0, 255), "W": (255, 255, 255)}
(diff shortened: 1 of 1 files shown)


somstrom:
I added test for infinite loop and tests for fadein using BitmapClip

somstrom:
I would like to fix loop issue, so I got one question.

Loop function is called 2 times so previous_duration is rewritten. But i can't figure out why it is called 2 times yet.

I have replaced `previous_duration` with numeric value of clip duration and it works.

tburrows13:
Check the Gitter for discussion - someone else is also working on it :)

tburrows13:
Good job, I'm happy to merge.

PR #1375, opened 2020-11-05
Allow get_frame to return None to end clip when needed

SaraGoat:

### DIFF
--- a/moviepy/Clip.py
+++ b/moviepy/Clip.py
@@ -487,6 +487,7 @@ class Clip:
             t = frame_index / fps
 
             frame = self.get_frame(t)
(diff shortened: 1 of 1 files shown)


mondeja:
What is the purpose of this change?

SaraGoat:
I added it as a way to make clips with the callback interface that are not exactly an even amount of seconds long. I later discovered that You can use ImageSequenceClip with an iterable of numpy arrays later but it was hard to find that in the documentation

mondeja:
I suppose that you don't need this change anymore, so I'm closing this. Feel free to reopen this pull or open another.

PR #1383, opened 2020-11-24
Update quick_recipes.rst

pietelite:
The gaussian filter method from skimage is called `gaussian`, not `gaussian_filter`

- [x ] I have properly documented new or changed features in the documentation or in the docstrings


tburrows13:
Thanks for the change. It looks like `gaussian_filter` was deprecated in 2016 (0.12: https://scikit-image.org/docs/0.12.x/api/skimage.filters.html).

PR #1399, opened 2021-01-04
Bug fix: for rotating mask images.

gokuld:


mondeja:
Thank you for the fix @gokuld, I've not tested it yet. Could you provide a minimal reproducible example to start with, and we can add a proper test to make sure that this failure does not happen again in the future?

mondeja (approved):
Perfect, thank you! :wink:

Issue #1388, opened 2020-12-16
audio_normalize() of mute video leads to division by zero


ksfire:
```
ytclip = ytclip.subclip(0, 10).audio_normalize()
```

#### Expected Behavior
Normalize the audio. In this case, the input video is mute, no audio.

#### Actual Behavior

If the input video has no audio, division by zero is attempted.

```
Traceback (most recent call last):
  File "test.py", line 141, in <module>
    ytclip = ytclip.subclip(0, 10).audio_normalize()
  File "<decorator-gen-96>", line 2, in audio_normalize
  File "/Users/xxx/lib/python3.7/site-packages/moviepy/decorators.py", line 70, in audio_video_fx
    newclip.audio =  f(clip.audio, *a, **k)
  File "/Users/xxxx/lib/python3.7/site-packages/moviepy/audio/fx/audio_normalize.py", line 23, in audio_normalize
    return volumex(clip, 1 / mv)
**ZeroDivisionError: division by zero**
```

#### Steps to Reproduce the Problem
1. pick a video with no audio
2. clip = VideoFileClip("test.mp4").audio_normalize()

#### Workaround
Convert audio to soundarry and check if there are only zeros. In that case, do not normalize. ytclip.audio.to_soundarray(nbytes=4, buffersize=1000, fps=fps))

#### Specifications

  - Python Version: 3.7
  - Moviepy Version: 1.0.3
  - Platform Name: OSX
  - Platform Version: 10.14.6

PR #1401, opened 2021-01-05
Fix audio normalizing for mute clips

jcbrockschmidt:
Resolves issue #1388. Code that only runs when the fix is applied can be found in `tests/test_fx.py` within `test_normalize_muted`, and here for convenience:

```python
make_frame = lambda t: np.array([0.0])
clip = AudioClip(make_frame, duration=1, fps=44100)
clip = audio_normalize(clip)
```


jcbrockschmidt:
Does this change warrant a change to the docstring (check box 3)?

Also, does this change warrant a comment explaining what is happening (check box 4), or is it simple enough to be left as is?

tburrows13:
Hey, sorry for the late response. This looks great, and I like your docstring change and comment.
I'll merge it once I've added in a changelog entry.

moviepy/audio/fx/audio_normalize.py
```diff
@@ -19,4 +20,9 @@ def audio_normalize(clip):
     """
 
     max_volume = clip.max_volume()
-    return volumex(clip, 1 / max_volume)
+    if max_volume == 0:
```
mondeja, line 23: Why not just `max_volume or 1`? In that case, if is `0` the `1` will be used: 

```suggestion
    return volumex(clip, 1 / max_volume or 1)```

File: moviepy/audio/fx/audio_normalize.py [hunk 2]
```diff
@@ -9,6 +9,7 @@ def audio_normalize(clip):
 
     Return an audio (or video) clip whose audio volume is normalized
     so that the maximum volume is at 0db, the maximum achievable volume.
+    Audio that is muted is left unchanged.
```
mondeja, line 12: I think that this string is not needed because that "muted audio is left unchanged" is the expected behaviour of normalization.

tests/test_fx.py
```diff
@@ -456,5 +458,12 @@ def test_normalize():
     close_all_clips(locals())
 
 
+def test_normalize_muted():
+    make_frame = lambda t: np.array([0.0])
+    clip = AudioClip(make_frame, duration=1, fps=44100)
+    clip = audio_normalize(clip)
```
mondeja, line 464: To make sure, could you check that the returned array is properly muted? Something like this:
```suggestion
    z_array = np.array([0.0])
    make_frame = lambda t: z_array
    clip = AudioClip(make_frame, duration=1, fps=44100)
    clip = audio_normalize(clip)
    assert np.array_equal(clip.to_soundarray(), z_array)
```

mondeja (changes requested):
Thank you for the fix. Although I'm not a maintainer, I have some suggestions that could improve your pull request.

moviepy/audio/fx/audio_normalize.py
tburrows13: Agreed

tests/test_fx.py
tburrows13: Good idea

moviepy/audio/fx/audio_normalize.py
tburrows13: Not sure about your suggestion here... it would be shorter, but I don't think it is easier to read, and potentially worse performing, because it adds an extra layer to the fx stack and multiplies every audio frame by 1 every frame unnecessarily. Of course without actually benchmarking it, it might have negligible impact after all.

mondeja: :thinking: Maybe I'm not clear with the suggestion... I mean this:

```suggestion
   return volumex(clip, 1 / (max_volume or 1))```
```

Removing all the lines:

```python
    if max_volume == 0:
        # Nothing to normalize.
        # Avoids a divide by zero error.
        return clip.copy()
    else:
        return volumex(clip, 1 / max_volume)

tburrows13: No I understood that. I meant that it would be worse performing only when max_volume == 0 because it calls `volumex` unnecessarily in that situation.
I don't have a strong opinion so perhaps a comment to explain the `or 1`?

mondeja: Oh yes, my bad :man_facepalming:

moviepy/audio/fx/audio_normalize.py
jcbrockschmidt: Changed in aaf6300.

mondeja (approved):
For me is perfect. Feel free to merge it @tburrows13
Thank you for this fix @jcbrockschmidt!

PR #1405, opened 2021-01-10
Fix 'NameError' in 'docs/getting_started/videoclips.rst#Videoclip' example

mondeja:
The current example raises `NameError: name 'W' is not defined`. I've fixed adding global variables `HEIGHT` and `WIDTH`, added 2 spaces between comments, added spaces between numbers in formula and unified style of functions parameters.


tburrows13:
Looks good, thanks!

Issue #1409, opened 2021-01-11
CI is failing


tburrows13:
In case someone wants to investigate/fix:

The CI has stopped working properly in the last couple of months. Specifically, the macOS GH Actions and the Windows Appveyor tests both error early on with missing programs/files.

- [x] #1411 Windows Appveyor
- [ ] #1412 macOS GH Actions

tburrows13:
Thanks @mondeja for fixing these.

Its worth noting for the future that Travis will shut down/stop working soon. When that happens (or before it does) we can just get rid of the integration. It is duplicated in GH Actions now so it is no longer necessary anyway.

tburrows13:
Update: Removed Travis in https://github.com/Zulko/moviepy/commit/1800293e0631f0e5593c041d06991fbd24d380dd as it was already getting stuck in the queue for long periods of time.

PR #1411, opened 2021-01-11
Update 'Imagemagick.exe' download URL in Windows tests

mondeja:
The URL to download the imagemagick binary has been updated. Contributes to #1409.


tburrows13:
Nice, thanks for spotting that!

PR #1417, opened 2021-01-12
Remove TravisCI related code from tests

mondeja:
Remove obsolete code from tests, that depended on TravisCI.

### DIFF
--- a/tests/test_issues.py
+++ b/tests/test_issues.py
@@ -257,6 +257,40 @@ def test_issue_359():
         video.write_gif(filename=os.path.join(TMP_DIR, "issue_359.gif"), tempfiles=True)
 
 
+def test_issue_368():
+    import numpy as np
+    import matplotlib.pyplot as plt
+    from sklearn import svm
+    from sklearn.datasets import make_moons
+    from moviepy.video.io.bindings import mplfig_to_npimage
+
+    X, Y = make_moons(50, noise=0.1, random_state=2)  # semi-random data
+
+    fig, ax = plt.subplots(1, figsize=(4, 4), facecolor=(1,1,1))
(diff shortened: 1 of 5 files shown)


tests/test_issues.py
```diff
@@ -257,45 +257,6 @@ def test_issue_359():
...
-#         ax.contourf(xx, yy, Z, cmap=plt.cm.bone, alpha=0.8,
-#                     vmin=-2.5, vmax=2.5, levels=np.linspace(-2,2,20))
-#         ax.scatter(X[:,0], X[:,1], c=Y, s=50*weights, cmap=plt.cm.bone)
-#
-#         return mplfig_to_npimage(fig)
-#
-#     animation = VideoClip(make_frame, duration=2)
-#     animation.write_gif(os.path.join(TMP_DIR, "svm.gif"), fps=20)
-
-
```
tburrows13, line 298: I think that this test should stay. It was initially only skipped on Travis because it didn't work there. Then it was completely commented out (presumably because it didn't work in other places).

So you can remove the Travis checks from it, and leave it commented out if it still doesn't work, but I don't see why it should be removed entirely.

mondeja: Okay, I will try to make it work :+1:

tburrows13: Well I wouldn't worry about that too much, I just think that it shouldn't be deleted :)
The matplotlib parts of moviepy are very far of the main use case of the library and probably don't work very well thees days anyway.

mondeja: The test is working now. I've added optional requirements installation to Windows and MacOSX builds, to install matplotlib. Also, used the backend "agg" for matplotlib in the test.

PR #1421, opened 2021-01-12
Remove idle test and reduce media directory size

mondeja:
The removed test is doing nothing. The resized file is used only in `tests/test_ffmpeg_reader.py::test_ffmpeg_parse_infos` to check that audio is read and the audio FPS is 44100, so 4 minutes of video are not needed. This reduces the `media/` directory size from ~21mb to ~5.8mb.

### DIFF
--- a/tests/test_issues.py
+++ b/tests/test_issues.py
@@ -362,15 +362,6 @@ def test_issue_470():
     subclip.write_audiofile(os.path.join(TMP_DIR, "issue_470.wav"), write_logfile=True)
 
 
-def test_issue_246():
-    def test_audio_reader():
-        video = VideoFileClip("media/video_with_failing_audio.mp4")
-        subclip = video.subclip(270)
-        subclip.write_audiofile(
-            os.path.join(TMP_DIR, "issue_246.wav"), write_logfile=True
-        )
-
-
 def test_issue_547():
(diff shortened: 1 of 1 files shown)


tburrows13:
I wouldn't say that it is doing _nothing_, but agreed that the value of it is very low. For reference it was added in #377.
On the other hand, removing the large file doesn't change the pip install size (media aren't included in that), and it won't reduce the git clone size either because it will still be downloaded as part of the git history.

I'm happy to remove it if you still are though.

mondeja:
The test it's not executing nothing, it's only declaring the function. I've forced a exception inside the wrapped function and doesn't fail:

```python
def test_issue_246():
    def test_audio_reader():
        raise Exception("FOO")
        video = VideoFileClip("media/video_with_failing_audio.mp4")
        subclip = video.subclip(270)
        subclip.write_audiofile(
            os.path.join(TMP_DIR, "issue_246.wav"), write_logfile=True
        )
```

That's the reason because I'm saying that it is doing nothing. Please, can you double check it?

tburrows13:
Oh wow, I never noticed that before! You are right. I say scrap it then 👍🏼

mondeja:
Perfect, so go ahead with the merge :tada:

PR #1424, opened 2021-01-12
Add 'multiply_stereo_volume' FX; rename 'volumex' to 'multiply_volume'

mondeja:
This implements the FX that controls the volume level in two channels for stereo, but also works for more channels: mono (volume control), or more than 2 channels, in this case works properly if first channel of every pair is the left.


tburrows13:
This is great. A couple of thoughts:
I don't really like the name `audio_volume_lr`. I'd prefer not to use `l` and `r` in place of `left` and `right`. I can't think of a good name yet, so maybe you had some more ideas? `audio_left_right` isn't great either. `audio_adjust_stereo`?

Would you like to add a changelog entry? If not I'll do it before merging.

mondeja:
> I don't really like the name audio_volume_lr. I'd prefer not to use l and r in place of left and right.

Yes, I was thinking the same, but `audio_left_right` seems too much general to me, I prefer `audio_adjust_stereo`, but this could be confused with panning. Maybe `audio_stereo_volume`, or `audio_volume_stereo`, or `audio_volume_left_right`?

moviepy/audio/fx/audio_volume_lr.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Examples
+    ========
+
+    >>> music = AudioFileClip('music.ogg')
+    >>> audio_r = music.fx(audio_volume_lr, left=0, right=1)  # mute right channel
+    >>> audio_h = music.fx(audio_volume_lr, left=0.5, right=0.5)  # half audio
+    """
+
+    def fading(get_frame, t):
```
tburrows13, line 22: Why is it called `fading`?

moviepy/audio/fx/audio_volume_lr.py
```diff
@@ -0,0 +1,31 @@
...
+    of the left and right channel separately (with the factors `left`
+    and `right`). Makes a stereo audio clip in which the volume of left
+    and right is controllable.
+
+    Examples
+    ========
+
+    >>> music = AudioFileClip('music.ogg')
+    >>> audio_r = music.fx(audio_volume_lr, left=0, right=1)  # mute right channel
+    >>> audio_h = music.fx(audio_volume_lr, left=0.5, right=0.5)  # half audio
```
tburrows13, line 19:
```
    >>> audio_r = music.audio_volume_lr(left=0, right=1)  # mute right channel
    >>> audio_h = music.audio_volume_lr(left=0.5, right=0.5)  # half audio
```
I think that we'd prefer to push this method of using fx (particularly after #1340 which means that this will work no matter how you import `AudioFileClip`).

moviepy/audio/fx/audio_volume_lr.py
```diff
@@ -0,0 +1,31 @@
+import numpy as np
+
+from moviepy.decorators import audio_video_fx
+
+
+@audio_video_fx
+def audio_volume_lr(clip, left=1, right=1):
```
tburrows13, line 7: I think that the name should be in line with the current afx `volumex`. So have `volumex` and `stereo_volumex`? Or name this `audio_stereo_volume` and rename `volumex` to `audio_volume`? I'm not a massive fan of the name `volumex`, but I didn't change it in #1170.

mondeja: I love your suggestion `volumex` -> `audio_volume` and this as `audio_stereo_volume` because I think the same about `volumex` :+1:

moviepy/audio/fx/audio_volume_lr.py
mondeja: Replaced by `stereo_volume`.

mondeja:
:thinking: It seems that Coveralls fails because:

- Can't access to the secret since has been added to master before this pull?
- An open source project doesn't need a secret? I'm not using secrets in other [open source repos](https://github.com/mondeja/mdpo/blob/master/.github/workflows/ci.yml#L36-L40), but I'm not sure if it is needed.

mondeja:
Should wait for https://github.com/Zulko/moviepy/pull/1170#issuecomment-760961320.

PR #1425, opened 2021-01-12
Fix 'Clip::time_transform' docstring examples

mondeja:
Minor `t` argument errors in `moviepy/Clip.py::time_transform` lambdas of examples.


tburrows13:
I seem to remember that this was also fixed in a larger PR (that will be merged before 2.0) from several months ago. I don't remember which, but when I come across it I'll link it back to here.

PR #1427, opened 2021-01-12
Fix test 'test_matplotlib' in 'test_examples.py'

mondeja:
~~This suite has only one test that it's being skipped. It seems to me that is not necessary because I haven't found the corresponding example, but let me know if it's needed and I can unskip it.~~ Fixed test.


mondeja:
Okay, I haven't realized that this is an example included in docs :man_facepalming: So this should be fixed and keeped.

PR #1354, opened 2020-10-15
Fix #1338 - Replace deprecated functions

cryptobadger:
Closes #1338.

- s/frame.tostring()/frame.tobytes()
- replace generator as source for numpy stacks with a list

I'd like to contribute and this was marked as good first request and there is no activity on this issue albeit being quite simple to fix.

Issue #1338, opened 2020-10-08
Deprecation warnings


tburrows13:
When running the test suite it reveals some deprecations:

      /home/runner/work/moviepy/moviepy/moviepy/audio/AudioClip.py:121: FutureWarning: arrays to stack must be passed as a "sequence" type such as list or tuple. Support for non-sequence iterables such as generators is deprecated as of NumPy 1.16 and will raise an error in the future.

      /home/runner/work/moviepy/moviepy/moviepy/video/io/gif_writers.py:323: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.

Source: https://github.com/Zulko/moviepy/runs/1227641497 (under PyTest)
I think that the first one is NumPy and the second is Pillow (PIL).

It would be good to investigate and fix these 2 warnings.

vincenzoml:
I do have an "arrays to stack must be passed as 'sequence' type" error which I did not have before, with moviepy 1.0.3, when muxing audio and video (saving the video); isn't this merged? Now it's no longer a warning and the code fails.

hman009:
facing this issue while converting while running the to_soundarray() method 
TypeError: arrays to stack must be passed as a "sequence" type such as list or tuple.

vinitnalhe:
I was also facing the same issue, when using python3.11. 

I tried using the earlier version of python ( python 3.8) and  to_soundarray() ran successfully.

agamm:
> `pip install git+https://github.com/Zulko/moviepy.git@bc8d1a831d2d1f61abfdf1779e8df95d523947a5`

Why is this not added to a release yet?

SohamTilekar:
> > `pip install git+https://github.com/Zulko/moviepy.git@bc8d1a831d2d1f61abfdf1779e8df95d523947a5`
> 
> Why is this not added to a release yet?

Because The Zulko is Not Active and #2085.

keikoro:
The above discussion is not relevant, we are not actively using discussions. See issue #1874 instead.

shllgtca:
by using this source of moviepy I get not only error on moviepy/video/io/ffmpeg_reader.py and have to add on l.259 infos = error.decode('utf8')

but also I get error when trying to set_fps to an moviepy.audio.io.AudioFileClip.AudioFileClip object , bith errors which works in the newest version.

I still don't know how to solve the set_fps error. If anyone can help, tks in advance.

:/

> For anyone stumbling on this like I did:
> 
[...]

PR #1429, opened 2021-01-12
Replace 'tostring' by 'tobytes' in 'video.io.gif_writers::write_gif'

mondeja:
The method `array.tostring` was deprecated in Python v3.2, being replaced by [`array.tobytes`](https://docs.python.org/3/library/array.html#array.array.tobytes), so I've just changed their name. This method [it's removed in Python 3.9](https://bugs.python.org/issue38916) ([Release notes](https://cpython-ericholscher.readthedocs.io/en/sphinx-hoverxref/whatsnew/3.9.html#removed)) and the change prevents next warning:

```
/.../moviepy/video/io/gif_writers.py:323: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ Lots of method and parameter names have been changed. This will be explained bet
 ### Fixed <!-- for any bug fixes -->
 - Fixed BitmapClip with fps != 1 not returning the correct frames or crashing [#1333]
 - Fixed `rotate` sometimes failing with `ValueError: axes don't match array` [#1335]
-
+- Changed deprecated `tostring` method by `tobytes` in `video.io.gif_writers::write_gif` #1429
 
 ## [v2.0.0.dev2](https://github.com/zulko/moviepy/tree/v2.0.0.dev2)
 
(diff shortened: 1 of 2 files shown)


tburrows13:
Done in #1365 and #1354 :)

mondeja:
> Done in #1365 and #1354 :)

Ups :sweat_smile: Thank you for pointing it, any chance of merging the changes? I would like to use moviepy in Python3.9 without worries :)

tburrows13:
I didn't merge them because I was hesitant about the performance/memory impact of changing the yield into a return and I hadn't had time to properly investigate.

mondeja:
> I didn't merge them because I was hesitant about the performance/memory impact of changing the yield into a return and I hadn't had time to properly investigate.

I see, what about handle the cast to list in `to_soundarray`?

mondeja:
> I didn't merge them because I was hesitant about the performance/memory impact of changing the yield into a return and I hadn't had time to properly investigate.

I see, what about handle the cast to list in `to_soundarray`?

mondeja:
Could you consider this change separately? This breaks GIF creation in Python3.9 but the other deprecation even with latest Numpy can be ignored.

tburrows13:
Yes certainly. Feel free to merge this PR.

mondeja:
This has fixed second warning of #1338.

Issue #1355, opened 2020-10-15
Add more tests


tburrows13:
The test coverage is currently at 67%. It would be good to be able to automatically test more things in the test suite. You can see which files are covered at https://coveralls.io/github/Zulko/moviepy (or PyCharm has a test coverage plugin that can be enabled).

In particular:

- some effects in moviepy/video/fx are not tested at all in their respective test function in test_fx.py.
- video/compositing/transitions.py
- audio/fx/audio_fadein and audio_fadeout

For the video tests it is better to use BitmapClip because it allows you to test pixel-perfect effects. But generally just testing that an action can be performed without crashing is better than nothing.

mondeja:
# FX tests TODO

- [ ] `accel_decel`
- [x] `blackwhite`
- [ ] `blink`
- [x] `multiply_color`
- [x] `crop`
- [x] `even_size`
- [ ] `fadein` (more `duration` arguments)
- [ ] `fadeout` (rewrite using BitmapClip`)
- [x] `freeze`
- [ ] `freeze_region` (`mask` argument)
- [ ] `gamma_corr`
- [ ] `headblur`
- [x] `invert_colors`
- [ ] `loop` (use BitmapClip testing all behaviours)
- [ ] `lum_contrast` (rewrite using BitmapClip`)
- [ ] `make_loopable`
- [ ] `margin` (test `left`, `right`, `top`, `bottom` and `opacity` arguments)
- [ ] `mask_and`
- [ ] `mask_color`
- [ ] `mask_or`
- [x] `mirror_x`
- [x] `mirror_y`
- [ ] `painting`
- [ ] `resize` (rewrite using BitmapClip)
- [ ] `rotate` (other arguments beyond `angle`)
- [ ] `scroll`
- [x] `speedx`
- [ ] `supersample`
- [x] `time_mirror`
- [x] `time_symmetrize`
- [x] `audio_normalize`
- [ ] `multiply_volume`
- [x] `multiply_stereo_volume`

tburrows13:
It would be quite hard to write BitmapClip tests for all of the effects, I think. Although you did manage with `blackwhite` in #1430 I wouldn't say it is worth it for some of the more 'color-gradienty' ones (such as `headblur` and `painting`).

mondeja:
> It would be quite hard to write BitmapClip tests for all of the effects...

I have found it much easier to use ColorClip in tests for `mask_or` and` mask_and`.

mondeja:
[...]

PR #1430, opened 2021-01-12
Test properly 'blackwhite' FX using 'BitmapClip'

mondeja:
`blackwhite` FX properly tested using `BitmapClip` class. Contributes to #1355.

* Added class property `DEFAULT_COLOR_DICT` to `BitmapClip` class to avoid the repeat of default colors definition. Let me know if this is not wanted.
* Tested all posible argument combinations.

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -1337,6 +1337,19 @@ class TextClip(ImageClip):
 
 
 class BitmapClip(VideoClip):
+    DEFAULT_COLOR_DICT = {
+        "R": (255, 0, 0),
+        "G": (0, 255, 0),
+        "B": (0, 0, 255),
+        "O": (0, 0, 0),
+        "W": (255, 255, 255),
+        "A": (89, 225, 62),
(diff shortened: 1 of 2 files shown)


moviepy/video/VideoClip.py
```diff
@@ -1337,6 +1337,19 @@ def search(string, arg):
 
 
 class BitmapClip(VideoClip):
+    DEFAULT_COLOR_DICT = {
```
mondeja, line 1340: I have reviewed this pull in depth and I see everything correct. I don't ask you for a complete review because I understand that the test is a bit complex, but I would like to ask you if this `DEFAULT_COLOR_DICT` class property could have any downside. Also, do you agree with the name `DEFAULT_COLOR_DICT` @tburrows13?

tburrows13: Yep, I'm not going to pretend to understand the test, but I really like the `DEFAULT_COLOR_DICT` change

tburrows13 (approved):

PR #1434, opened 2021-01-13
Preserve links to pull requests in CHANGELOG

mondeja:
After v1.0.3, all CHANGELOG pull requests has been wrapped by `[` `]`, but this removes the link rendered by Github for Markdown files, so I've removed the wrapping characters. See the difference in next image:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,7 @@ Install with `pip install moviepy --pre --upgrade`.
 ### Removed <!-- for now removed features -->
 - Support removed for Python versions 2.7, 3.4 & 3.5
 - Setting paths to ImageMagick and FFMpeg binaries in ``config_defaults.py`` is no longer possible [\#1109](https://github.com/Zulko/moviepy/pull/1109)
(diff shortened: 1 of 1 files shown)


tburrows13:
It isn't actually that simple. The <=1.0.3 changelogs were done with an automatic tool that make the links look like proper GitHub links but are actually just standard markdown links (as you can see here: https://raw.githubusercontent.com/Zulko/moviepy/master/CHANGELOG.md)

Since then I've been curating the changelog manually as I go, and not bothering with the full markdown links. This means that the links don't work in [CHANGELOG.md](https://github.com/Zulko/moviepy/blob/master/CHANGELOG.md), but they do still work in the [release page](https://github.com/Zulko/moviepy/releases).

Before v2.0.0 releases I thought that I might go through them and make them proper markdown links like they were before, but I've not done that so far.

mondeja:
>  isn't actually that simple. They work exactly the same after your change:

Ups, that's right :sweat_smile: 

> Before v2.0.0 releases I thought that I might go through them and make them proper markdown links like they were before, but I've not done that so far.

Don't worry about it, I will do in this PR :+1:

PR #1436, opened 2021-01-13
Test Python3.9 in Linux

mondeja:
Since Python3.9 was released 3 months ago, I think that is time to start testing that version in CI.


tburrows13:
Rather strange that the 3.9 version takes 5 times as long (14 minutes) as the other linux ones. Its only a few minutes longer than the macOS and windows tests though, so it doesn't matter _too_ much (windows takes ~4 minutes, but doesn't run in parallel, so ~8 total).

It seems that Py3.9 must build the scripy wheel manually.

PR #1442, opened 2021-01-14
Remove 'coreader' methods adding formal '__copy__' API

### LINKED ISSUE #1441: 'coreader' methods of 'VideoFileClip' and 'AudioFileClip' are buggy
These methods were created at #1332 and https://github.com/Zulko/moviepy/commit/fdc51d29b7ce2f3b2c4d0d907c1c62b5da168128, but IMHO I think that are a misunderstanding of how `copy.copy` and `copy.deepcopy` works in Python.

- `coreader` methods currently does a shallow copy (like `copy.copy`) but only taking into account `filename` parameter.
- `Clip.copy` method is working like `copy.deepcopy`, but even for `AudioClip` the `mask` and `audio` attributes are checked in the instance.

### Actual Behavior

The `coreader` functionality is not working. It's implemented as `return AudioFileClip(self.filename, self.buffersize)`, but `AudioFileClip` accepts `buffersize` as a kwarg argument, so it's not taking the `buffersize`. The same applies to `VideoFileClip`.

The documentation of `coreader` methods are very poor, what are they supposed to do? I think that should be added an example, if they are really implemented and are not bad `copy.copy`-like code.

```python
>>> from moviepy.editor import *
>>> clip = AudioFileClip('media/crunching.mp3', buffersize=1000)
>>> clip_2 = clip.coreader()
>>> clip.buffersize
1000
>>> clip_2.buffersize
200000
```

### Expected behaviour

I'm not sure, are a failed copy, I'm thinking in two possible cases:

- New instance with only filename as the same argument
- New instance with filename and buffersize as same arguments 
  
I suspect that are just weird copy code, and `copy` method should be used instead (ovewriting it in `AudioClip` to because a `deepcopy` in those are not needed).

### Specifications

  - Python Version: 3.8.5
  - Moviepy Version: master
  - Platform Name: Ubuntu
  - Platform Version: 20.04

mondeja:
+ Removed `coreader` methods of `VideoFileClip` and `AudioFileClip`.
+ Added safe support for ``copy.copy(clip)`` and ``copy.deepcopy(clip)``.
+ Replace `copy(clip)` by `clip.copy()` in `video.fx.blink`.
+ Removed test of `video.fx.blink` that was added because `copy()` was removed from builtin Python functions.

I haven't tested performance, but this change should speed up the library, since almost all the transormations returns copies and now some unnecesary checks are not executed in `Clip.copy`.

Closes #1441


tburrows13:
I fixed the changelog, so feel free to merge when you like.

mondeja:
Blocked until #1453 merges.

moviepy/video/VideoClip.py
```diff
@@ -116,6 +117,33 @@ def h(self):
...
+        clip.subclip, etc.)
+
+        Acts like a deepcopy except for the fact that readers and other
+        possible unpickleables objects are not copied.
+        """
+        cls = self.__class__
+        new_clip = cls.__new__(cls)
+        for attr in self.__dict__:
+            value = getattr(self, attr)
+            if attr in ("mask", "audio"):
```
tburrows13, line 137: I only just noticed this, but aren't lines 137-138 and 141-142 doing the same thing?

mondeja: Yes, I didn't notice :man_facepalming:  Let me fix it

mondeja: Fixed in https://github.com/Zulko/moviepy/commit/fac722df7c4c0c33a9d98877b13647f4115f7b53

PR #1444, opened 2021-01-14
Improve 'resizer' function discovering in 'moviepy.video.fx.resize'

mondeja:
- Improve `resizer` function autodiscovering in `video.fx.resize`. If any resizer can't be found, shows why and how to solve the problem, like (example without install PIL nor OpenCV, using Scipy v1.4.1):

```
ImportError: fx resize needs OpenCV or Scipy or PIL
- scipy.misc.imresize not found (was removed in scipy v1.3.0, you are using v1.4.1, install 'Pillow')
- PIL not found (install 'Pillow')
- OpenCV not found (install 'opencv-python')
```

- Scipy v1.3.0 was removed the function [`scipy.misc.imresize`](https://docs.scipy.org/doc/scipy-1.2.0/reference/generated/scipy.misc.imresize.html) and, since moviepy supports that version, this lacking is checked and shown to the user if needed.

### DIFF
--- a/moviepy/video/fx/resize.py
+++ b/moviepy/video/fx/resize.py
@@ -60,7 +60,7 @@ def _get_scipy_resizer():
             return (
                 None,
                 [
-                    "scipy.misc.imresize not found (was deprecated in scipy v1.3.0,"
+                    "scipy.misc.imresize not found (was removed in scipy v1.3.0,"
                     f" you are using v{__scipy_version__}, install 'Pillow')"
                 ],
             )
--- a/moviepy/video/fx/resize.py
+++ b/moviepy/video/fx/resize.py
@@ -1,10 +1,8 @@
-resize_possible = True
-
-try:
-    # TRY USING OpenCV AS RESIZER
-    # raise ImportError #debugging
-    import cv2
-    import numpy as np
+def _get_cv2_resizer():
+    try:
+        import cv2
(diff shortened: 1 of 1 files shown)


moviepy/video/fx/resize.py
```diff
@@ -16,43 +14,105 @@ def resizer(pic, new_size):
...
+        scipy_version_info = tuple(
+            int(num) for num in __scipy_version__.split(".") if num.isdigit()
+        )
+
+        # ``scipy.misc.imresize`` was removed in v1.3.0
+        if scipy_version_info >= (1, 3, 0):
+            return (
+                None,
+                [
+                    "scipy.misc.imresize not found (was deprecated in scipy v1.3.0,"
```
tburrows13, line 63: I'm not sure that you understand the definition of "deprecated" (which I corrected in https://github.com/Zulko/moviepy/pull/1442/commits/c6922e52fd02369a995785f4ce14054a82383b89 as well). 

When a function is deprecated it is still available, and has not yet been removed. The deprecation is a signal that it is no longer being supported and will be removed in the future. According to https://docs.scipy.org/doc/scipy/reference/release.1.3.0.html#scipy-interpolate-changes, `imresize` was _deprecated_ in v1.0.0 and _removed_ in v1.3.0.

Issue #1447, opened 2021-01-15
Fix coveralls integration


tburrows13:
Since removing Travis in https://github.com/Zulko/moviepy/commit/1800293e0631f0e5593c041d06991fbd24d380dd the coveralls integration has stopped updating: https://coveralls.io/github/Zulko/moviepy. The [replacement in GH Actions](https://github.com/Zulko/moviepy/blob/master/.github/workflows/test_suite.yml#L156-L160) has coveralls working, but it isn't integrated with this repo yet. The token there was added for testing, and is for [my fork's coveralls](https://coveralls.io/github/tburrows13/moviepy) instead.

@Zulko, would you be able to access the token from https://coveralls.io/github/Zulko/moviepy, add it as a repo secret according to https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets, and then replace the token at https://github.com/Zulko/moviepy/blob/master/.github/workflows/test_suite.yml#L158 with the secret?

Thanks!

PR #1448, opened 2021-01-15
test_suite.yml now uses a secret coveralls token

Zulko:
Should address #1447 . Thanks Tom!


mondeja:
Coveralls [ran successfully](https://coveralls.io/builds/36354124), so let's go ahead with the merge!

PR #1449, opened 2021-01-15
Test coveralls from pull request

mondeja:
This is a check to clarify [this](https://github.com/Zulko/moviepy/pull/1424#issuecomment-760962244), don't merge, will be closed after the CI execution.


tburrows13:
I think that the secret is set up wrong. It appears that it is reading the env variable as empty.

PR #1451, opened 2021-01-15
Remove audio_loop afx

tburrows13:
`audio_loop` is unneeded because `loop` can be applied to audio clips the same as video clips.


mondeja (commented):
[This file](https://github.com/Zulko/moviepy/blob/ef63c95b99b31d58385e92422199b0a048985f45/docs/ref/audiofx/moviepy.audio.fx.all.audio_loop.rst) and [this line](https://github.com/Zulko/moviepy/commit/20309f16a4e441b918dba7f10676244b70a78884#diff-ca0afe8702e881df0f3eac867b2ba47ed4146981f50b5f36b2bc3baf552572b4R44) from `docs/ref/audiofx.rst` should be removed.

mondeja:
Just a suggestion, you can use `pytest -svv tests/test_fx.py::test_loop` to execute only that test and showing the output of prints.

tburrows13:
Ah I always forget to change back the temporary test function calls at the bottom 😞 I'll try that next time

PR #1454, opened 2021-01-16
Update Scipy version in optional dependencies

mondeja:
This change increases the installation speed of Scipy in Python 3.9, since it is not built from the `.tar.gz` source. I've tested the two imports of Scipy with v1.6.0 and both work:

- `import scipy.ndimage as ndi`
- `from scipy.ndimage.filters import sobel`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,7 @@ optional_reqs = [
     "opencv-python>=3.0,<4.0",
     "scikit-image>=0.13.0,<1.0",
     "scikit-learn",
-    "scipy>=0.19.0,<1.5",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -86,7 +86,7 @@ def run_tests(self):
     "opencv-python>=3.0,<4.0",
     "scikit-image>=0.13.0,<1.0",
     "scikit-learn",
-    "scipy>=0.19.0,<1.5",
+    "scipy>=0.19.0,<=1.6.0",
```
tburrows13, line 89: Better to go "<1.7.0"?

PR #1455, opened 2021-01-16
Update dependency version requirements

tburrows13:
Some of these were very out of date, so I went through them and updated them to current versions. I removed the version restrictions on the optional requirements because they are used so little in moviepy, and only in niche corners of it that in my opinion there is no point spending a lot of time ensuring that each version is properly compatible. They are very likely to be compatible anyway because of how little each library is used.
Better to let users use whichever version of those libraries that they like.

Closes #1454.


mondeja (approved):
Perfect, thank you!

PR #1456, opened 2021-01-16
Delete redundant on_color.py

tburrows13:
Both `on_color.py` and `VideoClip.on_color()` were added in v0.2 (https://github.com/Zulko/moviepy/commit/fdc51d29b7ce2f3b2c4d0d907c1c62b5da168128), but only `VideoClip.on_color()` has been maintained/updated since.

Somewhat related: is there a better name for `on_color` and/or its argument `col_opacity`? (-> `background_opacity`?)


mondeja:
I'm not a native English speaker, but perhaps `VideoClip.over_color` and `color_opacity`?

mondeja (commented):
Should the block after line 757 be moved to `ImageClip` calling `VideoClip.on_color` before this block using `super()`?

```python
        if (
            isinstance(self, ImageClip)
            and (not hasattr(pos, "__call__"))
            and ((self.mask is None) or isinstance(self.mask, ImageClip))
        ):
            new_result = result.to_ImageClip()
            if result.mask is not None:
                new_result.mask = result.mask.to_ImageClip()
            return new_result.with_duration(result.duration)
```

keikoro:
> Somewhat related: is there a better name for on_color and/or its argument col_opacity? (-> background_opacity?)

I agree, `on_color` is not very clear. Would `color_underlay` (equivalent to overlay) make sense? I don't know if "background" is unambiguous enough considering that photo + film images have foregrounds and backgrounds as well. The other one could then be `underlay_opacity`.

Also, there is no real consistency in the naming of functions – actions vs. nouns – is there? (Though some could be either, I suppose.) -> `color_underlay` vs `underlay_color`...

mondeja:
We can change the name later, I'm going to merge this as is. Thank you @tburrows13 :+1:

PR #1459, opened 2021-01-16
Add AudioClip stereo example

mondeja:
Just like the mono example, but using two sine waves of A note at different frequencies.


mondeja:
I've [reported the failure with tqdm](https://github.com/tqdm/tqdm/issues/481#issuecomment-761803390) in Python3.6. Any inconvenience with the example @tburrows13?

tburrows13 (approved):

PR #1460, opened 2021-01-16
Speed up 'AudioClip.to_soundarray'

mondeja:
This simple change improves the performance of `AudioClip.to_soundarray`. Variables `fps`, `max_duration` and `stacker` are not used if `tt is not None`, so can be moved inside the block. This function is executed multiple times from `AudioClip.iter_chunks` iterator, so performance increasements here are worth.


mondeja:
I'm totally sure that this does not introduces any error, so merging as is.

PR #1471, opened 2021-01-19
Add support for retrieve clip frames number using 'clip.n_frames'

### LINKED ISSUE #1058: how to get the frames number of the whole video ?
hi,
Just wanna get the frames number directly,
do not want to use **duration*fps**,
any help ?
thx

mondeja:
Closes #1058


tburrows13:
I prefer `n_frames` to `nframes` (and I'll likely change similar ones like `nchannels` to that in the future). Other than that its good.

PR #1438, opened 2021-01-14
Documentation improvements

tburrows13:
This will be WIP for quite a few days/weeks.

You can see the changes live here: https://moviepy-tburrows13.readthedocs.io/en/improve-docs/index.html

I'm still getting the hang of Sphinx, so there are some broken links for now.

PR #1474, opened 2021-01-20
Add 'center', 'translate' and 'bg_color' arguments to 'rotate' FX

mondeja:
- Added support for arguments `center`, `translate` and `bg_color` (in PIL is named as `fillcolor`).
- Add PIL version checking to make sure that arguments are supported by the PIL version of the user.
- Fix documentation.


mondeja:
I'm not sure if `bg_color` is the name that would use moviepy. What do you think @tburrows13?

tburrows13:
Just fyi for when you are updating docstrings bear in mind that we are using numpydoc (https://numpydoc.readthedocs.io/en/latest/format.html#sections), and in my doc improvements PR (#1438) I'm adding things like `optional` and argument type (as in numpydoc standards) to the parameter documentation. Feel free to do the same when you are adding new documented parameters.

`bg_color` is what is used elsewhere in moviepy so you should use that for now. I didn't change it in #1170, but maybe it should be changed in the future, but if that happens, we can easily change this one along with the rest.

PR #1477, opened 2021-01-21
Remove 'make_loopable' from transitions

### LINKED ISSUE #1476: 'make_loopable' function repeated
`make_loopable` function is repeated in `video.compositing.transitions` and `video.fx.make_loopable`.
One of both must die :fire:

mondeja:
Closes #1476


tburrows13 (approved):
I think that I prefer the `overlap_duration` name here to the `overlap_time` parameter in fx.make_loopable. If you agree, then that could either be changed here or in a separate PR.

PR #1478, opened 2021-01-21
Rename 'speedx' FX by 'multiply_speed'

mondeja:
Following from #1424, renamed `speedx` FX by `multiply_speed`, link #1170.


tburrows13:
Another good reason for changes the names of `effectx` effects is to stop them getting mixed up with `mirror_x`, which has a completely different meaning :D

tburrows13 (approved):

PR #1479, opened 2021-01-21
Setup 'isort' and 'pre-commit'

### LINKED ISSUE #1473: Set up pre-commit and isort
To avoid #1084 periodically (with what that means for GIT blames) and to simplify the development process, I would do:

- Set up [`pre-commit`](https://pre-commit.com/) ([example config](https://github.com/mondeja/mdpo/blob/master/.pre-commit-config.yaml#L1-L16)). The developer will only need to do `pre-commit install` after the installation and literally forget to run `black`, `flake8` and `isort` anymore, since they will be executed before each commit.
- Set up isort inside the configuration of pre-commit.
- Run isort checking in CI.
- Document this in CONTRIBUTING.
- Remove "Linting and formatting" from README.

mondeja:
To avoid #1084 periodically and to facilitates the development process, I've configured isort (also in CI) and pre-commit. Closes #1473

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,4 +6,4 @@ Delete them if they are not appropriate for this pull request.
 - [ ] I have added suitable tests demonstrating a fixed bug or new/changed feature to the test suite in `tests/`
 - [ ] I have properly documented new or changed features in the documentation or in the docstrings
 - [ ] I have properly explained unusual or unexpected code in the comments around it
-- [ ] I have formatted my code using `black -t py36` 
+- [ ] I have ran pre-commit hooks while doing the commit or running `pre-commit run --all-files`
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,4 +6,3 @@ Delete them if they are not appropriate for this pull request.
 - [ ] I have added suitable tests demonstrating a fixed bug or new/changed feature to the test suite in `tests/`
 - [ ] I have properly documented new or changed features in the documentation or in the docstrings
 - [ ] I have properly explained unusual or unexpected code in the comments around it
-- [ ] I have ran pre-commit hooks while doing the commit or running `pre-commit run --all-files`
--- a/.github/workflows/format_check.yml
+++ b/.github/workflows/format_check.yml
@@ -36,7 +36,7 @@ jobs:
         run: flake8 moviepy --show-source
   isort:
     runs-on: ubuntu-latest
-    name: Isort
+    name: isort
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python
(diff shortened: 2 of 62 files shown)


tburrows13:
Great. Is it possible to get it to run automatically when you commit?

mondeja:
> Is it possible to get it to run automatically when you commit?

Yes, after you configure the hook pre-commit with `pre-commit install` it will run when you do `git commit -m "foo bar"` and if it fails, the commit is canceled.

CONTRIBUTING.md
```diff
@@ -48,7 +49,7 @@ You do not have to have finished your feature or bug fix before submitting a PR;
 
 Before submitting PRs:
 
-- make sure your code still conforms to the formatting standard (using Black formatter: `$ black -t py36 .`)
+- make sure that your code passes code checks: `$ pre-commit run --all-files`
```
tburrows13, line 52: In which case we probably don't need lines like this?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,4 +6,4 @@ Delete them if they are not appropriate for this pull request.
 - [ ] I have added suitable tests demonstrating a fixed bug or new/changed feature to the test suite in `tests/`
 - [ ] I have properly documented new or changed features in the documentation or in the docstrings
 - [ ] I have properly explained unusual or unexpected code in the comments around it
-- [ ] I have formatted my code using `black -t py36` 
\ No newline at end of file
+- [ ] I have ran pre-commit hooks while doing the commit or running `pre-commit run --all-files`
```
tburrows13, line 9: And maybe this should be changed to something like "I have commit hooks installed with `pre-commit install`"?

CONTRIBUTING.md
mondeja: Sure, I was doubting about include it

.github/PULL_REQUEST_TEMPLATE.md
mondeja: To the stake with it!

CONTRIBUTING.md
mondeja: Sure, I was doubting about to include it

.github/workflows/format_check.yml
```diff
@@ -34,3 +34,18 @@ jobs:
         run: flake8 tests --show-source
       - name: Lint moviepy
         run: flake8 moviepy --show-source
+  isort:
+    runs-on: ubuntu-latest
+    name: Isort
```
tburrows13, line 39:
```suggestion
    name: isort
```
A bit of a nitpick, but unlike Black and Flake8, isort is always lowercase :)

mondeja: Yep, looks so bad capitalized :+1:

mondeja:
Okay, I'm merging this as is. Opened pulls must be rebased before merge to make sure that are in line with the isort check.

Issue #1358, opened 2020-10-18
Gap (delay) between loops in 'loop' FX


ArztKlein:
I'd like to request an argument with the audio_loop function which is a gap duration - a gap of time with no audio between the loops.
https://zulko.github.io/moviepy/ref/audiofx/moviepy.audio.fx.all.audio_loop.html

tburrows13:
It's a good idea. Probably worth noting that I was thinking that we could probably get rid of `audio_loop` entirely in favour of the existing `loop` (which should work for both video and audio) because whilst the implementation is quite different, the end result is the same. Presumably the 'gap' could also work in the video version of `loop` and just have a blank (or transparent) screen for the gap duration.

DJ73:
I would like to try this out. Although, if there is a common loop option for audio and video, then the function should have a different file structure. I think `audio_loop` should ideally return just the audio, without the video and similarly, `video_loop` could return the video without audio. A common `loop` function can probably have options to retain audio or video such that it calls the other two for further processing

tburrows13:
I'd like to keep the current `loop` as it is, as it follows a standard set out by other effects in video.fx (such as e.g. time_mirror), that is, have it always apply to video, audio and mask, if available. In the rare case where the user wishes to apply it only to the audio of a video, then they can simply do `clip.audio = clip.audio.loop(...)`. Similar solutions can be done to any variation that you can think of.

Before you add the gap parameter, I'd recommend that you do #1368 first :)

tburrows13:
`audio_loop` is now gone in #1451, so we just have `loop` that works on audio and video. Anyone is still welcome to implement a `gap` parameter to it.

OsaAjani:
Thank you for your contributions and for reporting issues in this repository. With the release of v2, which introduces significant changes to the codebase and API, we’ve reviewed the backlog of open PRs and issues. Due to the length of the backlog and the likelihood that many of these are either fixed or no longer applicable, we’ve made the decision to close all previous PRs and issues.

If you believe that any of these are still relevant to the current version or if you'd like to reopen a related discussion, please feel free to create a new issue or pull request, referencing the old one.

Thank you for your understanding and continued support!

PR #1481, opened 2021-01-22
Add 'audio_delay' FX

mondeja:
This is a common effect working with audio. Keep in mind that this effect does not fit a `loop` FX behaviour, since the sound is multidimensional because you can have multiples copies playing at the same time. Even with an `offset` attribute in current `loop` FX this effect would be impossible to achieve.


moviepy/audio/fx/audio_delay.py
```diff
@@ -0,0 +1,55 @@
...
+
+    decay (float)
+      Multiplication factor for the volume level of the last repetition. Each
+      repetition will have a value in the linear function between 1 and this value,
+      increasing or decreasing constantly. Keep in mind that the last repetition
+      will be muted if this is 0, and if is greater than 1, the volume will increase
+      for each repetition.
+
+    Examples
+    ========
```
tburrows13, line 30:
```suggestion
    offset : float, optional
      Gap between repetitions start times, in seconds.

    n_repeats : int, optional
      Number of repetitions (without including the clip itself).

    decay : float, optional
      Multiplication factor for the volume level of the last repetition. Each
      repetition will have a value in the linear function between 1 and this value,
      increasing or decreasing constantly. Keep in mind that the last repetition
      will be muted if this is 0, and if is greater than 1, the volume will increase
      for each repetition.

    Examples
    --------
```

According to https://numpydoc.readthedocs.io/en/latest/format.html#sections, this is the preferred formatting style.

tburrows13:
I'm not sure about this one... isn't it basically the same as #1358? I wonder if it would make more sense to implement this one in `loop`, if you think that the `decay` would work there?

mondeja:
> I'm not sure about this one... isn't it basically the same as #1358? I wonder if it would make more sense to implement this one in loop, if you think that the decay would work there?

Keep in mind that is not the same as a loop. In a loop, the whole loop is played and after that starts another copy. In a delay effect, you can have multiple copies running at the same time.

tburrows13:
> Keep in mind that is not the same as a loop. In a loop, a whole clip is played and after that starts another copy. In a delay effect, you can have multiple copies playing at the same time.

Oh right, in which case this is fine. There would be no way to implement loop with #1358 that allows a 'negative' gap, so this is a good way to do it.

I have a feeling that the description or name could be improved, but I can't take that further right now. Depending on how those change, perhaps the default decay should be 1?

mondeja:
> I have a feeling that the description or name could be improved

The description sure, but the name is a standard in music and sound areas, it feel to me that is the right name.

mondeja:
Do you see the description right now @tburrows13?

PR #1510, opened 2021-02-28
VideoFileClip crop direct from ffmpeg (not fx.crop)

biojet1:


tburrows13:
Hello, thanks for the contribution. I'm too busy right now to look at merging it, but I can't see any reason not to include it in v2.0. My main observation is that your `calc_crop` looks very similar to some of the logic in `vfx.crop`, so maybe have a look at merging that somehow? If you think that wouldn't make sense then don't worry about it.

8secz-johndpope:
@tburrows13  - consider adding it a key contributor - we need gpu support - you look like you'll never have time.

OsaAjani:

Issue #1533, opened 2021-04-05
Black lint failures


tburrows13:
Just a note that the GH Actions "Code Format Check" is currently broken due to Black deleting their stable branch. The fix is easy as described [here](https://github.com/psf/black/issues/2079#issuecomment-812359146) but it sounds like they'll readd `stable` at some point so I'm not going to bother unless the issue persists.

PR #1534, opened 2021-04-06
Fix black version in Code Format Check workflow of CI

mondeja:
Just an update to the format check workflow to keep it green again.


tburrows13:
There's also https://github.com/Zulko/moviepy/blob/master/.pre-commit-config.yaml#L19 (linking #1533).

mondeja:
Good point :+1:

PR #1537, opened 2021-04-09
Update preview.py, fix issue #575

Victrollia:
Fixed issue #575. Added two pg.quit statements at lines 146 and 162. For line 146, this statement adds quit functionality for " event.type == pg.KEYDOWN and event.key == pg.K_ESCAPE". This allows the window to close whenever the event triggers. For line 162, once the for loop ends (the video ends), the window closes instead of crashing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed `Clip.cutout` transformation not being applied to audio [\#1468](https://github.com/Zulko/moviepy/pull/1468)
 - Fixed arguments inconsistency in `video.tools.drawing.color_gradient` [\#1467](https://github.com/Zulko/moviepy/pull/1467)
 - Fixed `fps` not defined in `CompositeAudioClip` at initialization [\#1462](https://github.com/Zulko/moviepy/pull/1462)
+- Fixed `clip.preview()` crashing at exit when running inside Jupyter Notebook in Windows [\#1537](https://github.com/Zulko/moviepy/pull/1537)
 
 
 ## [v2.0.0.dev2](https://github.com/zulko/moviepy/tree/v2.0.0.dev2) (2020-10-05)
--- a/moviepy/video/io/preview.py
+++ b/moviepy/video/io/preview.py
@@ -143,6 +143,7 @@ def preview(
                 if audio:
                     video_flag.clear()
                 print("Interrupt")
+                pg.quit()
                 return result
 
             elif event.type == pg.MOUSEBUTTONDOWN:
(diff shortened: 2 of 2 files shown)


mondeja (commented):
Thank you for opening this pull request @Victrollia

I'm really worried about if this could have side effects, for example: have you tried this change concatenating multiples previews in the same Python session? Something like:

```python

filename = r"media\chaplin.mp4"
clip1 = VideoFileClip(filename)
clip2 = VideoFileClip(filename)

clip1.preview()
clip2.preview()
```

Also, I've tried to reproduce the problem and I couldn't in the next environment:

- Windows: 10 version 1909
- Moviepy: **master** branch
- pygame: 2.0.1

Could you specify what are the versions of Windows, moviepy and pygame for which this problem happens? This only happens when running the preview in a Jupyter Notebook, or also appears standalone?

Victrollia:
> Thank you for opening this pull request @Victrollia
> 
> I'm really worried about if this could have side effects, for example: have you tried this change concatenating multiples previews in the same Python session? Something like:
> 
> ```python
> from moviepy.editor import VideoFileClip
> 
> filename = r"media\chaplin.mp4"
> clip1 = VideoFileClip(filename)
> clip2 = VideoFileClip(filename)
> 
> clip1.preview()
> clip2.preview()
> ```
> 
> Also, I've tried to reproduce the problem and I couldn't in the next environment:
> 
> * Windows: 10 version 1909
> * Moviepy: **master** branch
> * pygame: 2.0.1
> 
> Could you specify what are the versions of Windows, moviepy and pygame for which this problem happens? This only happens when running the preview in a Jupyter Notebook, or also appears standalone?

Tested on Jupyter Notebook, on Windows 10 10.0.19041 Build 19041. Tested with two previews like in the sample code you provided, once one closes, the next clip preview opens. Didn't add a checkmark on tests because it was mentioned prior that it is tricky to test pygame. Only two lines were added to close the pygame to minimize the chances of any side effects. The first one closes the window if it was interrupted midway while the clip was playing, but then I noticed that it would still crash if the preview ended, that's why another line of code was added that closes the window once it ended the preview. Contributing to this project has been an assignment for a capstone class for CS degree, and therefore my team (3 other people) had the same issue with crashing previews on their Jupyter Notebooks. The Moviepy master branch was cloned recently so it's the latest version. Tested on Python 3.8 & 3.9. Using the latest version of pygame (2.0.1). Let me know if you have any other questions.

mondeja (approved):
Thanks for the detailed response @Victrollia :pray:

I finally have been able to reproduce it. Seems that your fix has no side effects in any platform and solves the problem, it's perfect :+1:

mondeja:
Thanks again for the fix! If you or someone from your team finds another problem, do not hesitate to open a pull request :100:

PR #1572, opened 2021-05-21
Add 'start_time' and 'end_time' parameters to multiply_volume FX

### LINKED ISSUE #1571: Allow to specify time range in multiply_volume effect
Currently, if you want to create a silence in a clip, you must do something like:

```python
import numpy as np

from moviepy.editor import *


def make_frame(t):  # create stereo A note
    return np.array(
        [np.sin(440 * 2 * np.pi * t), np.sin(160 * 2 * np.pi * t)]
    ).T.copy(order="C")

wave_clip = AudioClip(make_frame, duration=2, fps=44100)
silenced_clip = concatenate_audioclips([
    wave_clip.subclip(0, .5),  # half second playing
    wave_clip.subclip(.5, 1).multiply_volume(0),  # half second silence
    wave_clip.subclip(1, 2),  # one second playing
])
```

Looks crazy and is not optime. I've been looking for other solutions but I can't find nothing more. I would prefer something like this:

```python
import numpy as np

from moviepy.editor import *


def make_frame(t):  # create stereo A note
    return np.array(
        [np.sin(440 * 2 * np.pi * t), np.sin(160 * 2 * np.pi * t)]
    ).T.copy(order="C")

wave_clip = AudioClip(make_frame, duration=2, fps=44100)
silenced_clip = wave_clip.multiply_volume(0, start=.5, end=1)
```

So just adding two parameters `start` and `end` to specify a range in which the volume transform would be applied for `multiply_volume` FX seems obvious to me.

mondeja:
This introduces the optional arguments `start_time` and `end_time` for `multiply_volume` FX. The volume multiplication, if these parameters are not `None`, is only applied to the seconds in the range from `start_time` to `end_time`. Note that after this, is a lot easier to introduce silences in a clip: `clip.multiply_volume(0, start_time=5, end_time=10)`. The code is totally backwards compatible. I've also properly documented the function and rewritten the tests parametrizing them.

> Closes #1571

### DIFF
--- a/moviepy/audio/fx/multiply_volume.py
+++ b/moviepy/audio/fx/multiply_volume.py
@@ -57,10 +57,12 @@ def multiply_volume(clip, factor, start=None, end=None):
             keep_duration=True,
         )
 
-    start = clip.start if start is None else start
-    end = clip.duration if end is None else end
-
     return clip.transform(
-        _multiply_volume_in_range(factor, start, end, clip.nchannels),
+        _multiply_volume_in_range(
+            factor,
+            clip.start if start is None else start,
+            clip.duration if end is None else end,
+            clip.nchannels,
+        ),
         keep_duration=True,
     )
--- a/moviepy/audio/fx/multiply_volume.py
+++ b/moviepy/audio/fx/multiply_volume.py
@@ -59,17 +59,7 @@ def multiply_volume(clip, factor, start=None, end=None):
     start = clip.start if start is None else start
     end = clip.duration if end is None else end
 
-    if hasattr(clip, "nchannels"):
-        nchannels = clip.nchannels
-    else:
-        first_frame = clip.get_frame(start)
-        nchannels = (
-            first_frame.shape[1]
-            if (hasattr(first_frame, "shape") and len(first_frame.shape) > 1)
-            else 1
-        )
-
     return clip.transform(
-        _multiply_volume_in_range(factor, start, end, nchannels),
+        _multiply_volume_in_range(factor, start, end, clip.nchannels),
         keep_duration=True,
     )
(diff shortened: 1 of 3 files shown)


moviepy/audio/fx/multiply_volume.py
```diff
@@ -1,22 +1,68 @@
...
+    def multiply_mono_volume(get_frame, t):
+        return np.multiply(get_frame(t), factors_filter(factor, t))
+
+    return multiply_mono_volume if nchannels == 1 else multiply_stereo_volume
 
 
 @audio_video_fx
-def multiply_volume(clip, factor):
+@convert_parameter_to_seconds(["start", "end"])
+def multiply_volume(clip, factor, start=None, end=None):
```
mondeja, line 24: I'm wondering about what would be the correct name for these new arguments `start` and `end`. In #1170 I've seen that you've renamed `t_start` and `t_end` by `start_time` and `end_time` @tburrows13, but for example `Clip` uses `start` and `end` property names instead of `start_time` and `end_time` (also the correspondient method `with_start`). So, this shouldn't be normalized in the codebase? What do you think about `start` and `end` argument names, it would be better to use `start_time` and `end_time` here?

tburrows13: Sorry, I missed this message earlier! Yeah, consistency is good. I'm not sure which is better though. I'd lean towards `start_time` instead of `start` but I haven't investigated to see which is used where, and it could be that in some usages `start` makes more sense.

mondeja: Thank you for the feedback, I would opt for `start_time` in this case to make it explicit :+1:

Issue #1396, opened 2021-01-01
moviepy.video.fx.all.scroll() FUNCTION unable to work properly


quest-X:
**PURPOSE**
> I want to make a line of Chinese scroll horizontally, reciprocating from left to right, but this function doesn't seem to work, I don’t know if my usage is wrong. Thank you in advance for your willingness to check my question :D

**THE CODE**
```
import moviepy
from moviepy.editor import *

screensize = (1280,720)


txt_clip = TextClip("你好，世界#！",font = 'SIMYOU.TTF',fontsize=100,color='white') 
txt_clip2 = TextClip("hello world",font = 'SIMYOU.TTF',fontsize=100,color='red')
txt_clip3 = moviepy.video.fx.all.scroll(txt_clip, h=200, w=300, x_speed=10, y_speed=20, x_start=30, y_start=40)

txt_clip = txt_clip.set_pos('top').set_duration(10)
txt_clip2 = txt_clip2.set_pos('center').set_duration(10)
txt_clip3 = txt_clip2.set_pos('bottom').set_duration(10)

cvc = CompositeVideoClip( [txt_clip, txt_clip2, txt_clip3], size=screensize)

cvc.write_videofile("chinese_test_9.mp4",fps = 24) 
```

**THE SCREENSHOT**
![image](https://user-images.githubusercontent.com/10896384/103434114-ccee1780-4c37-11eb-9c6a-8d047133ff9a.png)

PR #1591, opened 2021-05-26
Fix bug in scroll FX

ZxYuan:
Fix `x_max` and `y_max` in `scroll.py`


mondeja (approved):
Thank you @ZxYuan

> Link #1396

PR #1689, opened 2021-11-09
ImageMagick change detection from convert.exe to magick.exe

Nex4rius:
convert.exe only exists when you explicitly add it during installation -> magick.exe always exists

Dialog during installation (on win 10)


mondeja (changes requested):
Good change, but could you provide a fallback for old `convert.exe`? Just enclosing it into a `try: .. except Exception` statement after the the current `except Exception` should be enough.

Nex4rius:
Sure it now check for both.

mondeja (approved):
Great, thank you!

Issue #1682, opened 2021-11-01
Can't cut out a video more than once!


skullleeep:
I am trying to make a video editing program and I made it so that the user can input values for triming. The problem with this is that after the audio is written I get a "index out of bounds" error. I would appreciate it if someone could help me out here. I will type the code under "steps to reproduce".

#### Expected Behavior
The video should get trimmed 2 times

#### Actual Behavior
You get an "index out of bounds" error after the audio has been written

#### Steps to Reproduce the Problem
Just run this script:
```
from moviepy.editor import *

clip = VideoFileClip("test.mp4") # Video Length is 3 min
values = [[0, 30], [50, 70]]
for i in values:
    clip = clip.cutout(i[0], i[1])
    
clip.write_videofile("new.mp4")
```

#### Specifications

  - Python Version: 3.9.7
  - Moviepy Version: 1.0.3
  - Platform Name: Linux Mint
  - Platform Version: 20.2

OsaAjani:
Thank you for your contributions and for reporting issues in this repository. With the release of v2, which introduces significant changes to the codebase and API, we’ve reviewed the backlog of open PRs and issues. Due to the length of the backlog and the likelihood that many of these are either fixed or no longer applicable, we’ve made the decision to close all previous PRs and issues.

If you believe that any of these are still relevant to the current version or if you'd like to reopen a related discussion, please feel free to create a new issue or pull request, referencing the old one.

Thank you for your understanding and continued support!

PR #1705, opened 2021-12-30
Fix for issue #1682 ( audio cutout )

selentium:


keikoro:
@selentium You can link issues by using # followed by the issue number – doing so makes things easier for everyone, but particularly maintainers in the case of PRs. Also, if you remove the spaces from the square brackets which you ticked off (by putting an `x` between them) in the PR template, they'll show up as ticked boxes.

keikoro:
The file which originally caused the Code format check to fail has just been fixed. It'd make sense to rebase `master` so all the automated checks can run over this PR again.

PR #1781, opened 2022-05-30
Parse FFMPEG streams with square brackets

B3QL:
Hikvisions CCTV recordings use custom codec which causes FFMPEG
to return stream "value" in square brackets.

Original FFMPEG issue:
https://trac.ffmpeg.org/ticket/3566

Sample files:
https://samples.ffmpeg.org/ffmpeg-bugs/trac/ticket4182/


B3QL:
@mondeja can you take a look? This should be a quick one ;) Thanks!

B3QL:
> Could you handle it, please? Also, would you mind to add a CHANGELOG entry? Thanks!
Done! Should I add an extra test for it or extend the `test_stream_square_brackets` will be sufficient?

mondeja (approved):
Awesome, thank you! 👍🏼

PR #1783, opened 2022-06-01
Fix audio processing for streams with missing `audio_bitrate`

B3QL:
Set default `audio_bitrate` to `None`.

Similar to #1781 if the file is footage from Hikvision CCTV the bitrate is not available to read even if the `audio` stream is present.

https://user-images.githubusercontent.com/3527244/171369718-6feb18ef-29a5-49af-b5df-b1be03565916.mp4


B3QL:
@mondeja can you take a look? thanks!

mondeja (approved):
Great, thank you!

PR #1785, opened 2022-06-02
Fixing ffmpeg error with opus audio in mp4

reoccurcat:

I have fixed a bug with mp4 when ffmpeg does not clip the video due to opus support being experimental. My fix only adds a command flag to enable this support so the error does not occur. This is the error in the image below:

image

I have further added this flag to the other functions in the file so as to fix the problem elsewhere. I have not, however, tested these functions. The function that I originally added the fix to was the clipping one, although I assume the flag to work in the other commands.
Please make changes as needed. Thank you.

### DIFF
--- a/moviepy/video/io/ffmpeg_tools.py
+++ b/moviepy/video/io/ffmpeg_tools.py
@@ -19,6 +19,7 @@ def ffmpeg_movie_from_frames(filename, folder, fps, digits=6, bitrate='v'):
              "-i", os.path.join(folder,folder) + '/' + s,
              "-b", "%dk"%bitrate,
              "-r", "%d"%fps,
+             "-strict", "-2",
              filename]
     
     subprocess_call(cmd)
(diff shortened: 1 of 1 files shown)


mondeja (changes requested):
Please:

- Could you explain what does `-strict 2`?
- Would you mind to points the fix to the *master* branch?

reoccurcat:
> Could you explain what does `-strict 2`?
`-strict -2` allows experimental access to opus in MP4 support, as said by the ffmpeg error message:
`opus in MP4 support is experimental, add '-strict -2' if you want to use it.`

> Would you mind to points the fix to the _master_ branch?
I originally developed the fix for version 1.0.3 to be used in a project I also created a pull request on that uses this version. If need be I can port the changes to the master branch as well.

reoccurcat:
> We're only accepting changes against master, which will be included in the next major version.

Alright, thank you for letting me know.

> Do you know in which version of `ffmpeg` was this option added?

I do not know the exact version, but from what I can find, the option has been existent in the `ffmpeg` documentation for as long as 2013, maybe earlier, but the archives don't go any further than 2013.

PR #1837, opened 2022-10-07
Handle brackets and language in FFMPEG output

LoipesMas:
After #1781 we capture whatever is in square brackets as language. Newer ffmpeg versions can provide both the value in square brackets (whatever that is) and language in parenthesis, which breaks the regex.
I changed the regex to ignore the value in square brackets (using non-capturing group) and always get the language from the parenthesis.
All previous tests pass without changing, so I assume backwards compatibility is intact.

### DIFF
--- a/moviepy/video/io/ffmpeg_reader.py
+++ b/moviepy/video/io/ffmpeg_reader.py
@@ -435,9 +435,8 @@ class FFmpegInfosParser:
                 stream_number = int(stream_number)
                 stream_type_lower = stream_type.lower()
 
-                if language is not None:
-                    if language == "und":
-                        language = None
+                if language == "und":
+                    language = None
 
                 # start builiding the current stream
                 self._current_stream = {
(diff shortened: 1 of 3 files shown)


moviepy/video/io/ffmpeg_reader.py
```diff
@@ -435,7 +436,7 @@ def parse(self):
                 stream_type_lower = stream_type.lower()
 
                 if language is not None:
-                    if language == "und" or language.startswith("0x"):
+                    if language == "und":
```
mondeja, line 439: You can rewrite this just as `if language == "und"`

mondeja (changes requested):

moviepy/video/io/ffmpeg_reader.py
LoipesMas: You're right, thanks!

mondeja (approved):
Thank you for the patch!

Issue #1798, opened 2022-06-28
test_PR_529 fails in debian sid - possibly ffmpeg 5.0


emollier:
Greetings,

For information, there is an open [Debian Bug#1004642](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1004642) in which it has been noticed that MoviePy was failing its unit test with ffmpeg 5.0 (or some other change that might have appeared at the same time, but that sounds unlikely).

#### Expected Behavior
All unit tests should pass

#### Actual Behavior
The _test_PR_529_ fails with error:
```
|     def test_PR_529():
|         with VideoFileClip(SOURCE_PATH+"media/fire2.mp4") as video_clip:
| >           assert video_clip.rotation == 180
| E           assert 0 == 180
| E            +  where 0 = <moviepy.video.io.VideoFileClip.VideoFileClip object at 0x7ff9fd890880>.rotation
```

#### Steps to Reproduce the Problem
Run the test suite using:
```
python3.10 -m pytest tests
```

#### Specifications

  - Python Version: 3.10.5
  - Moviepy Version: 1.0.3
  - FFmpeg Version: 5.0.1
  - Platform Name: Debian
  - Platform Version: Sid

Thank you for your work on MoviePy!

Have a nice day,  :)\
Étienne.

emollier:
urgh, the debian build log is somewhat old, I should have done a test build myself earlier, I'm seeing a lot more unit test failures (including the _test_PR_529_) at the moment:
```
=========================== short test summary info ============================
FAILED tests/test_ImageSequenceClip.py::test_1 - TypeError: must be real numb...
FAILED tests/test_PR.py::test_PR_458 - TypeError: must be real number, not No...
FAILED tests/test_PR.py::test_PR_528 - TypeError: must be real number, not No...
FAILED tests/test_PR.py::test_PR_529 - assert 0 == 180
FAILED tests/test_VideoClip.py::test_write_image_sequence - TypeError: unsupp...
FAILED tests/test_VideoClip.py::test_subfx - TypeError: must be real number, ...
FAILED tests/test_VideoClip.py::test_oncolor - TypeError: must be real number...
FAILED tests/test_VideoClip.py::test_setaudio - TypeError: must be real numbe...
FAILED tests/test_VideoClip.py::test_setaudio_with_audiofile - TypeError: mus...
FAILED tests/test_VideoClip.py::test_setopacity - TypeError: must be real num...
[...]

PR #1872, opened 2022-11-30
Parse rotation from displaymatrix

LoipesMas:
Newer ffmpeg versions no longer provide `rotate` field in `Metadata` section. So I changed the parsing to get this value from `Side data`'s `displaymatrix` field. But that angle value is in opposite direction, however it's in-line with ffmpegs rotation. I.e. if we rotate video using ffmpeg with `rotate="90"`, old value would show `270`, while new value shows `90`. So it's potentially a breaking change (but IMO for the better). Also, I think this new value is in range `-180:180`, instead of `0:360`, not sure if that could cause other problems.
Older ffmpeg versions (4.4) do provide `displaymatrix` field as well, so this code works for both older and newer versions.

I renamed the test file and changed the test value, instead of changing the file contents, to avoid increasing repo size.

### DIFF
--- a/moviepy/video/io/ffmpeg_reader.py
+++ b/moviepy/video/io/ffmpeg_reader.py
@@ -504,7 +504,9 @@ class FFmpegInfosParser:
 
                 if self._current_stream["stream_type"] == "video":
                     if field == "displaymatrix":
-                        match = re.search(r"rotation of (-?\d+\.\d+) degrees", value.strip())
+                        match = re.search(
+                            r"rotation of (-?\d+\.\d+) degrees", value.strip()
+                        )
                         if match is not None:
                             rotation = float(match.groups()[0])
                             self.result["video_rotation"] = rotation
@@ -736,6 +738,7 @@ class FFmpegInfosParser:
         """
         raw_field, raw_value = line.split(":", 1)
         return (raw_field.strip(" "), raw_value.strip(" "))
+
 
 def ffmpeg_parse_infos(
     filename,
--- a/moviepy/video/io/ffmpeg_reader.py
+++ b/moviepy/video/io/ffmpeg_reader.py
@@ -503,9 +503,11 @@ class FFmpegInfosParser:
                 field, value = self.parse_metadata_field_value(line)
 
                 if self._current_stream["stream_type"] == "video":
-                    field, value = self.video_metadata_type_casting(field, value)
-                    if field == "rotate":
-                        self.result["video_rotation"] = value
+                    if field == "displaymatrix":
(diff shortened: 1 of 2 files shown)


B3QL:
@mondeja Can you take a look?
BTW the CI is failing on `linux (3.6)` due to Ubuntu 22.04 rollout on GitHub Actions.

Related issues:
https://github.com/actions/setup-python/issues/543
https://github.com/actions/runner-images/issues/6399

keikoro:
@B3QL Thanks for the info on the build failure.

@LoipesMas Which ffmpeg versions are we talking about here? You're only referencing 4.4 as "older" here, so it's unclear if this is applicable to other 4.x versions as well.

I'm asking because ffmpeg 5 seems to be the cause of what's discussed in #1798. The test that fails in the other issue is for the last PyPI release, i.e. MoviePy 1.0.3. Am I assuming correctly you are using the latest development version of MoviePy? Thanks.

mondeja:
Please, don't mention me again. I've said yet that I haven't publication rights, so I'm only reviewing those changes that could compromise my personal usage of moviepy, I haven't motivation to maintain this project.

That said, if we're going to accept support for FFMpeg v5 the CI should be updated to test against the both versions, right? Seems that the only one with this concerns are me but the library has 2.5 million downloads every month, that doesn't have sense.

B3QL:
> Please, don't mention me again. I've said yet that I haven't publication rights, so I'm only reviewing those changes that could compromise my personal usage of moviepy, I haven't motivation to maintain this project.
> 

OK, sorry for bothering you, I just assumed that you partially maintain the project based on the last few PRs that you accepted and were merged :)

keikoro:
@B3QL No worries, I didn't know either and I'm another maintainer for this repo, so...

@LoipesMas Thanks for the update!

The last release in this repo was [v2.0.0.dev2](https://github.com/Zulko/moviepy/releases/tag/v2.0.0.dev2) from October 2020. All PRs which were merged since got merged right into `master`. I'm assuming all releases so far were also created directly from master – instead of keeping changes for both upcoming releases and various "random" fixes contained to, say, one or more`dev` branches.

I'm saying I'm _assuming_ that's what happened because I wasn't involved in the process so far. I also didn't contribute to development toward MoviePy 2.0. @tburrows13 was the main person behind that, #1089 is the ticket that contains the most info on that.

What I do know with certainty is that I can't create new releases on PyPI, where the last release is 1.0.3 (!) and I believe that's the relevant thing here for Debian.

Regarding GitHub Actions, I guess FFMPEG 5 should get added to the CI tests as Álvaro pointed out earlier, and Python 3.10 probably as well if we want to test against what the new Debian release will come with.

OsaAjani:

PR #1885, opened 2022-12-28
Fixed MoviePy being displayed as Moviepy in some places.

EvieUwU:
This is a minor fix. It just changes any text that is stylised as Moviepy, and displayed to the user, to the proper capitalisation (MoviePy)


keikoro:
Thanks, @EvieUwU!

PR #1897, opened 2023-01-17
Fix #1895 printing `File Not Found` on Windows

### LINKED ISSUE #1895: File Not Found printed on the console after fresh install of pre-release
The package works, I can run all functionality but it prints something strange.
After installing the pre-release verion using pip install moviepy --pre or upgrading using pip install moviepy --pre --upgrade and from a simple script:

```
from moviepy.editor import *


def main() -> int:
    VideoClip()
    return 0


if __name__ == "__main__":
    main()
```

My console prints: "File Not Found"

#### Expected Behavior
Not to print anything

#### Actual Behavior
Prints "File Not Found"

Install the pre-release version using pip
Run the script below:
```
from moviepy.editor import *


def main() -> int:
    VideoClip()
    return 0


if __name__ == "__main__":
    main()
```
Check the console

#### Specifications

  - Python Version: 3.11.1
  - Moviepy Version: v2.0.0.dev2
  - Platform Name: Windows
  - Platform Version: 10

kifuan:

Fix #1895.

Stop using system commands and use `pathlib` to find the ImageMagick binary under `C:\Program Files\ImageMagick-*` instead.

### DIFF
--- a/moviepy/config.py
+++ b/moviepy/config.py
@@ -38,6 +38,37 @@ def try_cmd(cmd):
         return True, None
 
 
+def detect_imagemagick_for_windows():
+    """Detects imagemagick binary for Windows."""
+
+    # Try a few different ways of finding the ImageMagick binary on Windows
+    try:
+        key = wr.OpenKey(wr.HKEY_LOCAL_MACHINE, r"SOFTWARE\ImageMagick\Current")
+        result = wr.QueryValueEx(key, "BinPath")[0] + r"\magick.exe"
+        key.Close()
+        return result
+    except Exception:
+        pass
+
+    imagemagick_dir = [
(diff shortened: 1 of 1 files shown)


kifuan:
Several tests failed when installing dependencies for the same reason, likely caused by the version string.

image

As for the code reformatting, I think it should be fixed in another PR or somewhere else.

keikoro:
@kifuan Thanks for your PR. The black check errors should be fixed now.

mgaitan:
> As for the code reformatting, I think it should be fixed in another PR or somewhere else.

Fixed in #1943

kifuan:
ping @mgaitan @keikoro

PR #1906, opened 2023-01-30
Use audio_codec option even if audio is given as filename

zalgo3:

`moviepy.video.VideoClip.write_videofile` ignores `audio_codec` when `audio` is a file name. I have fixed it in this PR.

### DIFF
--- a/moviepy/video/io/ffmpeg_writer.py
+++ b/moviepy/video/io/ffmpeg_writer.py
@@ -44,6 +44,9 @@ class FFMPEG_VideoWriter:
     audiofile : str, optional
       The name of an audio file that will be incorporated to the video.
 
+    audio_codec : str, optional
+      FFMPEG audio codec.
+
     preset : str, optional
       Sets the time that FFMPEG will take to compress the video. The slower,
       the better the compression rate. Possibilities are: ``"ultrafast"``,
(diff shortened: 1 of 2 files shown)


keikoro:
Thanks for your PR.

I think I might nab just the black formatting fixes for now to prevent our checks from failing (even though they don't even fail consistently...).

moviepy/video/io/ffmpeg_writer.py
```diff
@@ -81,6 +84,7 @@ def __init__(
         fps,
         codec="libx264",
         audiofile=None,
+        audio_codec=None,
```
mgaitan, line 87: why not default to "copy" directly here ?

zalgo3: Thanks for your question! The default value of audio_codec in VideoClip.write_videofile is None, so I also adapted FFMPEG_VideoWriter to that.
https://github.com/Zulko/moviepy/blob/e7cbfd38615fa265e94b20a47e4d21fc473001ed/moviepy/video/VideoClip.py#L209

This pull request intends that in VideoClip.write_videofile, if a filename is given for the audio argument and a codec name is given for audio_codec, then it will be encoded according to that codec, but if audio If audio_codec=None, I thought it would be better to have it "copied" as before.

OsaAjani:

PR #1914, opened 2023-02-05
Fix GitHub workflow badge

EFord36:
The build badge is currently broken:

Screenshot 2023-02-05 at 17 48 55

This PR fixes it, so it looks like:

Screenshot 2023-02-05 at 17 49 34

In summary, this badge was changed in a breaking fashion such that the badge was just linking to the github issue rather than showing the data. Updating the url resolves this.

See the linked GitHub issue from the previous version of the badge for further details: badges/shields#8671

I also updated the link for the badge, so it matches the behaviour of the badge more closely (matching on the workflow file, rather than the name).

I don't think any of the PR template check boxes applies, since this is a doc-only change - but hopefully the screenshots above are evidence of the 'fix'.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,8 +7,8 @@ MoviePy
 .. image:: https://img.shields.io/gitter/room/movie-py/gitter?color=46BC99&logo=gitter
     :target: Gitter_
     :alt: Discuss MoviePy on Gitter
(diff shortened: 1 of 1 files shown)


keikoro:
Thank you, @EFord36!

PR #1932, opened 2023-03-13
Adding default behaviour for clips array when input shape isn't adequate

EmreCenk:
**This PR Includes:**
- Implemented changes for clips_array default behaviour. Previously when a non-rectangular array was passed in ( ex: [ [1, 2], [3] ] ), it would throw an error. Now, the remaining space is just filled with the background color as default behaviour instead

**Checklist:**
- [X] I have properly documented new or changed features in the documentation or in the docstrings
- [X] I have properly explained unusual or unexpected code in the comments around it


mgaitan:
@EmreCenk thank you very much for this first contribution. Please, consider fixing the broken tests and add at least one showing the case you are fixing with this change. Also, could you use a spellchecker in the docstring and comments?

thanks!

OsaAjani:

PR #1943, opened 2023-04-01
Fix CI. 

mgaitan:
- Dropped Python 3.6 as it reached its end of life 15 months ago and it's no supported 
- Start running CI with python 3.7 to 3.11 in every environment
- Fix `__version__` parsing in `setup.py` to trim and extra ´\r` in windows that make the package uninstallable in Windows.


mgaitan:
I gonna merge this to make master a bit more robust.

keikoro:
Thx.

Issue #1920, opened 2023-02-14
ValueError: cannot reshape array of size 777600 into shape (720,1080,5)


samuelbraun04:
The following lines:
```
currentImageClip = ImageClip(r'path/to/image')
imageClipWithCaption = CompositeVideoClip([ColorClip(size=(1080, 720), color='black'), currentImageClip])
```
gives me the following error:
```
Traceback (most recent call last):
  File "C:\Users\samlb\Documents\folder\pythonFile.py", line 140, in makeVideo
    imageClipWithCaption = CompositeVideoClip([ColorClip(size=(1080, 720), color='black'), currentImageClip])
  File "C:\Users\samlb\AppData\Roaming\Python\Python39\site-packages\moviepy\video\VideoClip.py", line 1157, in __init__
    np.tile(color, w * h).reshape(shape), is_mask=is_mask, duration=duration
ValueError: cannot reshape array of size 777600 into shape (720,1080,5)
```
How can I avoid this error? Does it have something to do with the type of image I'm trying to composite?

This is the image I'm using (.png):

![16763856438648417](https://user-images.githubusercontent.com/90664891/218770737-8b352ee2-c0d4-4d7d-9b18-3f80b87a0ea7.png)

#### Specifications

  - Python Version: 3.9.13
  - MoviePy Version: 2.0.0.dev2-
  - Platform Name: Windows
  - Platform Version: 10

keikoro:
@stef405 Generally, any issue is up for grabs, but best to check older open PRs first to avoid potential duplication of efforts. Open issues regarding the v2 release are also worth a look because parts of the codebase are being rewritten.

stef405:
@keikoro  Is there a seperate github repository for v2 release?

stef405:
So far, I have pulled a recent version of the repository, and used the same snippet code and image in the original post, but no errors popped up. I don't know if this is specifically a problem with older versions of Windows, which may or may not be relevant for project maintenance going forward

s-2ndd:
Update: Was able to recreate error on v2, and will continue to figure out this bug

stef405:
Update: Was able to reproduce bug, and will work on trying to figure it out

stef405:
[...]

PR #1952, opened 2023-04-17
Added code to raise exception with string

stef405:
I added two lines to VideoClip.py that raises an exception if a user passes in a string as the color parameter. This is in reaction to this [Bug](https://github.com/Zulko/moviepy/issues/1920). The problem the original user faced was that it raised a ValueError. The user inputted 'black' as the color in ColorClip, which is not an acceptable parameter for the function. Instead, I just created some code that checks if color is a string, and raises an exception that mentions the wrong user input. In the future, a user should hopefully realize quicker that they passed in incorrect input with the Exception message, instead of not knowing why a ValueError was thrown later on. 

To test_color() in test_VideoClip.py, I added a test that attempts to create a Clip with color = 'black'. The test passes, since an exception should be thrown when color is passed in as a string.

I also checked for PEP8 conventions in both files, and none of the code I added resulted in a style complaint. 

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -1149,6 +1149,10 @@ class ColorClip(ImageClip):
                 color = (0, 0, 0)
             elif not hasattr(color, "__getitem__"):
                 raise Exception("Color has to contain RGB of the clip")
+            elif isinstance(color, str):
+                raise Exception(
+                    "Color cannot be string. Color has to contain RGB of the clip"
+                )
             shape = (h, w, len(color))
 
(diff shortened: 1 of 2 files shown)


keikoro:
@stef405 I fixed the code for the checkboxes and moved it to the end of your PR description. They are meant to be ticked off to give an idea of what was done at a quick glance.

keikoro:
Looks like the formatting needs some more work.

stef405:
I combined the commits. There should now be only one commit associated with this PR

mgaitan (approved):

PR #1956, opened 2023-04-25
Use imageio.v3 when read or write images

### LINKED ISSUE #1894: Recommend to use imageio.v3 instead of imageio
I would like to recommend to use `imageio.v3` instead of `imageio` when just import `imread` from `imageio`, specifically in such two position:
https://github.com/Zulko/moviepy/blob/858bb81fba7e09f1f562283ed6d1394db883b6c7/moviepy/video/VideoClip.py#L14
https://github.com/Zulko/moviepy/blob/0751d072d9b61be3466f39184beba004d61104f1/moviepy/video/io/ImageSequenceClip.py#L8

This is because `imageio.v3.imread` support more image format than `imageio.imread` such as `.webp`. And if necessary I would like to submit a PR.

xunmeibuyue:
Using `imageio.v3` instead of `imageio` when read or write images, which has the advantage that `imageio.v3` support more image format than `imageio` such as `.webp`.

cumented new or changed features in the documentation or in the docstrings

Fix #1894

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -11,7 +11,7 @@ import tempfile
 
 import numpy as np
 import proglog
-from imageio.v3 import imread, imwrite
+import imageio.v3 as iio
 from PIL import Image
 
 from moviepy.Clip import Clip
@@ -190,7 +190,7 @@ class VideoClip(Clip):
         else:
             im = im.astype("uint8")
 
-        imwrite(filename, im)
+        iio.imwrite(filename, im)
 
     @requires_duration
     @use_clip_fps_by_default
@@ -1040,7 +1040,7 @@ class ImageClip(VideoClip):
 
         if not isinstance(img, np.ndarray):
             # img is a string or path-like object, so read it in from disk
-            img = imread(img, index=0)
+            img = iio.imread(img, index=0)
 
         if len(img.shape) == 3:  # img is (now) a RGB(a) numpy array
             if img.shape[2] == 4:
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -1040,7 +1040,7 @@ class ImageClip(VideoClip):
 
         if not isinstance(img, np.ndarray):
             # img is a string or path-like object, so read it in from disk
-            img = imread(img)
+            img = imread(img, index=0)
 
         if len(img.shape) == 3:  # img is (now) a RGB(a) numpy array
             if img.shape[2] == 4:
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -11,7 +11,7 @@ import tempfile
 
 import numpy as np
 import proglog
-from imageio import imread, imsave
+from imageio.v3 import imread, imwrite
 from PIL import Image
 
 from moviepy.Clip import Clip
@@ -190,7 +190,7 @@ class VideoClip(Clip):
         else:
             im = im.astype("uint8")
 
-        imsave(filename, im)
+        imwrite(filename, im)
 
     @requires_duration
     @use_clip_fps_by_default
--- a/moviepy/video/io/ImageSequenceClip.py
+++ b/moviepy/video/io/ImageSequenceClip.py
@@ -5,7 +5,7 @@ of image files.
 import os
 
 import numpy as np
-from imageio import imread
+from imageio.v3 import imread
 
 from moviepy.video.VideoClip import VideoClip
 
--- a/moviepy/video/io/ImageSequenceClip.py
+++ b/moviepy/video/io/ImageSequenceClip.py
@@ -5,7 +5,7 @@ of image files.
 import os
 
 import numpy as np
-from imageio.v3 import imread
+import imageio.v3 as iio
 
 from moviepy.video.VideoClip import VideoClip
 
@@ -62,7 +62,7 @@ class ImageSequenceClip(VideoClip):
         if isinstance(sequence, list):
             if isinstance(sequence[0], str):
                 if load_images:
-                    sequence = [imread(file, index=0) for file in sequence]
+                    sequence = [iio.imread(file, index=0) for file in sequence]
                     fromfiles = False
                 else:
                     fromfiles = True
@@ -78,14 +78,14 @@ class ImageSequenceClip(VideoClip):
 
         # check that all the images are of the same size
         if isinstance(sequence[0], str):
-            size = imread(sequence[0], index=0).shape
+            size = iio.imread(sequence[0], index=0).shape
         else:
             size = sequence[0].shape
 
         for image in sequence:
             image1 = image
             if isinstance(image, str):
-                image1 = imread(image, index=0)
+                image1 = iio.imread(image, index=0)
             if size != image1.shape:
                 raise Exception(
                     "MoviePy: ImageSequenceClip requires all images to be the same size"
@@ -117,12 +117,12 @@ class ImageSequenceClip(VideoClip):
                 index = find_image_index(t)
 
                 if index != self.last_index:
-                    self.last_image = imread(self.sequence[index], index=0)[:, :, :3]
+                    self.last_image = iio.imread(self.sequence[index], index=0)[:, :, :3]
                     self.last_index = index
 
                 return self.last_image
 
-            if with_mask and (imread(self.sequence[0], index=0).shape[2] == 4):
+            if with_mask and (iio.imread(self.sequence[0], index=0).shape[2] == 4):
                 self.mask = VideoClip(is_mask=True)
                 self.mask.last_index = None
                 self.mask.last_image = None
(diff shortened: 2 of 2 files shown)


mgaitan:
@xunmeibuyue for the next time, there is no need to open a new PR for each update after a review comment. Simply you can push new commit to the same remote branch you pushed. In that way the context of the review will keep in a single place. 

Regarding the PR itself, we mandatorily need all the CI builds passing. So, could you review what have broken and fix it?  You can follow the CONTRIBUTING.md doc to install the dependencies and run the test locally before push it. 

I look forward to your updates

xunmeibuyue:
OK, I opened a new PR just for a cleaning commit, but it seems that keeping the context of the review in a single place will be more helpful.

I can't believe a change of just few lines arises so much errors :sob:, I will review and fix it later.

moviepy/video/VideoClip.py
```diff
@@ -1040,7 +1040,7 @@ def __init__(
 
         if not isinstance(img, np.ndarray):
             # img is a string or path-like object, so read it in from disk
-            img = imread(img)
+            img = imread(img, index=0)[0]
```
FirefoxMetzger, line 1043:
```suggestion
            img = imread(img, index=0)
```

If you add a slice to the call you will get the first row of the image. I suppose you meant to just get the first frame/image in the file?

moviepy/video/VideoClip.py
```diff
@@ -11,7 +11,7 @@
 
 import numpy as np
 import proglog
-from imageio import imread, imsave
+from imageio.v3 import imread, imwrite
```
FirefoxMetzger, line 14: While not strictly necessary, it is recommended to import Imageio as `import imageio.v3 as iio`. This gives you access to the full API and reads nicer than only importing the methods used.

You would then consume it as `iio,imwrite` and `iio.imread` respectively.

moviepy/video/io/ImageSequenceClip.py
```diff
@@ -5,7 +5,7 @@
 import os
 
 import numpy as np
-from imageio import imread
+from imageio.v3 import imread
```
FirefoxMetzger, line 8:
```suggestion
import imageio.v3 as iio
```

Same here, I would recommend to import the entire namespace. This will give you access to some niceties (see below).

xunmeibuyue:
Sorry, I just try to commit a change (follow the guidance of @FirefoxMetzger ) on the fork of mine, but it was wrongly submitted to here. Hence there is no need to run the actions @mgaitan and I'll update it once I finished the local test.

moviepy/video/io/ImageSequenceClip.py
```diff
@@ -62,7 +62,7 @@ def __init__(
         if isinstance(sequence, list):
             if isinstance(sequence[0], str):
                 if load_images:
-                    sequence = [imread(file) for file in sequence]
+                    sequence = [imread(file, index=0)[0] for file in sequence]
```
FirefoxMetzger, line 65:
```suggestion
                    sequence = [img for img in iio.imiter(file)]
```

We now have [`imiter`](https://imageio.readthedocs.io/en/stable/_autosummary/imageio.v3.imiter.html#imageio.v3.imiter) which can iterate over containers and give you images in the order in which they appear in the file. 

This not only reads nicer, but is also more performant because (1) you only need to open the file once, and (2) you tell ImageIO "I want to read more than one image, just not all at once". This way we can optimize for this scenario and do things like buffering future frames/images, carry decoder state over to the next frame, select more optimized threadding models, etc. In a nutshell: we can optimize for reading and serving images sequentially.


That said, it seems odd to read all images and then store them in a list, because it can be memory hungry. Where is `sequence` consumed? Perhaps it is better to move the generator there and load images one at a time.

moviepy/video/io/ImageSequenceClip.py
```diff
@@ -78,14 +78,14 @@ def __init__(
 
         # check that all the images are of the same size
         if isinstance(sequence[0], str):
-            size = imread(sequence[0]).shape
+            size = imread(sequence[0], index=0)[0].shape
```
FirefoxMetzger, line 81:
```suggestion
            size = iio.improps(sequence[0], index=0).shape
```

Another advantage of using the full namespace is that you get [`improps`](https://imageio.readthedocs.io/en/stable/_autosummary/imageio.v3.improps.html#imageio.v3.improps) which gives you standardized metadata (including the shape of the resulting image) without decoding pixels (if possible).

This needs less memory and is _much_ faster since we only need to read a small header instead of a potentially large image.

moviepy/video/io/ImageSequenceClip.py
```diff
@@ -78,14 +78,14 @@ def __init__(
...
         else:
             size = sequence[0].shape
 
         for image in sequence:
             image1 = image
             if isinstance(image, str):
-                image1 = imread(image)
+                image1 = imread(image, index=0)[0]
```
FirefoxMetzger, line 88:
```suggestion
                image1 = imread(image, index=0)
```

moviepy/video/io/ImageSequenceClip.py
```diff
@@ -117,20 +117,20 @@ def make_frame(t):
                 index = find_image_index(t)
 
                 if index != self.last_index:
-                    self.last_image = imread(self.sequence[index])[:, :, :3]
+                    self.last_image = imread(self.sequence[index], index=0)[0][:, :, :3]
```
FirefoxMetzger, line 120:
```suggestion
                    self.last_image = imread(self.sequence[index], index=0)[:, :, :3]
```

This looks a bit like we can push down the entire operation to ImageIO, but I lack context of the class itself. Does `last_image` mean that you wish to read the last image in the file?

moviepy/video/io/ImageSequenceClip.py
```diff
@@ -117,20 +117,20 @@ def make_frame(t):
...
                     self.last_index = index
 
                 return self.last_image
 
-            if with_mask and (imread(self.sequence[0]).shape[2] == 4):
+            if with_mask and (imread(self.sequence[0], index=0)[0].shape[2] == 4):
```
FirefoxMetzger, line 125:
```suggestion
            if with_mask and iio.improps(self.sequence[0], index=0).shape[2] == 4:
```

moviepy/video/io/ImageSequenceClip.py
```diff
@@ -117,20 +117,20 @@ def make_frame(t):
...
                 self.mask = VideoClip(is_mask=True)
                 self.mask.last_index = None
                 self.mask.last_image = None
 
                 def mask_make_frame(t):
                     index = find_image_index(t)
                     if index != self.mask.last_index:
-                        frame = imread(self.sequence[index])[:, :, 3]
+                        frame = imread(self.sequence[index], index=0)[0][:, :, 3]
```
FirefoxMetzger, line 133:
```suggestion
                        frame = imread(self.sequence[index], index=0)[:, :, 3]
```

FirefoxMetzger (commented):
Some quick comments - and suggestions on where new API functions might be more useful :)

FirefoxMetzger:
> Sorry, I just try to commit a change (follow the guidance of @FirefoxMetzger ) on the fork of mine, but it was wrongly submitted to here.

@xunmeibuyue That's no problem at all. It's actually best practice to commit all your changes here even if the result is an intermediate or broken state. There is no judgement here (we all know that coding usually works in multiple steps) and having a more fine-grained history of commits makes it much easier to roll back to earlier work should the need ever arise.

Also, GitHub will show us (reviewers) a summary of all your changes so when we come back for a review, we see the final result and not the steps in-between.

(There are some repos like numba (JIT compiler for python code), where this practice is discouraged, but that is because they have a massive CI matrix of dozens of GPUs, CPUs, OSes, and python versions they need to test against. Repos like that will tell you in big BOLD letters that you shouldn't push everything, so unless you find something like that it's usually okay to develop against the PR branch ;) )

xunmeibuyue:
> > Sorry, I just try to commit a change (follow the guidance of @FirefoxMetzger ) on the fork of mine, but it was wrongly submitted to here.
> 
> @xunmeibuyue That's no problem at all. It's actually best practice to commit all your changes here even if the result is an intermediate or broken state. There is no judgement here (we all know that coding usually works in multiple steps) and having a more fine-grained history of commits makes it much easier to roll back to earlier work should the need ever arise.
> 
> Also, GitHub will show us (reviewers) a summary of all your changes so when we come back for a review, we see the final result and not the steps in-between.
> 
> (There are some repos like numba (JIT compiler for python code), where this practice is discouraged, but that is because they have a massive CI matrix of dozens of GPUs, CPUs, OSes, and python versions they need to test against. Repos like that will tell you in big BOLD letters that you shouldn't push everything, so unless you find something like that it's usually okay to develop against the PR branch ;) )

I just force-pushed a new branch before seeing your suggestion 😭, because I‘m a little bit neat freak on coding.  I am really sorry. Thanks for the kind suggestion!

xunmeibuyue:
> Some quick comments - and suggestions on where new API functions might be more useful :)

The latest commit, which is based on the comments of @FirefoxMetzger, seems passing the pytest.

moviepy/video/VideoClip.py
xunmeibuyue: Thanks for pointing out. I'll push a new commit

keikoro:
@xunmeibuyue Looks like there were issues with the code formatting.

OsaAjani:

PR #1969, opened 2023-05-09
fix none error

gaotong2055:


moviepy/video/tools/subtitles.py
```diff
@@ -176,7 +176,8 @@ def file_to_subtitles(filename, encoding=None):
             if times:
                 current_times = [convert_to_seconds(t) for t in times]
             elif line.strip() == "":
-                times_texts.append((current_times, current_text.strip("\n")))
+                if current_times:
```
mgaitan, line 179: thanks! could you add a test?

PR #1998, opened 2023-06-22
Add warning to README.rst

B3QL:
Hi folks,
since nothing much had happened from the beginning of 2023
in terms of finishing version 2.0 or building a team around the project.

I see no choice but to warn other developers to not depend on MoviePy.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,10 @@
 MoviePy
 =======
+**WARNING!**
+
+The MoviePy project is no longer maintained. More information can be found here https://github.com/Zulko/moviepy/issues/1874
+
+**DO NOT USE FOR PRODUCTION**
 
 .. image:: https://badge.fury.io/py/moviepy.svg
     :target: PyPI_
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,10 @@ MoviePy
 =======
 **WARNING!**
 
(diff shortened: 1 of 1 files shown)


B3QL:
@keikoro can you take a look?

keikoro:
> @keikoro can you take a look?

I wouldn't agree it's "not maintained" – I'm still labelling issues here. ^^ But agree that a warning makes sense.

Maybe you could change it to something that doesn't make it sound like it's **dead dead**, but lets users know that what's on PyPI is outdated and that the project is struggling with releasing a new version? Otherwise people will be confused why the repo wasn't also set to archived.

B3QL:
Hey @keikoro, you're right. The message was crafted before the rebirth of the discussion. I'll reword it next week when I will return from holidays. Sorry for late response :)

B3QL:
@keikoro I tried to make it more informative and less scary, let me know if it's OK :)

B3QL:
@keikoro depending on https://github.com/Zulko/moviepy/issues/1874#issuecomment-1652724032 the PR might not be needed :) I would love to close it, but let's wait.

B3QL:
Closing in favour of https://github.com/Zulko/moviepy/pull/2024

Issue #2002, opened 2023-07-05
module 'PIL.Image' has no attribute 'ANTIALIAS' - resize.py


yoichitgy:
#### Expected Behavior

`resize` function works without an error when Pillow version 10.0.0 is installed.

#### Actual Behavior

The following error is raised at `moviepy/video/fx/resize.py", line 37, in resizer`

```
AttributeError: module 'PIL.Image' has no attribute 'ANTIALIAS'
```

#### Steps to Reproduce the Problem

1. Re-install MoviePy. As its dependency, Pillow version 10.0.0 is installed.
2. Call `resize` function on a `ImageClip`.

This issue is caused by the deprecation of `ANTIALIAS` in Pillow version 10.0.0.

Reference:
https://stackoverflow.com/a/76616129/4522678

#### Specifications

  - Python Version: 3.11
  - MoviePy Version: 1.0.3
  - Platform Name: macOS
  - Platform Version: 10.15.7

Mayank19j:
What should I do in my local to fetch the fix?

motishaku:
> What should I do in my local to fetch the fix?

did you figure it out?

delans:
I'm having the same error: 

Python Version: 3.11.5
MoviePy Version: 1.0.3
Pillow Version: 10.0.0
Platform Name: macOS
Platform Version: 12.6.7

bwang514:
Had the same issues.  

After digging into the source code (https://github.com/Zulko/moviepy/blob/master/moviepy/video/fx/resize.py)

I found that moviepy will try to use opencv as resizer first, if opencv not installed it will use PIL which leads to the error we are seeing here. 

Instead of fixing the deprecated PIL package, simply installing opencv-python works for me.

Kaszanas:
> Had the same issues.
> 
> After digging into the source code (https://github.com/Zulko/moviepy/blob/master/moviepy/video/fx/resize.py)
> 
> I found that moviepy will try to use opencv as resizer first, if opencv not installed it will use PIL which leads to the error we are seeing here.
> 
> Instead of fixing the deprecated PIL package, simply installing opencv-python works for me.

Then I suspect that this means that moviepy will add `opencv-python` as a dependency in future releases?

COROYO:
> > Had the same issues.
> > After digging into the source code (https://github.com/Zulko/moviepy/blob/master/moviepy/video/fx/resize.py)
[...]

PR #2003, opened 2023-07-05
Fix the issue caused by deprecation of `ANTIALIAS` in Pillow version 10.0.0

yoichitgy:
Fixed Issue #2002.

Reference:
https://stackoverflow.com/a/76616129/4522678

I removed checkboxes because I supposed non of them were applicable.

(Thank you for making this amazing library in open source!😃)

### DIFF
--- a/moviepy/video/fx/resize.py
+++ b/moviepy/video/fx/resize.py
@@ -37,7 +37,7 @@ def _get_PIL_resizer():
         #     newshape = (new_size[0], new_size[1])
 
         pil_img = Image.fromarray(pic)
-        resized_pil = pil_img.resize(new_size[::-1], Image.ANTIALIAS)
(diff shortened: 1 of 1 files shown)


simonbarker:
+1 for getting this done and merged in as soon as possible please 😀

tburrows13:
Thanks, it looks like LANCZOS was added in pillow 2.7.0 (from 2015), so probably don't need to consider minimum version constraints.

jwbats:
I need this fix. Is this coming out in a new release soon?

kevinlinxc:
When will another Pypi release happen?

muhammad-mobeen:
You guys need to update this to PyPi fast! I was going to add this issue again!

tburrows13:
I’m afraid that a PyPI release is unlikely to happen any time soon - see #1874

PR #2010, opened 2023-07-12
Fix version parsing in setup.py

vanmatten:
The last character of the version is cut out by `[:-1]` in `Path("moviepy/version.py").read_text().strip().split('"')[1][:-1]`.
Eg: `2.0.0.dev2` become `2.0.0.dev`, which Python build reads as `2.0.0.dev0`.

In fact `Path("moviepy/version.py").read_text().strip().split('"')` gives `['__version__ = ', '2.0.0.dev2', '']` so removing the last character isn't needed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ if "build_docs" in sys.argv:
 
     cmdclass["build_docs"] = BuildDoc
 
-__version__ = Path("moviepy/version.py").read_text().strip().split('"')[1][:-1]
+__version__ = Path("moviepy/version.py").read_text().strip().split('"')[1]
 
 
 # Define the requirements for specific execution needs.


keikoro:
I'm not sure why this was closed/deleted, I can only see that tests failed (not why), did this not solve the problem after all?

vanmatten:
Tests failed as in the main branch when I created the commit and I don't know why.
This commit do exactly what I have written in the title, I closed it because the pull request of v2.0 fix this issue as others.

keikoro:
Thanks for the info.

>  I closed it because the pull request of v2.0 fix this issue as others

If another PR fixes the issue anway, there is indeed no need to reopen this one. Thanks, in any case!

PR #2048, opened 2023-10-07
Support for specific workflows and automation for various movie genres in moviepy?

xanmankey:
This is a follow up on https://github.com/Zulko/moviepy/discussions/2046#discussion-5701521. It's an example of how I envisioned templates, sections, and automation using moviepy. I'm not certain as to whether the actual implementation belongs in a forked version of the library or the main library itself, but I suppose it could be implemented in a fork and then merged over if successful. The basic idea is as follows:

In the /workflows directory:
- Automater class: an abstract class that can be implemented to determine the Sections that a video should be divided into
- Section class: an abstract class that can be implemented to determine the Template that the sectioned videos should use
- BaseTemplate class: a parent class for (eventually) providing more advanced video layouts (currently I only abstracted support for orientation and effects, but there's so much that could be extended)

In the /workflows/automaters directory:
- Implementations of Automater (an unimplemented version, MusicAutomater, can be found, which ideally will eventually be implemented to account for music videos. In the case of music videos, sectioning would be done using the ```librosa``` library and by obtaining waveform information such as energy, beat, and harmonics over time)

In the /workflows/sections
Implementations of Section (an unimplemented version, MusicSection, can be found, which will ideally eventually be implemented to determine different orientations depending on factors such as the number of videos in the section, the most prominent waveform/the extent of the most prominent waveform, and various other factors)

In the /workflows/templates directory:
- EqualDivision.py (unimplemented instantiation of a BaseTemplate)
- PictureInPicture.py (unimplemented instantiation of a BaseTemplate)
- Title.py (contains an unimplemented extension on BaseTemplate, as well as an unimplemented instantiation of that BaseTemplate)

I think if it were to be implemented in this way, it would be accommodating for contribution to different genres of videos as well as different workflows without diluting the core implementation of the library. I'm not sure where the community or maintainers stand on this, which is why I left things unimplemented in places; I think this kind of support could really bolster Moviepy as a library and help contribute to reducing overhead in various video genres, but it would be a fairly large addition and thus should be open to debate. I hope this at least gets the ball rolling though and can help dictate what the next steps should be!

### DIFF
--- a/moviepy/workflows/Automater.py
+++ b/moviepy/workflows/Automater.py
@@ -1,0 +1,23 @@
+import abc
+
+# An abstract base class for automation
+# The idea is you can create custom Automater classes that take videos as parameters
+# and implement a method where you can automate the splitting of your video into sections
+class Automater(abc.ABC): 
+    def __init__(self, videos):
+        self.videos = videos # Your list of full video clips
+
+    # A method for the user to implement
+    # Provides time data to automates the splitting of your video into sections
+    # using the sectionVideos function
+    @abc.abstractmethod
+    def determineSections(self):
+        pass
+
+    # Returns a list of sectioned videos dependent on the output of determineSections
+    def sectionVideos(videoTimeData):
+        videos = []
+        # Iterate over key values
+        for video, time in videoTimeData: 
+            videos.append(video.cutout(time[0], time[1])) # Cut the video out based on the time interval
+        return videos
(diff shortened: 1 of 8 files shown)


keikoro:
This is just a heads-up that no-one really monitors the discussion board, so I'm not sure anyone has seen your post there. 

We (admins) are also really low on resources, so I can't tell you when (if) anyone will have time to read through what sounds like a proposal for a new feature (?!) – sorry. |:

OsaAjani:

PR #2070, opened 2023-11-19
Fix deprecated PIL resizing

Alex-Carter01:
Image.ANTIALIAS is deprecated.
Replacing with the suggested replacement Image.Resampling.LANCZOS

see https://github.com/mpetroff/pannellum/issues/1119

### DIFF
--- a/moviepy/video/fx/resize.py
+++ b/moviepy/video/fx/resize.py
@@ -37,7 +37,7 @@ def _get_PIL_resizer():
         #     newshape = (new_size[0], new_size[1])
 
         pil_img = Image.fromarray(pic)
-        resized_pil = pil_img.resize(new_size[::-1], Image.LANCZOS)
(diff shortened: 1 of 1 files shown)


Alex-Carter01:
Hi all. I also am reading from #1874 that the project could use some maintenance help - please let me know what I can do to help.

keikoro:
The first sentence in the linked PR says (emphasis mine):

> When using **recent** Pillow, a deprecation warning occurs

Is MoviePy currently using this particular version of Pillow? Or is this a fix that will only become relevant in the future?

keikoro:
I just found PR  #2072 which links to a commit (not a PR) which looks like it's solving a similar – or potentially the same – issue, though the other user's solution looks different.

keikoro:
Ok, found it, was apparently already merged: #2003 – is your fix another fix on top of that one?

OsaAjani:
Fix with v2.0

Issue #2049, opened 2023-10-09
Resize gives wrong shaped output


EsbernTK:
#### Expected Behavior
I have a video file with frame size width: 4480 height: 3096, and i am doing a resize of (224, None)
The expected resized frame would be (224, 324)

#### Actual Behavior
The actual frame i get back is (223, 324), and it is caused by the integer conversion on this [line](https://github.com/Zulko/moviepy/blob/bc8d1a831d2d1f61abfdf1779e8df95d523947a5/moviepy/video/io/ffmpeg_reader.py#L52)
The ratio calculation resolves to 0.07235142118863049 which when multiplied with 3096 gives 223.99999999999997 and is thus rounded down to 223 by the integer conversion.
It would be lovely if the shape calculation forced the given axis to be the same as the requested shape, instead of recalculating it using the ratio calculation which is prone to error.

#### Steps to Reproduce the Problem

```
from moviepy import *
path = "test.mp4" #test.mp4 is a video with shape (3096, 4480, 3)
video = VideoFileClip(path, target_resolution=(224, None))
frame = video.get_frame(0)
frame.shape #frame.shape = (223, 324, 3)
```

#### Specifications

  - Python Version: 3.9
  - MoviePy Version: 1.0.3
  - Platform Name: Windows 10
  - Platform Version: 22H3

Dotrar:
I added a little change to fix but it raises a few questions on how should we cater to this case.

Should we:
1. leave as is - because if you want a specific resolution, you should specify it completely.
2. use the target dimension as specified, calculate the _other_. and/or
3. use proper rounding functions? perhaps use `ceil` or at the very least round even? 

One could argue that if a result comes to " 12.2 pixels" (or in the example case 223.9999) then it should be `ceil`'d so that we have a pixel to cover that extra little bit that should be there.

But on the whole, I would expect that if i said "I want target height to be 100" it should come out as 100.

OsaAjani:
Thank you for your contributions and for reporting issues in this repository. With the release of v2, which introduces significant changes to the codebase and API, we’ve reviewed the backlog of open PRs and issues. Due to the length of the backlog and the likelihood that many of these are either fixed or no longer applicable, we’ve made the decision to close all previous PRs and issues.

If you believe that any of these are still relevant to the current version or if you'd like to reopen a related discussion, please feel free to create a new issue or pull request, referencing the old one.

Thank you for your understanding and continued support!

PR #2093, opened 2024-01-12
Fix resize func for https://github.com/Zulko/moviepy/issues/2049

Dotrar:
Fixes https://github.com/Zulko/moviepy/issues/2049

Sorry - I couldn't get precommit to work without the first commit in my PR, feel free to drop that if you choose to merge.

From issue:
> I added a little change to fix but it raises a few questions on how should we cater to this case.

> Should we:
1. leave as is - because if you want a specific resolution, you should specify it completely.
2. use the target dimension as specified, calculate the _other_. and/or
3. use proper rounding functions? perhaps use `ceil` or at the very least round even? 

> One could argue that if a result comes to " 12.2 pixels" (or in the example case 223.9999) then it should be `ceil`'d so that we have a pixel to cover that extra little bit that should be there.

> But on the whole, I would expect that if i said "I want target height to be 100" it should come out as 100.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,19 +1,18 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 23.12.1
     hooks:
       - id: black
-        language_version: python3.6
         files: \.py$
   - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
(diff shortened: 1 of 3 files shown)


OsaAjani:
We’ve successfully resolved the issue in V2. Thank you for your valuable contribution and feedback – it really helped us improve! If you encounter anything else, feel free to reach out.

Issue #899, opened 2019-01-11
"KeyError: 'video_fps'" when trying to convert youtube webm to mp3


adhoc92:
I'm downloading audio in mp4/web containers using Pytube, but I'm getting "KeyError: 'video_fps'" when trying to convert them to mp3 with moviepy. 

Code:
`import moviepy.editor as mp`
`from pytube import YouTube`

`#download video`
`youtubeurl = "https://www.youtube.com/watch?v=jYypQueS3ew&index=3&list=PL1nNuwHORUCPajRTocTSCJheaMzbWRu23&t=0s"`
`yt = YouTube(youtubeurl)`
`stream = yt.streams.filter(subtype='webm',only_audio=True).last().download("/home/jaws/Desktop/Stuff/Moosepy/temp/",` `filename="temp_audio")`
`print` `stream`
`print` `"Done!"`

`#convert to mp3`
`clip = mp.VideoFileClip("/home/jaws/Desktop/Stuff/Moosepy/temp/temp_audio.webm").subclip()`
`clip.audio.write_audiofile("audio.mp3")`
`print "Done!"`

Output:

Traceback (most recent call last):
  File "/home/jaws/Desktop/Stuff/Moosepy/test.py", line 13, in <module>
    clip = mp.VideoFileClip("/home/jaws/Desktop/Stuff/Moosepy/temp/temp_audio.webm").subclip()
  File "/home/jaws/.local/lib/python2.7/site-packages/moviepy/video/io/VideoFileClip.py", line 91, in __init__
    fps_source=fps_source)
  File "/home/jaws/.local/lib/python2.7/site-packages/moviepy/video/io/ffmpeg_reader.py", line 34, in __init__
    self.fps = infos['video_fps']
KeyError: 'video_fps'

Am I missing the codec? It converts mp4 and Webm video files to mp3 fine. It's just these weird audio files in webm/mp4 containers that won't convert. I guess since the FPS is 0, that's throwing up the error? How can I get around this? 
`
Moviepy 0.2.3.5
Python 2.7.12

adalbertopc:
I have the same issue. But im trying to convert mp4(only audio) to mp3

lyraxvincent:
@adhoc92 provide the full path --> `clip.audio.write_audiofile("/home/jaws/Desktop/Stuff/Moosepy/temp/audio.mp3")`

utkarshmttl:
[...]

PR #2114, opened 2024-02-14
Fix ffmpeg_reader not selecting a default stream

### LINKED ISSUE #869: KeyError: 'video_fps'
### Expected Behavior

No error should occur  as per docs

### Actual Behavior

> Traceback (most recent call last):
>   File "intro_prepare.py", line 10, in <module>
>     
>   File "/Users/ishandutta2007/.pyenv/versions/3.5.0/lib/python3.5/site-packages/moviepy/video/io/VideoFileClip.py", line 91, in __init__
>     fps_source=fps_source)
>   File "/Users/ishandutta2007/.pyenv/versions/3.5.0/lib/python3.5/site-packages/moviepy/video/io/ffmpeg_reader.py", line 34, in __init__
>     self.fps = infos['video_fps']
> KeyError: 'video_fps'
> 
> 

### Steps to Reproduce the Problem

```
from moviepy.editor import *
from moviepy.video.io.ffmpeg_tools import ffmpeg_extract_subclip

from PIL import Image

ffmpeg_extract_subclip("./library/mySound_Effect.mp4", 6, 8.7, targetname="test.mp4")
intro_sound_vidio_clip = VideoFileClip('test.mp4')

```
### Specifications

  - Python Version: 3.5.0
  - Moviepy Version: 0.2.3.5
  - Platform Name: OSX 10.13
  - Platform Version: 10.13(High Sierra)

bzczb:
fix #869 #899
ffmpeg, at least the version that's installed automatically with moviepy, sometimes comes up with no default stream for a given input file. This PR does nothing more than set a default stream automatically when none is specified, which is the cause of the missing "video_fps" issue.

Unfortunately I fixed this bug a long time ago in a working branch, and can't find the file that raised this problem in the first place. I think it was a ".wmv" or ".asf".


keikoro:
@bzczb Thanks for the submission – I don't have time to review its actual contents, but saw it didn't pass the formatting checks. Could you update the formatting?

bzczb:
@keikoro formatting fixed. Can't get pre-commit setup in `CONTRIBUTING.md` to work any which way. 3.6 venv fails when doing `pip install -e`, 3.12 venv fails because it's not 3.6.

OsaAjani:
I would need a few tests added to show the proper working of this, the line 645 by itself make me doubt this work perfectly as intended.

bzczb:
I need to remake this PR to match latest master.

The line looks right to me, the intent is to select a default at the end instead of the middle, so the `global_data` variable has to be removed and new logic added.

OsaAjani:
Oh, I missed the rewrite of return on parsing methods. I let you check what need to be updated for the new master and add some tests showing an example of this in work. Thanks a lot for your work !

bzczb:
added test to demonstrate some brokenness with .wmv

OsaAjani:
Thanks for updating to v2 !

PR #2117, opened 2024-02-14
Fix inexact and inconsistent frame seek

### LINKED ISSUE #2115: Seeking by frame is inexact and inconsistent
#### Expected Behavior
`clip.make_frame(0)` returns frame 0, `clip.make_frame(1 / clip.fps)` returns frame 1, `clip.make_frame(2 / clip.fps)` returns frame 2, etc.

#### Actual Behavior

moviepy largely treats "get the frame at time x" as shorthand for "get the frame displayed at time x", so "get the frame at time 0.00001" should mean "get frame 0". But ffmpeg treats "get the frame at time x" as "skip all frames until reaching time x", so sometimes `clip.make_frame(0.00001)` will return frame 1, and sometimes it will return frame 0, depending on what's been cached in `last_read`.

Load a video clip which has a different image every frame. (Some clips will have the same image for multiple consecutive frames, which will make it seem like it works correctly)

Here is a clip you can use:

https://github.com/Zulko/moviepy/assets/114126196/9c03b087-5d36-4ef4-9b12-9de823678624

```
import moviepy.editor as edit

clip = edit.VideoFileClip(r"colorbars.mp4")

test1 = clip.make_frame(0)
test2 = clip.make_frame(0.001)
test3 = clip.make_frame(60)
test4 = clip.make_frame(0.001)
test5 = clip.make_frame(60)
test6 = clip.make_frame(0)
(test1 == test2).all()
True
(test1 == test6).all()
True
(test3 == test5).all()
False
(test2 == test4).all()
False
```

#### Specifications

  - Python Version: 3.10
  - MoviePy Version: master
  - Platform Name: Windows

#### Testing note

`test_unusual_order_frame_pos()` in `test/test_ffmpeg_reader.py` is insufficient to detect this bug, as the test video used runs at 1 frame per second exactly. The inconsistency between "moviepy seek" and "ffmpeg seek" does not appear if the timestamps of all frames can be represented exactly in decimal!

bzczb:
Fixes #2115 seeking by frame is inexact and inconsistent

### DIFF
--- a/moviepy/Clip.py
+++ b/moviepy/Clip.py
@@ -1,6 +1,7 @@
 """Implements the central object of MoviePy, the Clip, and all the methods that
 are common to the two subclasses of Clip, VideoClip and AudioClip.
 """
+
 import copy as _copy
 from functools import reduce
 from numbers import Real
--- a/moviepy/audio/fx/all/__init__.py
+++ b/moviepy/audio/fx/all/__init__.py
@@ -4,6 +4,7 @@ moviepy.audio.fx.all is deprecated.
 Use the fx method directly from the clip instance (e.g. ``clip.audio_normalize(...)``)
 or import the function from moviepy.audio.fx instead.
 """
+
 import warnings
 
 from .. import *  # noqa 401,F403
--- a/moviepy/decorators.py
+++ b/moviepy/decorators.py
@@ -1,4 +1,5 @@
 """Decorators used by moviepy."""
+
 import inspect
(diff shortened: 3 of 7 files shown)


OsaAjani:
Hi @bzczb can you tell me if you think this is the same thing as #2296 and if so which one we should use ?

OsaAjani:
Thanks a lot for your contribution, I have merge it with PR #2353

PR #2123, opened 2024-02-29
Fix Writing Transparent Video Files

stratty7:
This will allow for clips that are loaded with an alpha channel (a .mov file for example) to be writen back out as a .mov file with an alpha layer (transparent background). The code already largely exists for this, the write_videofile function simply needs to be allowed to pass the with_mask arg to the ffmpeg_write_video, which then handles the rest.

For example:
```python
transparentclip = VideoFileClip("1.mov", has_mask=True)
transparentclip.write_videofile("test.mov", codec="png", fps=30, with_mask=True)
```
Longer example:
```python
transparentclip = VideoFileClip("1.mov", has_mask=True)
transparentclip.write_videofile("test.mov", codec="png", fps=30, with_mask=True)

background = VideoFileClip("background.mp4")
overlay = VideoFileClip("test.mov", has_mask=True)
comp = CompositeVideoClip([background, overlay])
comp.write_videofile("final.mp4", fps=30, codec="libx264")
```
Even longer
```python
# Where 1.mov and 2.mov both have an alpha channel
clips = [VideoFileClip("1.mov", has_mask=True), VideoFileClip("2.mov", has_mask=True)]
concat = concatenate_videoclips(clips, method="chain")
concat.write_videofile("test.mov", codec="png", fps=30, with_mask=True)
background = VideoFileClip("background.mp4")
overlay = VideoFileClip("test.mov", has_mask=True)
comp = CompositeVideoClip([background, overlay])
comp.write_videofile("final.mp4", fps=30, codec="libx264")
```

### DIFF
--- a/moviepy/video/io/ffmpeg_writer.py
+++ b/moviepy/video/io/ffmpeg_writer.py
@@ -226,7 +226,7 @@ def ffmpeg_write_video(
     threads=None,
     ffmpeg_params=None,
     logger="bar",
-    pixel_format=None,
+    pixel_format="rgb24",
 ):
     """Write the clip to a videofile. See VideoClip.write_videofile for details
     on the parameters.
@@ -238,8 +238,6 @@ def ffmpeg_write_video(
     else:
         logfile = None
     logger(message="MoviePy - Writing video %s\n" % filename)
-    if not pixel_format:
-        pixel_format = "rgb24"
     with FFMPEG_VideoWriter(
         filename,
         clip.size,
--- a/moviepy/video/io/ffmpeg_writer.py
+++ b/moviepy/video/io/ffmpeg_writer.py
@@ -221,7 +221,6 @@ def ffmpeg_write_video(
     codec="libx264",
     bitrate=None,
     preset="medium",
-    with_mask=False,
     write_logfile=False,
     audiofile=None,
     threads=None,
@@ -240,7 +239,7 @@ def ffmpeg_write_video(
         logfile = None
     logger(message="MoviePy - Writing video %s\n" % filename)
     if not pixel_format:
-        pixel_format = "rgba" if with_mask else "rgb24"
+        pixel_format = "rgb24"
     with FFMPEG_VideoWriter(
         filename,
         clip.size,
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -218,6 +218,7 @@ class VideoClip(Clip):
         ffmpeg_params=None,
         logger="bar",
         pixel_format=None,
+        with_mask=False,
     ):
         """Write the clip to a videofile.
 
@@ -314,6 +315,9 @@ class VideoClip(Clip):
         pixel_format
           Pixel format for the output video file.
 
+        with_mask
+          Set to ``True`` if there is a mask in the video to be encoded.
+
         Examples
         --------
 
@@ -390,6 +394,7 @@ class VideoClip(Clip):
             ffmpeg_params=ffmpeg_params,
             logger=logger,
             pixel_format=pixel_format,
+            with_mask=with_mask
         )
 
         if remove_temp and make_audio:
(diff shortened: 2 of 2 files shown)


keikoro:
@stratty7 Looks like there are problems with formatting which would need fixing.

Also, looking at changes that came before this addition: shouldn't this be propagated to elsewhere as well? E.g. when `pixel_format` was introduced, it got added in more than one place.

keikoro:
Hmm, looking at the above examples as well as other files, I guess I'm confused about two things:

* What's the difference between `has_mask` and `with_mask`?
* What's the difference between your addition and [this instance of `with_mask`](https://github.com/Zulko/moviepy/blob/master/moviepy/video/VideoClip.py#L422)?

stratty7:
> Hmm, looking at the above examples as well as other files, I guess I'm confused about two things:
> 
> 1. What's the difference between `has_mask` and `with_mask`?
> 2. What's the difference between your addition and [this instance of `with_mask`](https://github.com/Zulko/moviepy/blob/master/moviepy/video/VideoClip.py?rgh-link-date=2024-03-05T21%3A36%3A01Z#L422)?
> 
> When I mentioned `pixel_format` above, which is a line above your change, I was specifically talking about that it got added to `VideoClip.py` and `VideoFileClip.py` simultaneously (as well as other places) in #1237 (where it was still called `pix_fmt`). Your change is only for the former. Looking at the latter, I can see that [`has_mask` originates there](https://github.com/Zulko/moviepy/blob/bc8d1a831d2d1f61abfdf1779e8df95d523947a5/moviepy/video/io/VideoFileClip.py#L27). Do they set the same thing?

I understand your points and see what you mean, i think. 

With mask is not used in the ffmpeg writer at all, and using the `pixel_format` can achieve the same result without the need of adding any more code. I have changed the code to reflect that and removed the has_mask arg from ffmpeg_writer.

```
transparentclip = VideoFileClip("1.mov", has_mask=True)
transparentclip.write_videofile("test.mov", codec="png", fps=30, pixel_format="rgba")
```

moviepy/video/io/ffmpeg_writer.py
```diff
@@ -221,13 +221,12 @@ def ffmpeg_write_video(
     codec="libx264",
     bitrate=None,
     preset="medium",
-    with_mask=False,
```
stratty7, line 224: `with mask` when set would only set the `pixel_format` to have alpha.
I think we can remove it and default to "rgb24". If anyone wants the `pixel_format` to have the alpha layer, they can pass it through the pixel_format itself.

stratty7:
I think using pixel_format="rgba" addresses my needs.
However, `with_mask` is currently not passed to `ffmpeg_write_video` anywhere, I think this could probably be cleaned up. I will close this PR and if I get time clean that up and open a new one. Thanks for the reviews.

keikoro:
To clarify my last comment: I only used `pixel_format` as an example for a change that was done in both files, VideoClip.py and VideoFileClip.py, so thought that maybe your suggested change should also happen in both files. Which _then_ led me to the `with_mask` vs. `has_mask` thing.

I didn't look at it in more detail, but am I understanding correctly that this discrepancy was actually introduced earlier, and `pixel_format` was potentially, partially (?) included to address the issue of one argument not getting passed down the chain? Or were you more saying that you just had the realisation `pixel_format` could be used to replace the other argument, even though it wasn't so far?

stratty7:
> Or were you more saying that you just had the realisation pixel_format could be used to replace the other argument, even though it wasn't so far?

Yeah this. 

I think `pixel_format` could be used in the place of `with_mask` at the moment. Only if 'with_mask` is implemented properly into `write_videofile` and `ffmpeg_write_video` (which it currently isn't, but would be nice to have) then I understand having both args, but currently having both seems somewhat pointless. I think implementing `with_mask` properly and fully (instead of my bandaid fix in the original code of this pr) into `write_videofile` and `ffmpeg_write_video` is the true fix here.

That is why I have closed this for now and will work on it when i find time.

PR #2152, opened 2024-04-10
Passing duration as milliseconds instead of seconds in write_gif_with_image_io. Fixes #2151.

### LINKED ISSUE #2151: Unable to set FPS in GIFs
Hey,

When using MoviePy to store frame sequences as GIFs, I noticed that setting the fps parameter seems to have no effect. After digging around in the code a little bit, I found the issue.

In `moviepy/video/io/gif_writers.py`, line 442:
```python
writer = imageio.save(
    filename, duration=1.0 / fps, quantizer=quantizer, palettesize=colors, loop=loop
)
```

Here, the duration passed to imageio is in seconds, which is consistent with [imageio's documentation](https://imageio.readthedocs.io/en/v2.14.0/reference/_backends/imageio.plugins.pillow_legacy.html#parameters-for-saving). Unfortunately, though, the documentation is erroneous in this case, as the underlying Pillow [expects milliseconds](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#gif).

I believe the best course of action is to ignore the documentation of imageio and pass the value as milliseconds.

Currently, this issue affects [PyTorch's](https://github.com/pytorch/pytorch/) `SummaryWriter`, which uses MoviePy to store videos as GIFs.

Best,
Tim

TimSchneider42:

### DIFF
--- a/moviepy/video/io/gif_writers.py
+++ b/moviepy/video/io/gif_writers.py
@@ -439,7 +439,11 @@ def write_gif_with_image_io(
     quantizer = 0 if opt != 0 else "nq"
 
     writer = imageio.save(
-        filename, duration=1.0 / fps, quantizer=quantizer, palettesize=colors, loop=loop
+        filename,
+        duration=1000.0 / fps,
+        quantizer=quantizer,
+        palettesize=colors,
+        loop=loop,
     )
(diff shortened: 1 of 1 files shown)


OsaAjani:
This bug have been fixed with the v2 !

Issue #2160, opened 2024-04-22
ffmpeg interpets video title as a command line argument when it begins with "-"


TheSpaceSheep:
#### Minimal reproducible example :
```
from moviepy.editor import *
clip = ColorClip((640, 480), color=(0, 0, 0), duration=5)
clip.write_videofile("-filenamethatbeginswith-.webm", fps=24)
```

#### Expected Behavior
write to a video file with name "-filenamethatbeginswith-.webm"

#### Actual Behavior
The following exception occurs (my interpretation is that a video file starting with "-" is interpreted as a command line argument for ffmpeg)

```
MoviePy error: FFMPEG encountered the following error while writing fil
e -myHolidays_edited.webm:

 b"Unrecognized option 'myHolidays_edited.webm'.\nError splitting the a
rgument list: Option not found\n"
```

#### Specifications

  - Python Version: 3.11
  - MoviePy Version: 1.0.3
  - Platform Name: Linux
  - Platform Version: 6.8.7-arch1-1

I will try to fix it myself !

OsaAjani:
Fixed by #2282

PR #2163, opened 2024-04-26
(fix) escape dashes in the beginning of filenames in ffmpeg commands

TheSpaceSheep:
This PR fixes https://github.com/Zulko/moviepy/issues/2160.
Filenames that starts with dash '-' are no longer interpreted as command line arguments by ffmpeg

See https://github.com/Zulko/moviepy/issues/2160


OsaAjani:
I've taken this fine PR and rewrite it for the new version with some improvements for filenames with special chars using shlex.quote. Thanks for the original work @TheSpaceSheep !

Issue #2170, opened 2024-05-07
Error fps in generated gif file of `write_gif_with_image_io`


Beanpow:
# Problem
The fps of the generated gif file of function `write_gif_with_image_io` is not correct.

# Minimal Case
The following code is similar to the function `write_gif_with_image_io`, but the generated gif's fps is not equal to the fps in the code.
https://github.com/Zulko/moviepy/blob/bc8d1a831d2d1f61abfdf1779e8df95d523947a5/moviepy/video/io/gif_writers.py#L422

```
import imageio
import numpy as np


filename = 'test.gif'
fps = 30
quantizer = 0
colors = 256
loop = 0

writer = imageio.save(
    filename,
    duration=1.0/fps,
    quantizer=quantizer,
    palettesize=colors,
    loop=loop
)

for i in range(100):
    writer.append_data(np.random.randint(0, 255, (3, 100, 100), dtype=np.uint8))
```

# Potential Reason
As the following imageio code, the `duration` is in `ms`. But the `write_gif_with_image_io` maybe think it is `second`.
https://github.com/imageio/imageio/blob/ac863934d590ffc64155475a0fc365c1c82a8267/imageio/plugins/pillow.py#L408-L413
```
        if "fps" in kwargs:
            warnings.warn(
                "The keyword `fps` is no longer supported. Use `duration`"
                "(in ms) instead, e.g. `fps=50` == `duration=20` (1000 * 1/50).",
                DeprecationWarning,
            )
            kwargs["duration"] = 1000 * 1 / kwargs.get("fps")
```

https://github.com/Zulko/moviepy/blob/bc8d1a831d2d1f61abfdf1779e8df95d523947a5/moviepy/video/io/gif_writers.py#L441-L443

PR #2171, opened 2024-05-07
Fix: change the unit of duration in func `write_gif_with_image_io`

Beanpow:
Please refer to the issue https://github.com/Zulko/moviepy/issues/2170.


OsaAjani:
Fixed with V3

Issue #2189, opened 2024-06-02
Unable to run moviepy on debian 10


belalsalih:
I am trying to run some processing on video uploaded by users using azure functions(python), my code runs perfectly on local pc(macOS) however it is not running at all once deployed on kubernetes.
logging into the pod and running: 'from moviepy.editor import *' will result in the following error:

```
>>> from moviepy.editor import *
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/moviepy/editor.py", line 36, in <module>
    from .video.io.VideoFileClip import VideoFileClip
  File "/usr/local/lib/python3.9/site-packages/moviepy/video/io/VideoFileClip.py", line 3, in <module>
    from moviepy.audio.io.AudioFileClip import AudioFileClip
  File "/usr/local/lib/python3.9/site-packages/moviepy/audio/io/AudioFileClip.py", line 3, in <module>
    from moviepy.audio.AudioClip import AudioClip
  File "/usr/local/lib/python3.9/site-packages/moviepy/audio/AudioClip.py", line 7, in <module>
    from moviepy.audio.io.ffmpeg_audiowriter import ffmpeg_audiowrite
  File "/usr/local/lib/python3.9/site-packages/moviepy/audio/io/ffmpeg_audiowriter.py", line 7, in <module>
    from moviepy.config import get_setting
  File "/usr/local/lib/python3.9/site-packages/moviepy/config.py", line 36, in <module>
    FFMPEG_BINARY = get_exe()
  File "/usr/local/lib/python3.9/site-packages/imageio/plugins/ffmpeg.py", line 173, in get_exe
    return imageio_ffmpeg.get_ffmpeg_exe()
  File "/usr/local/lib/python3.9/site-packages/imageio_ffmpeg/_utils.py", line 28, in get_ffmpeg_exe
    exe = _get_ffmpeg_exe()
  File "/usr/local/lib/python3.9/site-packages/imageio_ffmpeg/_utils.py", line 44, in _get_ffmpeg_exe
    exe = os.path.join(_get_bin_dir(), FNAME_PER_PLATFORM.get(plat, ""))
  File "/usr/local/lib/python3.9/site-packages/imageio_ffmpeg/_utils.py", line 69, in _get_bin_dir
    ref = importlib.resources.files("imageio_ffmpeg.binaries") / "__init__.py"
  File "/usr/local/lib/python3.9/importlib/resources.py", line 147, in files
[...]

PR #2190, opened 2024-06-02
fix: set default binary to auto-detect

antonpetrov145:
- Additional info: this PR is because issue #2189, to fix this I did set the default value of `FFMPEG_BINARY` to `auto-detect` and remade the test

### DIFF
--- a/moviepy/config.py
+++ b/moviepy/config.py
@@ -18,7 +18,7 @@ try:
 except ImportError:
     DOTENV = None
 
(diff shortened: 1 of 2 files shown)


studioj:
this is also happening on windows. A temporary fix would be to lock 
> imageio-ffmpeg<0.5

antonpetrov145:
Closing as per discussion in referred issue.

keikoro:
Link to the comment/thread in question, just for reference: https://github.com/Zulko/moviepy/issues/2189#issuecomment-2160598952

Issue #2183, opened 2024-05-27
Test suite fails for most Windows, Linux tests


keikoro:
[`test_rotate`](https://github.com/Zulko/moviepy/blob/f5a6fde76271a773f8b8e0c64464d371c8e07f57/tests/test_fx.py#L924) starts failing at some point in most Windows and Linux versions.

The only (Python) versions for which tests pass are:
* Windows 3.7
* Linux 3.7

Versions which fail:
* Window 3.8
* Window 3.9
* Window 3.10
* Window 3.11
* Linux 3.8
* Linux 3.9
* Linux 3.10
* Linux 3.11

Excerpt from log for Linux 3.8:

```
=========================== short test summary info ============================
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-bilinear-deg--360-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-bilinear-deg-0-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-bilinear-deg-360-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-bilinear-deg-720-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-bilinear-rad--360-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-bilinear-rad-0-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-bilinear-rad-360-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-bilinear-rad-720-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-nearest-deg--360-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-nearest-deg-0-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-nearest-deg-360-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-nearest-deg-720-True]
FAILED tests/test_fx.py::test_rotate[45-translate5-None-bg_color5-expected_frames5-nearest-rad--360-True]
[...]

PR #2195, opened 2024-06-06
fixes for tests

antonpetrov145:
- set requred version of pillow to 9.5.0
- test-ffmpeg-reader - change duration number to match video
- ffmpeg-reader - add displaymatrix to be read as video rotation
- config set binary to autodetect

addresses issue #2183

### DIFF
--- a/moviepy/config.py
+++ b/moviepy/config.py
@@ -18,7 +18,7 @@ try:
 except ImportError:
     DOTENV = None
 
-FFMPEG_BINARY = os.getenv("FFMPEG_BINARY", "ffmpeg-imageio")
+FFMPEG_BINARY = os.getenv("FFMPEG_BINARY", "auto-detect")
 IMAGEMAGICK_BINARY = os.getenv("IMAGEMAGICK_BINARY", "auto-detect")
 
 IS_POSIX_OS = os.name == "posix"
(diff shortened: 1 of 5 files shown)


keikoro:
@antonpetrov145 Thanks for the contribution!

I'm not sure if removing `ffmpeg-imageio` for the sake of fixing the tests is the right solution, though. See [this comment](https://github.com/Zulko/moviepy/issues/1874#issuecomment-1355269034) by Zulko in one of the big threads/discussions about the future of the project (the paragraph before the bullet point summary specifically, which is about dependency management).

keikoro:
Idk, maybe worth throwing your suggested fix into the discussion?

antonpetrov145:
Thank you for your time to look at the changes I made.

The idea behind setting the ffmpeg binary to auto-detect by default and removing the test is because of an issue that was because of ffmpeg-imageio setting the default binary to exe even on Linux (my other PR) causing Path error on importing moviepy.

keikoro:
@antonpetrov145 I've now had a look at the code, though haven't been able to test it – am I understanding correctly that changing `FFMPEG_BINARY` is not actually needed to fix the issue at hand? If so, I'd ask you leave off that particular change and only (force) push the relevant bits of code.

antonpetrov145:
@keikoro actually it is - without setting the binary to `auto-detect` tests are not passing, at least on my side. If you think I'm doing something wrong I'm happy to make the change you propose.

OsaAjani:

PR #2201, opened 2024-07-10
BugFix: Fix the alignment of audio duration and video duration after concatenate_videoflips

zyw241:
When I use concatenate_videoclips to concatenate my video clips, I find that the concatenated clip's duration property does not align with its audio's duration property when the last video clip to be concatenated has no audio. This will cause much error when using the concatenated clip after, such as clip.subclip will report an ValueError. So I modify the code in concatenate_videoclips to set the audio's duration align with video's duration.


OsaAjani:
Can you provide some code triggering the error, I cannot reproduce it.

Issue #2008, opened 2023-07-10
Overlay video with .webm format video with transparent background


doanaktar:
Hello, I have a video in .webm format with a transparent background(clip2) and a normal mp4 video(clip1). I want to overlay clip2 video on clip1 video, but when I do that, the transparent feature in the background of clip2 disappears and a red beam of light comes on.

Here is my code:

```
from moviepy.editor import VideoFileClip, CompositeVideoClip
clip1 = VideoFileClip("test.mp4")
clip2 = VideoFileClip("transparent.webm")
clip2 = clip2.resize(0.25)

video = CompositeVideoClip([clip1,
                            clip2.set_position(("right","bottom"))])

video.write_videofile("my_concatenation.mp4")
```

TheHimanshuRastogi:
Show both videos for better help.

Good0007:
You can set param: has_mask=True ,like this:
```python

background = VideoFileClip("background.mp4")
layer = VideoFileClip("overlay.mov", has_mask=True)
CompositeVideoClip([background, layer]).write_videofile("out.mp4", codec='libx264')

```

jitendra-koodo:
> You can set param: has_mask=True ,like this:
> 
> ```python
> background = VideoFileClip("background.mp4")
> layer = VideoFileClip("overlay.mov", has_mask=True)
> CompositeVideoClip([background, layer]).write_videofile("out.mp4", codec='libx264')
> ```

This does not show the background video. Transparency does not seem to work.

webm video : https://rotato.netlify.app/alpha-demo/movie-webm.webm

BackrndSource:
If you install the Pull Request #2229 , or you wait until the approval and update of moviepy, you can set parameters decoder="libvpx-vp9" and  has_mask=True in VideoFIleClip like this:

```py
from moviepy.editor import VideoFileClip, CompositeVideoClip
[...]
clip2 = VideoFileClip("transparent.webm", has_mask=True, decoder="libvpx-vp9")
[...]
```

OsaAjani:
Fixed with v2.1.2

PR #2229, opened 2024-10-13
Add 'decoder' Parameter to VideoFileClip and ffmpeg_reader

BackrndSource:
This commit allows users to specify the decoder used to decode the video file.

Fixed bug:
- When importing a .webm video with alpha channel, transparency are not working.

Solved issue:
 - #2008  

References:
 - https://www.reddit.com/r/ffmpeg/comments/fgpyfb/help_with_webm_with_alpha_channel/

"FFmpeg's native VPx decoders don't decode alpha. You have to use the libvpx decoder to preserve transparency in .webm videos. "

Example of use:

VideoFileClip("video.webm", decoder="libvpx-vp9")

### DIFF
--- a/moviepy/video/io/VideoFileClip.py
+++ b/moviepy/video/io/VideoFileClip.py
@@ -52,6 +52,11 @@ class VideoFileClip(VideoClip):
       can be set to 'fps', which may be helpful if importing slow-motion videos
       that get messed up otherwise.
 
+    decoder:
+       The decoder used to decode the video file. FFmpeg's native VPx decoders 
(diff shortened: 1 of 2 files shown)


BackrndSource:
To preserve the alpha channel also needs to set has_mask parameter to True:
```py
VideoFileClip("video.webm", has_mask=True, decoder="libvpx-vp9")
```

OsaAjani:
The original problem was fixed by #2269, do you think we still need to add a codec parameter in regard to this new information, or should we let ffmpeg choose by himself ?

OsaAjani:
Actually #2230 answer that question, I close

PR #2230, opened 2024-10-13
 Add 'decoder' parameter to VideoFileClip and ffmpeg_reader - v2 (dev)

BackrndSource:
Same as #2229 but for v2 (dev)

Fixed bug:
    - When importing a .webm video with alpha channel, transparency are not working.

Solved issue:
    - #2008 
 
References:
    - https://www.reddit.com/r/ffmpeg/comments/fgpyfb/help_with_webm_with_alpha_channel/

"FFmpeg's native VPx decoders don't decode alpha. You have to use the libvpx decoder to preserve transparency in .webm videos. "

To preserve the alpha channel also needs to set has_mask parameter to True.

Example of use:

```py
VideoFileClip("video.webm", has_mask=True, decoder="libvpx-vp9")
```

### DIFF
--- a/moviepy/video/io/VideoFileClip.py
+++ b/moviepy/video/io/VideoFileClip.py
@@ -56,6 +56,11 @@ class VideoFileClip(VideoClip):
       'rgb24' will be used as the default format unless ``has_mask`` is set
       as ``True``, then 'rgba' will be used.
 
+    decoder:
+       The decoder used to decode the video file. FFmpeg's native VPx decoders 
+       don't decode alpha. You have to use the libvpx decoder. Set this to 
+       'libvpx-vp9' if you want to preserve transparency in .webm video.
+
 
     Attributes
     ----------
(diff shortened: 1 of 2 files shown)


Zulko:
Do you think that instead of providing a decoder (which is complicated for the user) maybe we could have an internal check and if the video extension is `.webm` and `has_mask=True`, then the decoder is auto-set to `libvpx-vp9` ? I see two ways of doing this. Either keep the `decoder` parameter but set a default `decoder="auto"` and handle that case, or remove the `decoder` parameter completely and simply add   `['-c:v', "libvpx-vp9"]` when he video extension is `.webm` and `has_mask=True`.

BackrndSource:
Great, what you propose seems like the best option. I'll work on it.

OsaAjani:
Unknowingly of this particular attempt at fixing alpha layer reading, I pretty much implemented the solution described by @Zulko, -(auto adding vp9 when needed) so #2269 fix this and I will close

BackrndSource:
Thank you @OsaAjani ! So nice work :)

PR #2253, opened 2024-11-23
Make `FFMPEG_VideoReader` more robust to missing information.

### LINKED ISSUE #2256: MoviePy 2.0 throws exception on loading video previous version worked with
#### Expected Behavior

MoviePy should continue to work with the same videos it did previously, even if those videos aren't fully compliant (e.g. are missing some metadata).

#### Actual Behavior

The same video crashes on MoviePy 2.0 but works with MoviePy 1.x.

See the `corrupt_video.mp4` file from https://github.com/Breakthrough/PySceneDetect/tree/resources/tests/resources and the associated unit test in https://github.com/Breakthrough/PySceneDetect/blob/95d20ddca57bb8cba77354697cc092643bd04afb/tests/test_video_stream.py#L359

The issue seems to come from the ffmpeg info parser assuming the presence of certain metadata fields.  Instead of failing, MoviePy should probably try to set some reasonable default value for these fields if they are not critical, to improve compatibility with media files.

#### Specifications

Tested on a wide variety of OS and Python versions:
 -  os: [macos-13, macos-14, ubuntu-20.04, ubuntu-latest, windows-latest]
 - python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]

Breakthrough:
Allow `FFMPEG_VideoReader` to function more robustly in the presence of missing information.

I have a [variety of existing tests](https://github.com/Breakthrough/PySceneDetect/blob/main/tests/test_video_stream.py) in my project PySceneDetect that [began to fail suddenly](https://github.com/Breakthrough/PySceneDetect/issues/461) during the update to MoviePy 2.0.  Some of these tests include loading corrupt videos, and testing how the FFMPEG_VideoReader class responds to certain seeking and EOF behaviors.

In previous versions, a video under test would read the first 45 frames or so before failing, but with the latest version, it throws an exception when trying to create the `FFMPEG_VideoReader`.  This PR has the required changes to make loading the video more robust to the missing information to still allow corrupt files to be processed, even if incompletely.

Resolves #2256 

### DIFF
--- a/moviepy/video/io/ffmpeg_reader.py
+++ b/moviepy/video/io/ffmpeg_reader.py
@@ -35,8 +35,10 @@ class FFMPEG_VideoReader:
             decode_file=decode_file,
             print_infos=print_infos,
         )
-        self.fps = infos["video_fps"]
-        self.size = infos["video_size"]
+        # If framerate is unavailable, assume 1.0 FPS to avoid divide-by-zero errors.
+        self.fps = infos.get("video_fps", 1.0)
+        # If frame size is unavailable, set 1x1 divide-by-zero errors.
+        self.size = infos.get("video_size", (1, 1))
 
         # ffmpeg automatically rotates videos if rotation information is
         # available, so exchange width and height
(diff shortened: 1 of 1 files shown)


Breakthrough:
Passes unit tests locally and ensured code is formatted.

Zulko:
Thank you, I just released it under moviepy==2.1.1

Issue #2251, opened 2024-11-22
Font exception if empy


BornSaint:
```python
# Python Snippet from moviepy/video/VideoClip.py:1594
        try:
            print("f", font)
            _ = ImageFont.truetype(font)
        except Exception as e:
            raise ValueError(
                "Invalid font {}, pillow failed to use it with error {}".format(font, e)
            )
```
```javascript
/* Error thrown by snippet */

f Arial-Black
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
File ~/anaconda3/envs/cleofas/lib/python3.9/site-packages/moviepy/video/VideoClip.py:1594, in TextClip.__init__(self, font, text, filename, font_size, size, margin, color, bg_color, stroke_color, stroke_width, method, text_align, horizontal_align, vertical_align, interline, transparent, duration)
   1593     print("f", font)
-> 1594     _ = ImageFont.truetype(font)
   1595 except Exception as e:

```
#### Expected Behavior
No exception if font parameter in TextClip is empty or invalid

#### Actual Behavior

Unecessary Exception...

#### Steps to Reproduce the Problem

Set invalid font name or just not set font parameter (default to None)

#### Specifications

  - Python Version: 3.9.20
  - MoviePy Version: 2.0.0
  - Platform Name: Linux
  - Platform Version: Ubuntu 22.04 LTS
  
  #### Solution
  Include default truetype .ttf file to project that run in any system when parameter is empty or invalid instead of throwing an exception.

BornSaint:
Another way to solve this with class is to add Windows and Linux true types to the project, so names would be valid in any OS, also should set a default instead of None for parameter.

BornSaint:
realized that "fontsize" parameter also need a default value

OsaAjani:
Hi, in my opinion this isn't a bug but actually the expected behaviour. With the V2.0 release the logic of TextClip have changed a bit, and now we expect the user to provide the font he want to use, not the name, the actual path to a OpenType Font.

We also only support OpenType Font, I cannot exactly remember why but it had something to do with the way pillow handle fonts.

Also, fontsize dont need a default parameter because either can be left to default depending on other parameters (mostly method). I may be mistaking, but I think the way I wrote it was the optimal in regard to how devs could pass the fewer parameters possible using named parameters, at least it was my intention at the time ^^.

majidsh97:
code: TextClip(text="Hello",font="Tahoma", color='white',  size=video.size)

[...]

PR #2254, opened 2024-11-23
Fix invalid option in docstring of TextClip, as mentionned in #2251

OsaAjani:
Fix the docstring error on TextClip described in #2251, also remove a few debugging print previously forgotten in code.


Zulko:
thanks

Issue #2044, opened 2023-09-28
moviepy returns empty frames if frames are not accessed strictly sequentially


pgunn:
#### Expected Behavior
See below - the handle to a video file should not start returning zeros. The video file we're using with this is here:
https://caiman.flatironinstitute.org/~neuro/caiman_downloadables/msCam13.avi

We found this while spotting the error in pims, using the moviepy backend, and eventually tracked it down to moviepy when considering directly using moviepy to avoid what we believed was a bug in pims, instead to find that it's actually a bug in moviepy.

Oddly, accessing frame 0 of the movie fixes the handle to the movie

#### Actual Behavior
>>> mph = moviepy.editor.VideoFileClip(fn)
>>> mph.size
[752, 480]
>>> mph.duration
50.0
>>> mph.fps
20.0
>>> np.array(mph.get_frame(0)).max()
173
>>> np.array(mph.get_frame(2)).max()
206
>>> np.array(mph.get_frame(1)).max()
0
>>> np.array(mph.get_frame(0)).max()
173
>>> np.array(mph.get_frame(1)).max()
188

#### Specifications

  - Python Version: 3.11.5
  - MoviePy Version: 1.0.3
  - Platform Name: Linux (tested on Windows too)

pgunn:
Following it down into the backend:
```
mph = moviepy.video.io.ffmpeg_reader.FFMPEG_VideoReader(fn)
>>> mph.get_frame(0).max()
173
>>> mph.get_frame(1).max()
188
>>> mph.get_frame(2).max()
206
>>> mph.get_frame(1).max()
0
>>> mph.get_frame(0).max()
173
>>> mph.get_frame(1).max()
188
```

I'm looking at the logic here:
https://github.com/Zulko/moviepy/blob/master/moviepy/video/io/ffmpeg_reader.py#L218

But I find the existence of self.last_read and self.lastread as distinct variables to be puzzling. I'm guessing the bug is somewhere around there.

pgunn:
It looks like, at least with our datafiles, any use of `-ss` before `-i` in the call to ffmpeg causes ffmpeg itself to return empty frames, while the second call to `-ss` (while not performant) returns data correctly.

Found this by modifying moviepy/video/io/ffmpeg_reader.py to print out the commandline its building with various calls, and looking over the `if start_time != 0` block in initialize().

Something like:
```
[...]

PR #2257, opened 2024-11-23
Ensure `last_read` is always set.

Breakthrough:
The property `lastread` was renamed to `last_read` in #1170 however not all instances were changed. This leads to some cases where only one of the variables is updated.  For example, in the case where one calls `get_frame` with a frame exactly one ahead of the current position.  In this case, only `self.last_read` is updated, but `self.lastread` remains unchanged, leading the two properties to get out of sync.

There are third-party projects which relied on the lastread property as it was never declared as private in previous versions (e.g. no underscore prefix). To avoid breaking projects relying on this, we also provide a property with the old name for backwards compatibility.

This should also help maintain the code, as there is already some confusion over the two similarly named properties (#2044).


Breakthrough:
Passes unit tests locally and ensured code is formatted.

Zulko:
Thank you! Will do a new release on Sunday.

PR #2259, opened 2024-11-24
(mostly internal) variables renaming

Zulko:


Zulko:
womp womp not a good idea.

Issue #2247, opened 2024-11-21
ValueError in write_videofile after concatenate_videoclips


FedeSpu:
#### Expected Behavior
Export the video correctly after using concatenate_videoclips and write_videofile

#### Actual Behavior
ValueError, and the video is not produced

#### Steps to Reproduce the Problem

```
from moviepy import *
title_clip = TextClip(text=title_frame_txt, font="Arial", font_size=34, color='white', bg_color='black', size=(width, height), duration=1.5, text_align='center')
video = CompositeVideoClip(clip_list, use_bgclip=True)

final_video = concatenate_videoclips([title_clip, video], method='chain')
final_video.write_videofile(output_file, fps=out_fps, codec='libx264')
```

#### Solution
The error is in "ffmpeg_writer.py" file in line ``frame = np.dstack([frame, mask])``. The problem is that the mask has shape [1,1]. This is due to ``concatenate_videoclips``, that in line ``mask = clip.mask or ColorClip([1, 1], color=1, is_mask=True)`` produces mask of shape [1,1]. 
The solution that I've found is to replace this line with ``mask = clip.mask or ColorClip(clip.size, color=1, is_mask=True)``

#### Specifications

  - Python Version: 3.12.2
  - MoviePy Version: 2.0.0
  - Platform Name: Ubuntu
  - Platform Version: 22.04

Implosiv3:
I faced the same error and it is solved. Apparently, there is no reason to be [1, 1] shape in that situacion as it is fulfilling masks for clips without them.

OsaAjani:
Fixed with #2307

PR #2262, opened 2024-11-27
Fix CompositeVideoClip mask [1, 1] unexpected shape - Issue #2247

Implosiv3:
Fixes the issue when concatenating videoclips as mask is being considered as [1, 1] so the mask shape is different from the other videos and fails.

I think [1, 1] could be a default mask to avoid divisions by zero or something similar, but here it does not make sense.

Issue here: https://github.com/Zulko/moviepy/issues/2247

### DIFF
--- a/moviepy/video/compositing/CompositeVideoClip.py
+++ b/moviepy/video/compositing/CompositeVideoClip.py
@@ -289,7 +289,7 @@ def concatenate_videoclips(
             return clips[i].get_frame(t - timings[i])
 
         def get_mask(clip):
-            mask = clip.mask or ColorClip([1, 1], color=1, is_mask=True)
+            mask = clip.mask or ColorClip(clip.size, color=1, is_mask=True)
             if mask.duration is None:
                 mask.duration = clip.duration
             return mask


Implosiv3:
> Thank you! Would you have time to add a minimal test? Some test snippet that would fail on the last release, but succeed with that change?


If I try to execute this code below in which I try to concatenate one clip withou a mask and one with the default mask:
```python
from moviepy import concatenate_videoclips, VideoFileClip

video_without_mask = VideoFileClip('prueba.mp4')
video_with_mask = VideoFileClip('prueba.mp4').with_mask()

concatenate_videoclips([video_without_mask, video_with_mask]).write_videofile('test.mp4')
```

This is the result I receive:
```python
videos = concatenate_videos([
  File "<decorator-gen-68>", line 2, in write_videofile
  File "C:\Users\dania\AppData\Local\Programs\Python\Python39\lib\site-packages\moviepy\decorators.py", line 53, in requires_duration      
    return func(clip, *args, **kwargs)
  File "<decorator-gen-67>", line 2, in write_videofile
  File "C:\Users\dania\AppData\Local\Programs\Python\Python39\lib\site-packages\moviepy\decorators.py", line 143, in use_clip_fps_by_default
    return func(clip, *new_args, **new_kwargs)
  File "<decorator-gen-66>", line 2, in write_videofile
    return func(clip, *args, **kwargs)
  File "<decorator-gen-65>", line 2, in write_videofile
  File "C:\Users\dania\AppData\Local\Programs\Python\Python39\lib\site-packages\moviepy\decorators.py", line 94, in wrapper
    return func(*new_args, **new_kwargs)
  File "C:\Users\dania\AppData\Local\Programs\Python\Python39\lib\site-packages\moviepy\video\VideoClip.py", line 389, in write_videofile  
    ffmpeg_write_video(
  File "C:\Users\dania\AppData\Local\Programs\Python\Python39\lib\site-packages\moviepy\video\io\ffmpeg_writer.py", line 263, in ffmpeg_write_video
    frame = np.dstack([frame, mask])
  File "C:\Users\dania\AppData\Local\Programs\Python\Python39\lib\site-packages\numpy\lib\shape_base.py", line 715, in dstack
    return _nx.concatenate(arrs, 2)
ValueError: all the input array dimensions except for the concatenation axis must match exactly, but along dimension 0, the array at index 
0 has size 1080 and the array at index 1 has size 1
```

If you apply the change, it works:
```python
MoviePy - Done.
MoviePy - Writing video test.mp4

frame_index:  45%|█████████████████████████████████▉                                          | 111/249 [00:13<00:14,  9.21it/s, now=
```

As the first clip doesn't have any mask, the system applies one `[1, 1]` shape ColorClip default mask, which is incompatible with the `[1920, 1080]` of the other clip.

OsaAjani:
closed by #2307, thanks for pointing it out !

PR #2263, opened 2024-11-27
Proof-read docs

tburrows13:


Zulko:
Awesome, thank you! I was about to do a pass this weekend so great timing.

tburrows13:
I'm sure there's plenty I missed, especially as I didn't really cover the API reference bit. Also the "Version:" string on the front page has no version listed... should either be fixed or removed :)

PR #2267, opened 2024-11-29
Audio preview fixes

junhe:
I was trying follow the docs and found several issues in the doc and code. 

Please take a look.

### DIFF
--- a/moviepy/audio/io/ffplay_audiopreviewer.py
+++ b/moviepy/audio/io/ffplay_audiopreviewer.py
@@ -47,8 +47,6 @@ class FFPLAY_AudioPreviewer:
             "s%dle" % (8 * nbytes),
             "-ar",
             "%d" % fps_input,
-            "-ac",
-            "%d" % nchannels,
             "-i",
             "-",
         ]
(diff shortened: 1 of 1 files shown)


junhe:
I don't really know how to do pull requests. So please expect some weirdness... 

I split the doc fixes and code fixes to two requests.

junhe (commented):
Before this commit, there was obviously an issue. logfile was not defined.

junhe (commented):
If you go to ffplay documentation, you will find that `-ac` is not one of the options.

OsaAjani:
Hi, after checking -ac is actually a ffplay option at least up to ffplay version 5.1, so you would actually need to check ffplay version and use -ac option for version 5.* and maybe 6.* too, and migrate the equivalent in 7.*, which seems to be `-ch_layout` according to this ticket https://trac.ffmpeg.org/ticket/11077

OsaAjani:
I changed this to draft, pending the necessary improvements pointed in my previous message.

OsaAjani:
Supplented by PR #2331

PR #2270, opened 2024-12-02
Add bg_radius to text boxes

aaa3334:
Adding in a new optional parameter bg_radius. This parameter allows users to round the edges of the background text box so it is not sharp cornered.

This new parameter is optional and utilises the pil libraries rounded_rectangle.

https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.ImageDraw.rounded_rectangle

This has been around since pil 8.2.
Moviepy requires:  "pillow>=9.2.0,<11.0",
Therefore, this is compatible.

Note I have extensively tested on my local computer. I decided to  make the change as it was much simpler than me manually creating this on my end (I have also seen others eg. on reddit with the same issue).

I have not done many PR's before - please let me know if I have not done this correctly. I tried to follow the documentation but as I am newer to this with git I am unsure if I did it in the right place etc.

- [ x] I have added suitable tests demonstrating a fixed bug or new/changed feature to the test suite in `tests/` (I did this on my local computer but am unsure how to add it for this PR)
- [ x] I have properly documented new or changed features in the documentation or in the docstrings
- [ x] I have properly explained unusual or unexpected code in the comments around it

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -1465,7 +1465,7 @@ class TextClip(ImageClip):
         interline=4,
         transparent=True,
         duration=None,
-        bg_radius=0, # TODO : Move this with other bg_param on next breaking release
+        bg_radius=0,  # TODO : Move this with other bg_param on next breaking release
     ):
         def break_text(
             width, text, font, font_size, stroke_width, align, spacing
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -1727,13 +1727,14 @@ class TextClip(ImageClip):
 
         if bg_radius is None:
             bg_radius = 0
-        
+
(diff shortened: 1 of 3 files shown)


moviepy/video/VideoClip.py
```diff
@@ -1438,6 +1438,11 @@ class TextClip(ImageClip):
 
     duration
         Duration of the clip
+
+    bg_radius
+     A paramater to round the edges of the text background. Defaults to 0 if there
+     is no background. It will have no effect if there is no bg_colour added.
+     The higher the value, the more rounded the corners will become.
```
akx, line 1445:
```suggestion
     Round the edges of the text background. Defaults to 0, and has no
     effect if bg_color is not set.
     The higher the value, the more rounded the corners will become.
```

PR #2271, opened 2024-12-02
Correct SubtitlesClip example to not pass the generator lambda as `font`

akx:
Refs: https://stackoverflow.com/a/79243946/51685

Given the signature for `SubtitlesClip.__init__()` is `__init__(self, subtitles, font=None, make_textclip=None, encoding=None)`, the example would have attempted to pass in the generator function as `font`.


OsaAjani:
Nice catch !

PR #2282, opened 2024-12-05
Fix issue #2160 based on PR #2163 relative to filenames starting with…

OsaAjani:
… a dash and improved for filenames with specialchars
Writing to or reading from a filename starting with a dash "-" raised an error as described in #2160, this a port and improvement over the original #2163 PR.


OsaAjani:
I need to fix this for windows, I will take a look when I can.

PR #2292, opened 2024-12-18
Fix/documents

hasnainmohiuddin:
Documentation Changes

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,20 @@
+# This CITATION.cff file was generated with cffinit.
+# Visit https://bit.ly/cffinit to generate yours today!
+
+cff-version: 1.2.0
+title: MoviePy
+message: >-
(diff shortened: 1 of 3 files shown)


OsaAjani:
I feel like it's not our place to add a citation guideline but that it should be up to @Zulko to decide of the rules of citation.
I also feel like the code of conduct is unnecessary (though I totally agree with the content), and mostly I must say I dont really like that this PR only talk about "documentation changes" when I think (might be wrong obviously) really nothing relate to documentation and everything relate to code of conduct...

I would vote for closing this one, I let a few days go and if without counter arguments will proceed to it.

OsaAjani:
Last call on this one, I will close tomorrow or the day after without counter arguments.

OsaAjani:
With nobody arguing, I close this !

PR #2296, opened 2024-12-23
fixing ffmpeg reader bug on specific frames

alimirjahani7:


OsaAjani:
I feel like this may try to fix the same problem as pr #2117. What do you think ?

OsaAjani:
Without response I consider this is trying to achieve the same as #2117 and merge it.

Issue #2305, opened 2024-12-29
SubtitlesClip TypeError: 'float' object cannot be interpreted as an integer


yeohongred:
Trying to create a SubtitlesClip without make_textclip() causes a TypeError

#### Steps to Reproduce the Problem
```
from moviepy.video.tools.subtitles import SubtitlesClip
from moviepy import VideoFileClip, CompositeVideoClip

video = VideoFileClip(filename="video.mp4")
subtitles = SubtitlesClip(subtitles="subtitles.srt", font="font.otf") # ERROR
output = CompositeVideoClip([video, subtitles])
output.write_videofile("output.mp4", fps=24)
```

```
Traceback (most recent call last):
  File "c:\Users\Yeo Hong\Desktop\test\test.py", line 6, in <module>
    subtitles = SubtitlesClip(subtitles="subtitles.srt", font="font.otf")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Yeo Hong\Desktop\test\.venv\Lib\site-packages\moviepy\video\tools\subtitles.py", line 123, in __init__
    hasmask = bool(self.make_textclip("T").mask)
                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Yeo Hong\Desktop\test\.venv\Lib\site-packages\moviepy\video\tools\subtitles.py", line 76, in make_textclip
    return TextClip(
           ^^^^^^^^^
  File "C:\Users\Yeo Hong\Desktop\test\.venv\Lib\site-packages\decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Yeo Hong\Desktop\test\.venv\Lib\site-packages\moviepy\decorators.py", line 94, in wrapper
    return func(*new_args, **new_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Yeo Hong\Desktop\test\.venv\Lib\site-packages\moviepy\video\VideoClip.py", line 1763, in __init__
    draw.multiline_text(
  File "C:\Users\Yeo Hong\Desktop\test\.venv\Lib\site-packages\PIL\ImageDraw.py", line 775, in multiline_text
    self.text(
[...]

PR #2306, opened 2024-12-29
Fix issue #2305: Change stroke_width from float 0.5 to int 1

yeohongred:
Trying to create a SubtitlesClip without make_textclip() causes a TypeError. This TypeError is caused by the default make_textclip() providing a float value of 0.5 for stroke_width when it can only accept int values. Fixed by changing value of stroke_width to 1.

### DIFF
--- a/moviepy/video/tools/subtitles.py
+++ b/moviepy/video/tools/subtitles.py
@@ -72,6 +72,7 @@ class SubtitlesClip(VideoClip):
             if self.font is None:
                 raise ValueError("Argument font is required if make_textclip is None.")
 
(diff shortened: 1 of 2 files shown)


moviepy/video/tools/subtitles.py
```diff
@@ -72,14 +72,15 @@ def __init__(self, subtitles, font=None, make_textclip=None, encoding=None):
             if self.font is None:
                 raise ValueError("Argument font is required if make_textclip is None.")
 
+            # Changed stroke_width from float 0.5 to int 1
```
OsaAjani, line 75: Please, remove this comment as its not relevant out of the PR

Issue #2330, opened 2025-01-17
AudioClip write_audiofile fps setting bug


Implosiv3:
I think I found a bug in the `AudioClip` class `write_audiofile` method. This is the code in the `line 241`:

```
if not fps:
    if hasattr(self, "fps"):
        fps = 44100
    else:
        fps = self.fps
```

And I think it should be this one:

```
if not fps:
    if hasattr(self, "fps"):
        fps = self.fps
    else:
        fps = 44100
```

OsaAjani:
Close by your PR

PR #2336, opened 2025-01-20
Fix issue #2330 audio write wrong fps

Implosiv3:
Fixes a bug that is not setting the instance 'fps' value when it has the attribute, but doing it when it doesn't.

This is the related issue: https://github.com/Zulko/moviepy/issues/2330

If it has the attribute, set it, if not, use a default value.


OsaAjani:
Thanks, that seems a lot more logical !

PR #2345, opened 2025-01-25
Update config.py check() to log correct binary name

cerushdope:
When finding FFMPEG_BINARY, it correctly displays that it found ffmpeg. But when finding FFPLAY_BINARY it again displays ffmpeg. This changes it to display ffplay


OsaAjani:
Thanks for your work

PR #2384, opened 2025-02-25
make dependency management compatible with uv

07pepa:
this is small change  allowing to use [uv](https://docs.astral.sh/uv/) as dependecy manager on this project. With automatic updating of lockfile

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,7 @@ requires = ["setuptools>=42", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]
+requires-python=">=3.9"
 name = "moviepy"
 version = "2.2.0"
 description = "Video editing with Python"


07pepa:
I do understand.

Then  Please let me also propose sort of compromise for you

just keep changes in pyproject.toml
line:
`requires-python=">=3.9"`

is basicaly all uv need to work (if i remember corectly it is also used by other tools like ruff)

That is minimum this pr can be reduced to.

OsaAjani:
Sure no problem on that part, happy to :)

07pepa:
Lets wait for up to 3 weeks  for response... i can wait (this is low prio that just helps me make future PRs). 

and in case of no response i will commit  reduction to just pyproject.toml

In that case support for uv "will be"  but "unofficial" and only extra burden is updating requires_python when deprecating old python versions.

@OsaAjani  thank you for maintaining such simple  useful wrapper around fantastic ffmpeg

OsaAjani:
Sure

Le 23 juin 2025 22:10:47 GMT+02:00, Pepa ***@***.***> a écrit :

keikoro:
Familiarising myself with `uv` is actually on my TODO list for non-personal stuff. I'll check back in here as soon as I can contribute more informedly!

keikoro:
> having just .python-version causes uv to emit warning
> ```
> warning: No `requires-python` value found in the workspace. Defaulting to `>=3.12`
> ```

I see!


> commits have been squashed into one  and rebased on top of masted so pr should be ready to merge

Great, thanks!


> > so it would make sense to actually add a corresponding Python requirement to `pyproject.toml`.
> 
> `requires-python=">=3.9"` is added already

Yup, I was more commenting on my own earlier comment, saying that your change, this addition of the Python requirement, makes sense either way since we already reference it in the docs.

@OsaAjani since you already confirmed upthread that this addition should be fine, I'm accepting it.

keikoro:
Thanks, @07pepa !

Issue #2395, opened 2025-03-13
v1 is 10x faster than v2 for this example


rrherr:
My example is writing a 3 minute long video, horizontally scrolling through a wide image.

- [moviepy 2.1.2 took 18 minutes 32 seconds](https://colab.research.google.com/gist/rrherr/d87d0c803e2bba7af8620797436a1c3c/moviepy2.ipynb)
- [moviepy 1.0.3 took 1 minute 39 seconds](https://colab.research.google.com/gist/rrherr/01a8f78b0d94b311746d563eab0dd8f6/moviepy1.ipynb) (over 10x faster)

(The links above are to Colab Notebooks showing both the code and results.)

Is this expected?

I'm a new MoviePy user and I think it's awesome, so thank you to all who've helped develop it and maintain it!
I'm just trying to figure out whether to stay on verison 1 or pursue faster performance on version 2. Thanks in advance!

OsaAjani:
Definitely not expected, I will take a look when I have some times :)

seekme-ai:
same issue

rrherr:
Inspired by #2403, I also ran my [example with 2.1.1](https://colab.research.google.com/gist/rrherr/384a7b0f29a1ec79ced51eb706d9427c/moviepy211.ipynb). Here's my results:

```mermaid
xychart-beta
    title "MoviePy rendering time for my video"
    x-axis "MoviePy version" [1.0.3, 2.1.1, 2.1.2]
    y-axis "Time (seconds)"
    bar [99, 731, 1112]
```

iAndyHD3:
2.1.1 is painfully slow, I would like to port my code for 1.0.3 to compare but really should not be this slow.
It's a one minute video put together by concatenating like 10 videos each of them with a static text, to render it takes 5 minutes on a really good CPU

JayChen7777:
same issue. I want to use TextClip, but both 2.1.1 and 1.0.3 can't handle this, so I switched to 2.1.2. It works, but it's very, very slow.

Kopamed:
This is so cooked. V2 eats up all my memory when loading and concatenating an array of clips with a total duration of 2 minutes.
I have 32 GB RAM. Eats up all of that. And then another 30 gigs of swap. Same logic worked fine in v1 eating up much less memory

trentcowden:
Experiencing pretty slow render times as well, especially when using nested `CompositeVideoClips`. Would love some improvements here as well :)

liangsix:
[...]

PR #2416, opened 2025-04-12
ImageClip: copy base image. Improves performance in some cases.

bzczb:
Helps with performance in #2395 test to some extent.

If you check the `ndarray.flags` of certain frames, you see that `OWNDATA` and `C_CONTIGUOUS` are sometimes false, which I think is slowing down something in PIL a lot. The copy is improving performance on the #2395 test by about 3x for me, on OS X ARM.

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -1341,9 +1341,9 @@ class ImageClip(VideoClip):
 
         # if the image was just a 2D mask, it should arrive here
         # unchanged
-        self.frame_function = lambda t: img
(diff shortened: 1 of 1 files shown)


OsaAjani:
Hi, could you provide some example code + media that illustrate the performances gap ? I would like to try and reproduce and see how pr #2490 influence this.

bzczb:
> Hi, could you provide some example code + media that illustrate the performances gap ? I would like to try and reproduce and see how pr #2490 influence this.

It was the example given in #2395. I ran the same moviepy2 example twice, with the only difference being the change in this PR.

PR #2417, opened 2025-04-13
Update VideoClip.py to support Pillow<10.0.0

RaSan147:
- [x] I have tested with existing test cases

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -1881,7 +1881,12 @@ class TextClip(ImageClip):
         draw = ImageDraw.Draw(img)
 
         # Compute individual line height with spaces using pillow internal method
-        line_height = draw._multiline_spacing(font_pil, spacing, stroke_width)
+        try:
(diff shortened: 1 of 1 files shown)


RaSan147:
also `Pillow<10.0.0` doesn't support the default font with size

```python
       ...
            font_pil = ImageFont.load_default(font_size)
        ascent, descent = font_pil.getmetrics()
```

these 2 lines are not functional in older Pillow. please do something. Either unsupport older versions, or make custom implementation or a default font packaged with moviepy.

keikoro:
What does this solve that the [version bump to Pillow](https://github.com/Zulko/moviepy/blob/4dd5fe44b47a4a2f42eebc8b5eadd74a25001853/pyproject.toml#L36) doesn't solve?

RaSan147:
They changed a lot in the get font API from v10.0.0, and those changes aren't reflected in the code. That version bump alone didn't solve the problem; those are some breaking changes. The changelog URL is mentioned in the code comment.

RaSan147:
💀 After seeing the latest patch, still doesn't solve the HalfCut hello world test issue.

PR #2483, opened 2025-07-21
Fixed_ArchLinux_preview_not_working

ZippyBagi:
On Arch linux moviepy.video.io.ffmpeg_tools.tools.ffplay_versions() always returned an error - this made it so that the preview function of clips wasn't working. On arch ffmpeg versions start with an "n" and then the version number. I have fixed this, now it works. This bug only occured on arch linux as far as I can tell.

- [YES ] I have properly explained unusual or unexpected code in the comments around it

### DIFF
--- a/moviepy/video/io/ffmpeg_tools.py
+++ b/moviepy/video/io/ffmpeg_tools.py
@@ -328,5 +328,8 @@ def ffplay_version():
     result = subprocess.run(cmd, capture_output=True, text=True, check=True)
     # Extract the version number from the first line of output
     full_version = result.stdout.splitlines()[0].split()[2]
+    #On Arch linux ffmpeg versions start with n, then the version number, this makes sure that no errors occur.
+    if full_version.startswith("n"):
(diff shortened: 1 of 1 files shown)


moviepy/video/io/ffmpeg_tools.py
```diff
@@ -328,5 +328,8 @@ def ffplay_version():
     result = subprocess.run(cmd, capture_output=True, text=True, check=True)
     # Extract the version number from the first line of output
     full_version = result.stdout.splitlines()[0].split()[2]
+    #On Arch linux ffmpeg versions start with n, then the version number, this makes sure that no errors occur.
```
keikoro, line 331: I can't comment on the code side of this, but would ask you clean up the comment. Spelling of proper names and punctuation should be fixed (missing whitespace after # for consistency with other comments; use full stop or semi-colon to split the sentence in two for readability).

keikoro (changes requested):

PR #2524, opened 2025-09-30
Fixed a bug in Scroll.py

djp3:
Fix a bug in the Scroll Effect.

Basically it wasn't working.  The scroll effect was stopping before it should and would sometimes freak out and just start rolling over the content really quickly.

- [✅] I have provided code that clearly demonstrates the bug and that only works correctly when applying this fix
- [✅ ] I have added suitable tests demonstrating a fixed bug or new/changed feature to the test suite in `tests/`
- [✅] I have properly documented new or changed features in the documentation or in the docstrings
- [✅ ] I have properly explained unusual or unexpected code in the comments around it

Here is a python program that contains the corrected code that can be run against the current code to show the bugs.
[test_scroll_effect.py](https://github.com/user-attachments/files/22626043/test_scroll_effect.py)

I put a stripped down version of that file into the tests directory.

### DIFF
--- a/moviepy/video/fx/Scroll.py
+++ b/moviepy/video/fx/Scroll.py
@@ -46,8 +46,18 @@ class Scroll(Effect):
         if self.w is None:
             self.w = clip.w
 
-        x_max = self.w - 1
-        y_max = self.h - 1
+        # If the clip is smaller than the output, there isn't anything to scroll over.
+        # If the clip is bigger than the desired output then there is something to scroll over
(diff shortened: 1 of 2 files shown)


keikoro:
@djp3 Please include any "fixes" to your own PR within the original commit(s), i.e. rebase instead of tacking on more commits. And use descriptive commit messages for the original changes.

djp3:
Tried to squash it as you requested.  I'm not 100% sure that I got the workflow right though.

djp3:
Clean merge for the win!

PR #2530, opened 2025-11-05
[Feature Request] Proper text breaking by words in TextClip #2501

### LINKED ISSUE #2501: Subtitles+TextClip wrap by word
I'm using a TextClip (with method caption to autowrap) but currently is autowrapping by character, is possible to autowrap by word instead?
Thanks.

LasseSchmidt:
Fixed a bug where text breaking by words works for first line but not the lines after.

```
import moviepy as mp

text = 'MoviePy (online documentation here) is a Python library for video editing: cuts, concatenations, title insertions, video compositing (a.k.a. non-linear editing), video processing, and creation of custom effects. MoviePy can read and write all the most common audio and video formats, including GIF, and runs on Windows/Mac/Linux, with Python 3.9+.'

text_lines = mp.TextClip(text=text, font_size=20)._TextClip__break_text(width=500, text=text, font=None, font_size=20, stroke_width=0, align="left", spacing=4)
print("\n".join(text_lines))
```

### DIFF
--- a/moviepy/video/VideoClip.py
+++ b/moviepy/video/VideoClip.py
@@ -1908,6 +1908,7 @@ class TextClip(ImageClip):
         # We try to break on spaces as much as possible
         # if a text dont contain spaces (ex chinese), we will break when possible
         last_space = 0
+        last_break = 0
         for index, char in enumerate(text):
             if char == " ":
                 last_space = index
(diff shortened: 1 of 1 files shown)


JacobRosey:
i applied your fix locally by overriding the break_text method and it works great, thanks!

PR #2544, opened 2025-12-25
Fix issue 2539 ffmpeg pixel_format option for libx264 in FFMPEG_VideoWriter class

### LINKED ISSUE #2539: The 'pixel_format' option in moviepy.video.VideoClip.VideoCip.write_videofile() is ignored when video_codec is set to 'libx264'
The 'pixel_format' option in `moviepy.video.VideoClip.VideoCip.write_videofile()` is ignored when `video_codec` is set to 'libx264'. This can be seen on line 162 of `moviepy/video/io/ffmpeg_writer.py` in hash 7ffa4f00376237137a25fe1c777355c37753e9af on main. 

Handling all potential translations would require a static mapping of all ffmpeg modes to libx264 supported modes, which is outside the scope of this issue. Given that a professional user would likely already know to specify a yuv format, a better solution would be to replace any format that is not a yuv derivative with the default and pass through specified yuv formats unchanged.  As well, h264 does not support an alpha channel, so no need for yuva. A proposed patch would be:

```
diff --git a/moviepy/video/io/ffmpeg_writer.py b/moviepy/video/io/ffmpeg_writer.py
index 71c301c..7c528dd 100644
--- a/moviepy/video/io/ffmpeg_writer.py
+++ b/moviepy/video/io/ffmpeg_writer.py
@@ -159,7 +159,10 @@ class FFMPEG_VideoWriter:
             and (size[0] % 2 == 0)
             and (size[1] % 2 == 0)
         ):
-            cmd.extend(["-pix_fmt", "yuva420p"])
+            if pixel_format is None or not pixel_format.startswith('yuv'):
+                cmd.extend(["-pix_fmt", "yuv420p"])
+            else:
+                cmd.extend(["-pix_fmt", pixel_format])
         else:
             # For all other codecs, we use the pixel format specified by the user
             # or the default one.
```

#### Expected Behavior

The 'pixel_format' option should be honored when set, within the limitations of the codec.

#### Actual Behavior

The pixel format is always set to 'yuv420p'

#### Steps and code to Reproduce the Problem

```
import numpy as np
from PIL import Image
from moviepy import ImageClip
import ffmpeg
import os

image = Image.new('RGB', (512, 512), color=(255, 0, 0))
image_array = np.array(image)
clip = ImageClip(image_array).with_fps(24, change_duration=False).with_duration(5)
clip.write_videofile(filename='x264_yuv444_test.mp4', codec='libx264', pixel_format='yuv444p')
probe = ffmpeg.probe('x264_yuv444_test.mp4')
format = probe['streams'][0]['pix_fmt']
print(f'Pixel format used in the generated video: {format}')
os.remove('x264_yuv444_test.mp4')
```

Note that whatever you set 'pixel_format' to, it will always result in a 'yuv420' encoding.

#### Used medias

N/A

#### Specifications

  - Python Version: 3.12.11
  - MoviePy Version: 2.2.1
  - Platform Name: OSX
  - Platform Version: 15.6.1

kevinmagnopus:
- [ x] I have provided code that clearly demonstrates the bug and that only works correctly when applying this fix (repro code is in the issue [here](https://github.com/Zulko/moviepy/issues/2539))
- [ x] I have added suitable tests demonstrating a fixed bug or new/changed feature to the test suite in `tests/`
- [ x] I have properly documented new or changed features in the documentation or in the docstrings
- [ x] I have properly explained unusual or unexpected code in the comments around it

### DIFF
--- a/moviepy/video/io/ffmpeg_writer.py
+++ b/moviepy/video/io/ffmpeg_writer.py
@@ -61,9 +61,11 @@ class FFMPEG_VideoWriter:
       Set to ``True`` if there is a mask in the video to be encoded.
 
     pixel_format : str, optional
-      Optional: Pixel format for the output video file. If is not specified
+      Optional: Pixel format for the output video file. If not specified
       ``"rgb24"`` will be used as the default format unless ``with_mask`` is
-      set as ``True``, then ``"rgba"`` will be used.
+      set as ``True``, then ``"rgba"`` will be used. If the codec is
+      ``libx264`` or ``h264_nvenc``, the pixel format will be set to
+      ``"yuv420p"`` by default, and will only accept other ``yuv`` formats.
 
     logfile : int, optional
       File descriptor for logging output. If not defined, ``subprocess.PIPE``
@@ -160,7 +162,7 @@ class FFMPEG_VideoWriter:
             and (size[1] % 2 == 0)
         ):
             # h264 encoder only supports yuv pixel formats, so allow if requested
-            if pixel_format is None or not pixel_format.startswith('yuv'):
+            if pixel_format is None or not pixel_format.startswith("yuv"):
                 cmd.extend(["-pix_fmt", "yuv420p"])
             else:
                 cmd.extend(["-pix_fmt", pixel_format])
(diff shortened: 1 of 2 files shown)


kevinmagnopus:
@Zulko may I request that you please review this PR. This PR addresses a bug where it is not possible to generate a lossless h264 video. We have a use case for lossless h264, but we currently unable to achieve it due to the bug addressed in this PR. Thank you.

Issue #2358, opened 2025-02-01
ERROR While using SubtitlesClip: multiple values for argument 'font'


abhishek-shivale:
`  speech_audio = AudioFileClip(audio_path)

        # Load banner image and set duration
        banner = (
            ImageClip(banner_path)
            .with_duration(speech_audio.duration)
            .resized((1920, 1080))
        )
        # Load and style subtitles from .srt file
        font="/usr/share/fonts/truetype/dejavu/DejaVuSerif-Bold.ttf"
        
        subtitle_style = lambda text : TextClip(
            text,  
            font_size=45,  
            color="white", 
            font=font, 
            stroke_color="black", 
            stroke_width=2,  
            size=(1800, 300),  
            method="caption",  
            text_align="center",
            horizontal_align="center",
            vertical_align="bottom",
            bg_color=(0, 0, 0, 0),
        )

        subtitles = SubtitlesClip(subtitles=subtitle_path, make_textclip=subtitle_style, encoding="utf-8")

        # Combine video elements
        final_clip = CompositeVideoClip([banner, subtitles])
`
i tried different approach like without make_textclip is working but when i add make_textclip its giving saying 

` multiple values for argument 'font'
Done
Traceback (most recent call last):
  File "/home/blackpearl/Desktop/ai/bot/youtube/youtube.py", line 69, in <module>
    asyncio.run(get_video_info())
  File "/home/blackpearl/anaconda3/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/home/blackpearl/anaconda3/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/blackpearl/anaconda3/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/blackpearl/Desktop/ai/bot/youtube/youtube.py", line 42, in get_video_info
    subtitles = SubtitlesClip(subtitles=subtitle_path, make_textclip=subtitle_style, encoding="utf-8")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2551, opened 2026-02-05
Fix subtitles documentation

hannahfluch:
The old example in  the doc comments of subtitles did not compile and was generally error-prone (#2358 the issue here is also present in the old example)

I added a working example and updated the doc-comment to be correct.

### DIFF
--- a/examples/subtitles.py
+++ b/examples/subtitles.py
@@ -1,0 +1,13 @@
+from moviepy.video.tools.subtitles import SubtitlesClip
+from moviepy import TextClip, CompositeVideoClip, VideoFileClip
+
+
+# generate custom textclips for the subtitles
+generator = lambda text: TextClip(text=text, font="../media/doc_medias/example.ttf", font_size=24, color="white")
+subtitles = SubtitlesClip("../media/subtitles.srt", make_textclip=generator, encoding="utf-8")
+
+# video clip
+clip = VideoFileClip("../media/chaplin.mp4")
+
+final = CompositeVideoClip([clip, subtitles])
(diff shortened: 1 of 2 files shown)


moviepy/video/tools/subtitles.py
```diff
@@ -43,13 +43,18 @@ class SubtitlesClip(VideoClip):
...
+        )
+        subtitles = SubtitlesClip("media/subtitles.srt", make_textclip=generator, encoding="utf-8")
+        clip = VideoFileClip("./media/chaplin.mp4")
+        final = CompositeVideoClip([clip, subtitles])
+        final.write_videofile("final.mp4", fps=clip.fps)
+        subtitles = SubtitlesClip("subtitles.srt", make_textclip=generator, encoding='utf-8')
+        clip = VideoFileClip("myvideo.avi")
         final = CompositeVideoClip([clip, subtitles])
-        final.write_videofile("final.mp4", fps=myvideo.fps)
+        final.write_videofile("final.mp4", fps=clip.fps)
```
keikoro, line 57: Why do `subtitles`, `clip` and `final` get set multiple times?

examples/subtitles.py
keikoro, line 1: You mentioned an old example that "did not compile and was generally error-prone" but i only see a new addition here, no deletion of existing code (comments) containing the old example.

keikoro (commented):
@hannahfluch Your original submission failed the code formatting/linting test though there's no way now to check what the problem was.

When you rebase and push to the PR again, the tests can be run again.

PR #2554, opened 2026-02-14
fix AttributeError on CompositeVideoClip with non-composite mask

### LINKED ISSUE #2514: 'CompositeVideoClip' object has no attribute 'precomputed'
#### Expected Behavior

`CompositeVideoClip` should be able to use a mask created from `VideoClip.to_mask()` without raising an `AttributeError`, even when `memoize_mask=True`.

#### Actual Behavior

When applying a mask created from `to_mask()` to a `CompositeVideoClip`, the following error occurs:

```
AttributeError: 'CompositeVideoClip' object has no attribute 'precomputed'
```

This happens because `CompositeVideoClip` assumes the mask clip has a `.precomputed` attribute, but the `.precomputed` dictionary is currently defined on the `CompositeVideoClip` itself, not on the mask clip.

https://github.com/Zulko/moviepy/blob/70f6917138768f9b8458ce0e1a4d7b752e19ed92/moviepy/video/compositing/CompositeVideoClip.py#L154C28-L154C34

https://github.com/Zulko/moviepy/blob/70f6917138768f9b8458ce0e1a4d7b752e19ed92/moviepy/video/compositing/CompositeVideoClip.py#L207

Mr-Neutr0n:
When you set a mask on a `CompositeVideoClip` that isn't itself a `CompositeVideoClip` (e.g. created via `to_mask()`), calling `frame_function` crashes with `AttributeError: 'CompositeVideoClip' object has no attribute 'precomputed'`.

The `precomputed` dict only gets created on `CompositeVideoClip` instances where `is_mask=True` and `memoize_mask=True` (in `__init__`). But `frame_function` unconditionally accesses `self.mask.precomputed` when `self.mask` and `self.memoize_mask` are truthy, without checking whether the mask object actually has that attribute.

The fix just adds a `hasattr` guard so we only try to memoize when the mask supports it.

Fixes #2514


Mr-Neutr0n:
Closing this one to tidy up my open pull requests.

It's been open around five months with no review activity, and I'd rather withdraw it than leave it sitting in your queue. Nothing needed from you, and no hard feelings at all.

If the fix is still wanted, this can be reopened, or I'm happy to redo it properly against current main. Apologies for the noise.

Issue #2481, opened 2025-07-18
Wrong value of moviepy.__version__ in 2.2.1


gnuyens:
#### Expected Behavior
__version__

#### Steps and code to Reproduce the Problem

```

from moviepy import *
(mp2) greg@Mac MoviePy % pip show moviepy
Name: moviepy
Version: 2.2.1
Summary: Video editing with Python
Home-page: 
Author: Zulko 2024
Author-email: 
License: MIT License
Location: /Users/greg/envs/mp2/lib/python3.9/site-packages
Requires: decorator, imageio, imageio_ffmpeg, numpy, pillow, proglog, python-dotenv
Required-by: 

but

(mp2) greg@Mac MoviePy % python3
Python 3.9.6 (default, Apr 30 2025, 02:07:17) 
[Clang 17.0.0 (clang-1700.0.13.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from moviepy import *
>>> print(f'Moviepy imported, version: {__version__}')
Moviepy imported, version: 2.1.2
>>> ^D

The installed source for version.py assigns this 2.1.x  result instead of the correct 2.2.x value, it seems 
The repo shows more complicated code, but still with the old version string.
```

#### Specifications

  - Python Version: 3.9
  - MoviePy Version: 2.2.1
  - Platform Name: mac os 
  - Platform Version: 15.5

PR #2555, opened 2026-02-21
fix: update stale fallback version in version.py

danielalanbates:
## Summary

- Updates the hardcoded fallback version in `moviepy/version.py` from `"2.1.2"` to `"2.2.0"` to match `pyproject.toml`
- Adds a comment reminding maintainers to keep the fallback in sync with `pyproject.toml`

## Problem

When `importlib.metadata.version("moviepy")` fails (e.g., when running from source without installing the package), `moviepy.__version__` falls back to a hardcoded string. This string was never updated past `"2.1.2"`, so users see the wrong version number.

Reported in #2481.

## Test plan

- [x] Verify `moviepy/version.py` fallback now reads `"2.2.0"`
- [ ] Run `python -c "from moviepy.version import __version__; print(__version__)"` from source (without pip install) and confirm it reports `2.2.0`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


danielalanbates:
Thank you for your time reviewing this. We're withdrawing this PR. Apologies for any inconvenience.

Issue #2374, opened 2025-02-14
Closing Preview window causes traceback because of broken pipe


jmansilla:
This simple snippet evidences the problem

```
from moviepy import VideoFileClip
clip = VideoFileClip(movie_path)
clip.preview()
clip.close()
```

While the video is been played, proceed to close the window.

#### Expected Behavior
Program closes silently

#### Actual Behavior

This traceback is printed

```
/Users/jmansilla/.virtualenvs/myproject/lib/python3.12/site-packages/imageio_ffmpeg/binaries/ffmpeg-macos-x86_64-v7.1 -i movie.mp4 -loglevel error -f image2pipe -vf scale=1280:720 -sws_flags bicubic -pix_fmt rgb24 -vcodec rawvideo -
Traceback (most recent call last):
  File "/Users/jmansilla/.virtualenvs/myproject/lib/python3.12/site-packages/moviepy/video/io/ffplay_previewer.py", line 59, in show_frame
    self.proc.stdin.write(img_array.tobytes())
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/jmansilla/.virtualenvs/myproject/bin/mycommand", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/jmansilla/projects/repo/myproject/ui/cli.py", line 76, in main
    play_movie(movie_path)
  File "/Users/jmansilla/projects/repo/myproject/ui/cli.py", line 87, in play_movie
    clip.preview()
  File "/Users/jmansilla/.virtualenvs/myproject/lib/python3.12/site-packages/decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jmansilla/.virtualenvs/myproject/lib/python3.12/site-packages/moviepy/decorators.py", line 53, in requires_duration
    return func(clip, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jmansilla/.virtualenvs/myproject/lib/python3.12/site-packages/decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2573, opened 2026-07-14
Stop preview when FFplay closes normally

### LINKED ISSUE #2535: MoviePy clip preview interruption problem
If during a preview with `clip.preview()` the user presses [Esc] or [Q], MoviePy crashes, issuing the following error: **"MoviePy error: FFPLAY encountered the following error while previewing clip: None"**. 
Now,  to my knowledge, FFPLAY never crashes by pressing these keys.
Anyway, can this be prevented in some way?

janmarsino98:
## Summary

Fixes #2535 and #2374. When a user closes FFplay's preview window, the next stdin write raises a broken-pipe `IOError`. If FFplay exited normally (return code 0), the preview now stops cleanly. Nonzero FFplay exits still raise an actionable `IOError`.

The previewer's close path also tolerates stdin already being closed by `communicate()`. If video exits before its first frame while audio is enabled, it signals `video_flag` so the audio preview thread is not left waiting.

## Validation

- Added mocked tests for normal user-close, a genuine FFplay failure, and an already-closed stdin.
- `pytest tests/test_ffplay_previewer.py -q` (3 passed)
- `pytest -q` (604 passed, 7 skipped, 1 xpassed)
- `pre-commit run --files moviepy/video/io/ffplay_previewer.py tests/test_ffplay_previewer.py`

- [x] I have added suitable tests demonstrating a fixed bug to the test suite in `tests/`


keikoro:
@olfMombach Since you commented in one of the linked issues, could you have a look at this?

Issue #2581, opened 2026-07-30
Numpy Deprecation Warning During Reshape


Kaszanas:
#### Expected Behavior

Reading a frame from a video (e.g. via `VideoFileClip.get_frame()`) should not emit any warnings on a supported, up-to-date `numpy` version.

#### Actual Behavior

Every raw-frame decode emits a `DeprecationWarning` on numpy >= 2.5:

```
moviepy/video/io/ffmpeg_reader.py:228: DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
As an alternative, you can create a new view using np.reshape (with copy=False if needed).
  result.shape = (h, w, len(s) // (w * h))  # reshape((h, w, len(s)//(w*h)))
```

The cause is `FFMPEG_VideoReader.read_frame()` in `moviepy/video/io/ffmpeg_reader.py`, which assigns directly to `ndarray.shape` instead of using `np.reshape`/`.reshape()`:

```python
result.shape = (h, w, len(s) // (w * h))  # reshape((h, w, len(s)//(w*h)))
```

numpy 2.5 deprecated in-place shape assignment on arrays; a future numpy release is expected to turn this into a hard error. Since this line runs on every frame decode, it's the single most-triggered warning across the whole test suite (it showed up in every test that reads real video frames).

Suggested fix — replace the in-place assignment with a reshape call:

```python
result = result.reshape((h, w, len(s) // (w * h)))
```

#### Steps and code to Reproduce the Problem

This is self-contained — it generates its own tiny video, so no external media is needed:

```python
import warnings
warnings.simplefilter("always")

from moviepy import ColorClip, VideoFileClip

clip = ColorClip((64, 64), color=(255, 0, 0), duration=1)
clip.write_videofile("repro.mp4", fps=24, logger=None)

reader_clip = VideoFileClip("repro.mp4")
frame = reader_clip.get_frame(0)  # <-- DeprecationWarning fires here
print("frame shape:", frame.shape)
```

Output:

```
C:\...\moviepy\video\io\ffmpeg_reader.py:228: DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
As an alternative, you can create a new view using np.reshape (with copy=False if needed).
  result.shape = (h, w, len(s) // (w * h))  # reshape((h, w, len(s)//(w*h)))
frame shape: (64, 64, 3)
```

#### Used medias

[...]

PR #2582, opened 2026-07-30
[PR] Numpy Deprecation Fix

Kaszanas:
## Summary

Fixes a `DeprecationWarning` that numpy >= 2.5 raises on **every single frame decode**.

`FFMPEG_VideoReader.read_frame()` in `moviepy/video/io/ffmpeg_reader.py` built each frame from the raw bytes read off the ffmpeg subprocess, then reshaped the flat buffer by assigning directly to `result.shape`:

```python
result.shape = (h, w, len(s) // (w * h))
```

numpy 2.5 deprecated in-place shape assignment on an `ndarray` in favor of `.reshape()`, so this fired on every frame read:

```
DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
As an alternative, you can create a new view using np.reshape (with copy=False if needed).
```

## Reproduction

Self-contained — it generates its own tiny video, so no external media is needed. On the pre-fix code this prints the `DeprecationWarning` shown above before printing the frame shape; after the fix it prints only the frame shape.

```python
import warnings
warnings.simplefilter("always")

from moviepy import ColorClip, VideoFileClip

clip = ColorClip((64, 64), color=(255, 0, 0), duration=1)
clip.write_videofile("repro.mp4", fps=24, logger=None)

reader_clip = VideoFileClip("repro.mp4")
frame = reader_clip.get_frame(0)  # <-- DeprecationWarning fires here pre-fix
print("frame shape:", frame.shape)
```

## Checklist

- [x] I have provided code that clearly demonstrates the bug and that only works correctly when applying this fix — *not included in the diff; reproduced and verified manually during development instead.*
As shown above and within the https://github.com/Zulko/moviepy/issues/2581

- [x] I have added suitable tests demonstrating a fixed bug or new/changed feature to the test suite in `tests/` — *none added; the existing `test_sequential_frame_pos` exercises the code path but doesn't specifically assert the warning is absent.*
All of the previous tests are running, no new tests needed.

- [x] I have properly documented new or changed features in the documentation or in the docstrings — *CHANGELOG.md entry added.*
- [x] I have properly explained unusual or unexpected code in the comments around it — *N/A, delete: the fix removes a now-redundant comment rather than adding anything that needs explaining.*
Not applicable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,5 @@ docs/build/
 
 # Publishing
 .pypirc
+
+.venv/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This mean you need to specifically set `bg_color=None` if you want to generate a
 - Fix TextClip broken with Pillow > 11.2
 - `pixel_format` parameter was ignored when calling ffmpeg writer, as referenced in PR #2359
 - Fix incorrect handling of lines with a format different from "key: value" during FFmpeg infos parsing (see #2311, #1860, #2418, #2470)
+- Fix `DeprecationWarning` from in-place `ndarray.shape` assignment in `FFMPEG_VideoReader.read_frame()` on numpy >= 2.5
 
 ### Added
 - Possibility to select audio track when reading a file (#2429)
--- a/moviepy/video/io/ffmpeg_reader.py
+++ b/moviepy/video/io/ffmpeg_reader.py
@@ -225,7 +225,7 @@ class FFMPEG_VideoReader:
                 result = np.frombuffer(s, dtype="uint8")
             else:
                 result = np.fromstring(s, dtype="uint8")
-            result.shape = (h, w, len(s) // (w * h))  # reshape((h, w, len(s)//(w*h)))
+            result = result.reshape((h, w, len(s) // (w * h)))
             self.last_read = result
 
         # We have to do this down here because `self.pos` is used in the warning above


keikoro:
Your PR touches files that have nothing to do with what it pertains to do.

Closing as unreviewable suspected AI vibecoding.

Kaszanas:
> Your PR touches files that have nothing to do with what it pertains to do.
> 
> Closing as unreviewable suspected AI vibecoding.

Could you please read the diff?

1. Adding `.venv/` to gitignore because maybe it should not be in .git?
2. Fixing the issue?
3. Adding a record to changelog?

Have you used moviepy anywhere recently yourself, I am reading 8000 deprecation warnings per automation run on my side.

Otherwise, please actively maintain your software. Have you at least tried the reproduction?

keikoro:
I looked at the diff.
The PR shows changes for 3 files, 2 of which can't possibly have anything to do with what it says it does.

Have you looked at the PRs this repo has been getting? All these AI submissions are a review nightmare.

Kaszanas:
> I looked at the diff. The PR shows changes for 3 files, 2 of which can't possibly have anything to do with what it says it does.
> 
> Have you looked at the PRs this repo has been getting? All these AI submissions are a review nightmare.

Correct, I had to change these files for local development on my fork. Can't do much about it!

This is from my test suite.

```
tests/anon.py: 1456 warnings
  anon.py:226: DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
  As an alternative, you can create a new view using np.reshape (with copy=False if needed).
    result.shape = (h, w, len(s) // (w * h))  # reshape((h, w, len(s)//(w*h)))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

I was off by about a thousand (69 is accidental):
```
==== 93 passed, 9069 warnings in 47.13s ====
```

keikoro:
> Correct, I had to change these files for local development on my fork. Can't do much about it!

The keyword here is _local_. Not everything you do locally needs to be added to Git and thus to the shared codebase. In fact most personal dev tooling shouldn't.

In this particular case I'll give you ignoring the `.venv` directory might be helpful to other people too, but that doesn't change what I said initially. It's unrelated to what the PR advertised it was doing.

The change to the Changelog file additionally makes it look like either a drive-by submission (for which you didn't really look at the file) or like you're overestimating the value of the contribution. Its very first sentence says: 

https://github.com/Zulko/moviepy/blob/05b0babbcbe5a0d7676aa8d80d995bed689eb02b/CHANGELOG.md?plain=1#L3

Kaszanas:
> > Correct, I had to change these files for local development on my fork. Can't do much about it!
> 
> The keyword here is _local_. Not everything you do locally needs to be added to Git and thus to the shared codebase. In fact most personal dev tooling shouldn't.
> 
> In this particular case I'll give you ignoring the `.venv` directory might be helpful to other people too, but that doesn't change what I said initially. It's unrelated to what the PR advertised it was doing.
> 
> The change to the Changelog file additionally makes it look like either a drive-by submission (for which you didn't really look at the file) or like you're overestimating the value of the contribution. Its very first sentence says: 
> 
> https://github.com/Zulko/moviepy/blob/05b0babbcbe5a0d7676aa8d80d995bed689eb02b/CHANGELOG.md?plain=1#L3
> 

Might be, take it as you want it, cherry pick the changes if that's something you're up for. Leave the codebase as is if you don't want these changes.

In the end, if this deprecation notice is not addressed in time, and people start installing with new numpy they'll surely come back here and let you know.

I also did not perform any benchmarking on this change. If you have a suite for benchmarking and comparing runtimes that would also be helpful.

Feel free to copy the fix and try it yourself.

PR #2594, opened 2026-09-09
Preserve custom subtitle renderers when filtering captions

Bolt4243:
SubtitlesClip accepts a custom `make_textclip` renderer without a font path. Calling `match_expr` rebuilt the matching subtitles without that renderer, so every match raised `ValueError` before it could be used.

Pass the original renderer into the filtered clip and add a regression test using a lightweight custom ColorClip renderer.

Validation: the isolated AST harness fails against the baseline with the missing-font ValueError and passes with this change. The repository pytest suite was not run under the constrained validation environment.


keikoro:
Linking another open PR which changes the same file: #2529

